PR #264, opened 2022-03-31
Draft: Upgrading MLIR/LLVM and adapting our code to their breaking changes

daphne-eu:
In GitLab by @pdamme on Jul 31, 2021, 23:03

- I64IntegerAttr instead of IndexAttr in several places.
- Lowering to LLVM dialect: function calls must not have llvm.void as return type, otherwise it complains at runtime.
- Including the tablegen'erated DaphneOpsDialect.cpp.inc.
- Open problem: In the test cases for distributed operations (DistributedTest.cpp and WorkerTest.cpp), the LowerToLLVMPass fails (see details in the PR).

### DIFF
--- a/src/compiler/lowering/LowerToLLVMPass.cpp
+++ b/src/compiler/lowering/LowerToLLVMPass.cpp
@@ -17,9 +17,14 @@
 #include "ir/daphneir/Daphne.h"
 #include "ir/daphneir/Passes.h"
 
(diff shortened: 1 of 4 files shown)


pdamme:
I'm closing this outdated PR now, since I've just merged PR #468, which upgrades MLIR to an even newer version.

PR #318, opened 2022-03-31
DaphneLib Frames implementation

daphne-eu:
In GitLab by @dzcoralic on Mar 5, 2022, 16:14

Frames implementation for DaphneLib. Data transfer is handled via files.


pdamme:
Hi @dzcoralic, **thanks again for your pioneering work on DAPHNE's Python API**. I'm sorry that it took me so long to find the time to finalize it, but your work is highly appreciated and we will continue and build upon it now. This PR originally contained your work on the frame support in DaphneLib. As we have discussed recently, I have also copied over your remaining work from the branch `dzcoralic/daphne:daphnelib-shared-data-transfer` in your fork, which was about the zero-copy data exchange between DAPHNE and numpy. I have tidied up the code a bit and split the contributions into meaningful commits.

pdamme:
I will further handle and merge this PR myself. The commits shall not be squashed, but all put on main.

pdamme:
@corepointer: *I have a question regarding the CI and containers:* The changes in this PR require pandas for DAPHNE's Python API and test cases. I have added the package `python3-pandas` to `containers/daphne-dev.Dockerfile`. However, I assume some action is required to re-build and push the container that is used for the CI to DockerHub. It would be great if you could help me out with that.

pdamme (approved):
As mentioned above, I have tidied up @dzcoralic's contributions and everything works fine for now.

For those who are interested: This PR introduces a change worth mentioning: DAPHNE is now also built as a shared library (`libdaphnelib.so`), in addition to the executable (`daphne`). To this end, the existing code in `src/api/cli/daphne.cpp` has been moved to `src/api/internal/daphne_internal.cpp` (and slightly adapted). But nothing has changes in terms of how the `daphne` executable is used. Please read the commit message of 4c9fedad8f4479ddb60e7c5d3f1b44598433966e "DAPHNE as a shared library") for more details.

corepointer:
> @corepointer: _I have a question regarding the CI and containers:_ The changes in this PR require pandas for DAPHNE's Python API and test cases. I have added the package `python3-pandas` to `containers/daphne-dev.Dockerfile`. However, I assume some action is required to re-build and push the container that is used for the CI to DockerHub. It would be great if you could help me out with that.
> 
> **EDIT:** The fact that pandas is missing in the pre-built container is the reason why the CI tests fail.

Yes you're right, the missing pandas was the culprit - fixed [x]. Unfortunately we don't have auto-container-builds set up yet. 

BRM

pdamme:
@corepointer: Great, thank you so much so the quick help :+1:.

As we change those dependencies that need to be installed on the system (as opposed to being installed by the build script) rarely, auto-container-builds are not crucial right now, from my point of view.

Issue #221, opened 2022-03-15
Reference counter management for arguments/results of kernel calls


daphne-eu:
In GitLab by @pdamme on Mar 15, 2022, 13:53

DAPHNE's garbage collection of data objects (matrix/frame), e.g., intermediate results, is based on reference counting. The compiler makes sure that the reference counters are increased/decreased at the right points to make sure that each data object is freed exactly once.

However, the DAPHNE compiler cannot know what happens inside a kernel written in C++. While kernels typically create a new data object for their result, there are many situations when a kernel might want return (one of) its inputs as-is (not only the same underlying data, but really the same data object). Examples include elementwise addition with zero, sorting an already sorted frame, transposing a 1x1 matrix, etc. In many of these scenarios, a lot of unnecessary work can be saved by simply returning the input data object. Nevertheless, the problem is that this creates an additional reference to the same runtime data object in the IR. Thus, it must be ensured that the reference counter of that object is increased.

At the time of this writing, we don't have any kernels that return their input data object as-is (to the best of my knowledge), but this will be the case later.

There are multiple ways how to address this issue, with different responsibilities for the compiler/runtime and the kernel implementer.

PR #320, opened 2022-03-31
Draft: [DAPHNE-#221] Combined compile-time/run-time mechanism for increasing the reference counters...

daphne-eu:
In GitLab by @pdamme on Mar 15, 2022, 14:15

The the commit message for an overview of the current solution.

The code itself would be ready for a merge. However, the downside is that it works only for kernel calls from the JIT-compiled program. However, there are many situations when we wnat to call a kerner from a kernel.

**So we need to think about it again, before we finalize it.** Maybe a solution putting more responsibility on the kernel implementer will be better.


pdamme:
We will solve this problem in a different way.

PR #321, opened 2022-03-31
Adding deploy/ folder for packaging and deploying stages. It builds, packages,...

daphne-eu:
In GitLab by @aleszamuda on Mar 17, 2022, 14:13

Adding deploy/ folder for packaging and deploying stages. It builds, packages, and deploys the code. It also includes a demo.


pdamme (commented):
Thanks a lot, @aleszamuda, these are great contributions to the DAPHNE project. After your original merge request on GitLab, we have already discussed it offline. We came to the conclusion that:
1. The parts specific to our initial experiments should better be published as a separate repository to make them stand out better, and since they cannot directly be reused.
2. The parts for the general deployment on an HPC platform should be generalized to make them more reusable before we merge them into the DAPHNE repo. This led to a separate pull request: #329.

Therefore, I'll close this pull request now.

Issue #216, opened 2022-03-09
ANTLR not found while building


daphne-eu:
In GitLab by @EricMier on Mar 9, 2022, 16:50

On some systems the cmake config for ANTLR can't be found, resulting in following error message:

    CMake Error at src/parser/CMakeLists.txt:32 (find_package):
      By not providing "Findantlr4-generator.cmake" in CMAKE_MODULE_PATH this
      project has asked CMake to find a package configuration file provided by
      "antlr4-generator", but CMake did not find one.

      Could not find a package configuration file provided by "antlr4-generator"
      with any of the following names:

        antlr4-generatorConfig.cmake
        antlr4-generator-config.cmake

      Add the installation prefix of "antlr4-generator" to CMAKE_PREFIX_PATH or
      set "antlr4-generator_DIR" to a directory containing one of the above
      files.  If "antlr4-generator" provides a separate development package or
      SDK, be sure it has been installed.

This is due to incorrect path generation in [src/parser/CMakeLists.txt#L28](https://gitlab.know-center.tugraz.at/daphne/prototype/-/blob/master/src/parser/CMakeLists.txt#L28), where on some systems ${CMAKE_INSTALL_LIBDIR} is resolved to "lib/x86_64-linux-gnu" resulting in:

    antlr4-generator_DIR = <project>/thirdparty/antlr/antlr4-cpp-runtime-4.9.2-source/run/usr/local/lib/x86_64-linux-gnu/cmake/antlr4-generator

whereas the required cmake config files reside in:

    <project>/thirdparty/antlr/antlr4-cpp-runtime-4.9.2-source/run/usr/local/lib/cmake/antlr4-generator

Same goes for antlr4-runtime.

PR #328, opened 2022-03-31
Build script overhaul.

daphne-eu:
In GitLab by @EricMier on Mar 28, 2022, 20:36

- Check if dependencies are installed by indicator file.
- Made (the relevant) relative directory changes absolute (pls avoid errors by changing relatively)
- Complemented clean command (User has to confirm before deleting vital files)
- Made variable-/argument-handling bash-save
- Added some spice to the console output (had some spare time at the weekend :D )
- Dev-issue #216 included


EricMier:
Hi Mark, 
good point. I will add an option to opt-out the colorization. For the interactive part, since only the clean is interactive, I could simply add an -y option to skip the prompt.

Unfortunately I am not the owner of this Merge-Request anymore, so I can not change it to a Draft. So consider this Request as a Draft for now :)

Regards, Eric

corepointer:
SGTM - and don't worry about the draft state ;-)

EricMier:
Hi Mark,
my changes should be complete now. View commit 4f9638d for the changes. I will also add a separate docu page the next days, but this shouldn't require a another pull request.

Regards, Eric

corepointer (changes requested):
Hi! 
I have tried this out in an Ubuntu 20 container.
The initial build worked fine. But...
* There's an additional external dependency on the 'bc' package. No problem as long as we document it. I updated my container and it was fine.
* When I tried --cleanALL I got: Unknown option: '--cleanALL'
* --yes (and also -y) in combination with --clean still asked for a y/n
* -nf (also both, long and short option) in combination with --cleanALL still came back with a fancy error message that the option is unknown.

The bash version in the container is 5.0.17.

Once everything's fixed, I'd adapt my pending changes onto yours to merge it in one go, so other people would only face one recompile.

pdamme (changes requested):
I've now also had a look at it. In general, it **looks very good and useful** to me, and there are several things I like about this "overhaul":
- Success token files surely make the dependency management more robust.
- Colored `[DAPHNE]` messages make very clear which output comes from the build script and which from the build processes of the dependencies.
- Using `ninja` in more places makes it a bit faster (on my system).
- The logo looks nice :) .

However, there are a few things that **should be fixed/changed before we merge it**.
- I recommend removing the animations, since they (a) introduce artificial delay (not good, especially for repeated use during development) and (b) require `bc`, as @corepointer pointed out. *However, putting a `sleep` for a few seconds after the logo (which is only shown on the first build and after cleans) is totally fine, IMHO*.
- Some issues with cleaning:
  - The path to the build directory of ANTLR is wrong (missing `v*` or so).
  - Build after clean (`./build.sh --clean` followed by `./build.sh`) crashes because some OpenBLAS file is not found. Build after `--cleanAll` works, though.
- Minor things:
  - The indentation should be four spaces, consistently.

Besides that, I just want to mention that the overhauled `build.sh` crashes on an existing clone, because files/directories are expected with different names etc. We shouldn't invest effort in making it compatible, though, since the "old" version will soon be forgotten anyway. But once we merge it, we should notify everyone via the dev mailing list to do a clean first (not sure if `--cleanAll` removes the *old* files).

Finally, I've just pushed a few minor changes that I quickly fixed myself.

EricMier:
Hi Patrick @pdamme,

also thanks for your review and your comments.
Regarding your points:
- I removed the animation as suggested. It was hard to hit the specified time of the animation anyway :D
- Cleaning
    - Antl: I moved the folders created while building antlr out of the source code folder (e.g. v4.9.2) like for other dependencies. So we don't need to specify the version we want to clean, but simply wipe the dedicated build/run folder. 
    - OpenBLAS: I forgot to change into the OpenBLAS directory when only building. Should be fixed now.
- Added some spaces.

Also, the build should not crash on existing clone anymore. So a --cleanAll shouldn't be necessary, but to get rid of old files.

corepointer:
Thank you for your efforts! I wouldn't mind the extra dependency of bc, as it's rather painless to address that. It just needs to be mentioned somewhere. But if the animation gives troubles the the solution to remove it is just fine.
@ cleanALL - thanks for pointing it out/correcting. Could have looked more thoroughly myself :blush: 
I'd prepare my changes on top of yours to merge everything in at once, so disturbance of recompiling dependencies is minimized. 

Regards, Mark

EricMier:
Changes are included in #359. Closed.

PR #331, opened 2022-03-31
New IO kernels: Matrix Market, Parquet and DAPHNE Binary

daphne-eu:
In GitLab by @niclashedam on Mar 30, 2022, 13:34


niclashedam:
Hey @pdamme. Do you have time to review our new IO kernels at some point?

mboehm7:
Patrick is one vacation, but I try to have a look end of this week.

pdamme:
@mboehm7 Thanks, but you don't need to, I had already looked over it two weeks ago, just didn't find the time to summarize it.

@niclashedam @mortentc @phbonnet Thanks a lot for this contribution! **It's an impressive amount of work and definitely brings us forward**. I mainly paid attention to the integration into the code base rather than to the details of the readers/writers. In general, it **looks very good to me**. I've applied some minor adjustments myself (see 905a71a8df54bc459c4d87ee9e9cdd9e289bd4bf). However, I need your help/feedback on three points before we can merge it in:

1. Before my adjustments, the code didn't compile without Arrow. I quick-fixed that by inserting `#ifdef`s around the Arrow/Parquet case in `Read.h`. That works and **I hope you're fine with that**.
2. I've also tried to build/test with Arrow (by using the `-DUSE_ARROW=ON` you left as a comment in `build.sh`). Unfortunately, I get the following error:

   ```
   CMake Error at CMakeLists.txt:105 (find_package):
     Could not find a package configuration file provided by "Parquet" with any
     of the following names:
   
       ParquetConfig.cmake
       parquet-config.cmake
   
     Add the installation prefix of "Parquet" to CMAKE_PREFIX_PATH or set
     "Parquet_DIR" to a directory containing one of the above files.  If
    "Parquet" provides a separate development package or SDK, be sure it has
     been installed.
   ```
   I installed Arrow/Parquet following the instructions at https://arrow.apache.org/install/ (packages `libarrow-dev`, `libarrow-glib-dev`, `libparquet-dev`, `libparquet-glib-dev`). Is that the right way? **It would be great if you could add some hints in `GettingStarted.md`**. Or, if the Arrow-support is still meant to be experimental, then it's also fine to fix that later, please just let me know.

3. In the test case `WriteDaphneTest.cpp`/`"WriteDaphne AIG (Frame)"` the call to `writeDaphne()` was commented out; similarly the whole test case `ReadDaphneTest.cpp`/`"ReadDaphne AIG (Frame)"` was commented out. I uncommented it and it works well (tests pass). **Is there a particular reason why this was commented out?** I.e., should it rather stay inactive?

**Hint:** Since I saw you removed the DaphneLib/Python API test cases (I've reinserted them): You need to install `numpy` (e.g., by `sudo apt install python-numpy`) to make them work.

niclashedam:
Hi @pdamme. Thanks for your contribution in https://github.com/daphne-eu/daphne/commit/905a71a8df54bc459c4d87ee9e9cdd9e289bd4bf.

I can only comment on points 1 and 2. Arrow is a bit complex to get working. All Arrow/Parquet related stuff should have the USE_ARROW, so it feels completely OK to have it in `Read.h` as well.

I personally installed Arrow and Parquet from source because we need a few features like CSV as well. This should install all needed libraries. I can schedule some time to write the documentation for it next week. Maybe we can handle that as a separate PR to avoid blocking this?

pdamme (approved):
(See my comment above for my review. This comment is just for formal approval.)

pdamme:
Thanks for the response @niclashedam. As `USE_ARROW=ON` is currently commented out anyway, I'd say it's okay if the right way to install Arrow is not documented yet. Once you find some time, it would be great to add an argument for it, e.g., like `./build.sh --arrow`, and to briefly explain how to provide Arrow as an optional dependency in `GettingStarted.md`.

I also assume that point 3 about the test case should be fine, so I merge it in now. **Thanks again to everyone involved!**

I'm looking forward to the next contributions from your team. Whenever it is possible to divide contributions by different people into separate branches/pull requests, I recommend doing so. This often makes it easier to review a PR and apply/request changes if needed, as well as to squash commits if suitable. But in this case I can imagine that it was easier for you to jointly develop this initial set of readers/writers.

phbonnet:
Thanks. Welcome back. Hope you had a nice vacation.
About 3. This is my bad. No reason to take it out.
Best,
Philippe.
________________________________

PR #333, opened 2022-03-31
[DAPHNE-#31] Kernel for `diagMatrix`-operation

### LINKED ISSUE #31: Kernel for `diagMatrix`-operation
In GitLab by @pdamme on Apr 21, 2021, 13:25

The `diagMatrix`-operation creates a diagonal matrix from a column-matrix. That is, it takes a column-matrix `arg` (i.e., a matrix with *n* rows and *1* column) as input, and outputs an *n x n* matrix, whose diagonal contains the values of `arg`, while all other entries are zero.

**Example**

```
arg: 1,
     5,
     3,

res: 1, 0, 0,
     0, 5, 0,
     0, 0, 3,
```

**This issue asks you to implement**

1. The general template definition of the kernel struct.
2. The kernel function calling the `apply`-method of the kernel struct.
3. A partial specialization of the kernel struct for `DenseMatrix` as the data type implementation of the output matrix, and analogously for `CSRMatrix`.
4. Ideally some test cases.

**Hints**
- Have a look at the [guidelines on implementing a built-in kernel](https://gitlab.know-center.tugraz.at/daphne/prototype/-/blob/master/doc/ImplementBuiltinKernel.md).
- It is fine to assume that the input column-matrix `arg` is a `DenseMatrix`, irrespective of the data type implementation of the output matrix.

daphne-eu:
In GitLab by @psomas on Mar 31, 2022, 10:53

- Add CSR specializations for the diagMatrix kernel.
- Add sparsity inference for the diagMatrix() daphne op.

Closes #31

### DIFF
--- a/src/ir/daphneir/DaphneInferSparsityOpInterface.cpp
+++ b/src/ir/daphneir/DaphneInferSparsityOpInterface.cpp
@@ -57,6 +57,18 @@ double getSparsityOrUnknownFromScalar(Value v) {
 // Sparsity inference interface implementations
 // ****************************************************************************
 
+std::vector<double> daphne::DiagMatrixOp::inferSparsity() {
+    auto argTy = arg().getType().dyn_cast<daphne::MatrixType>();
+    auto k = argTy.getNumRows();
+    auto sparsity = argTy.getSparsity();
+
+    if(argTy.getSparsity() == -1.0) {
+        sparsity = 1;
+    }
+
+    return {sparsity / k};
+}
+
 std::vector<double> daphne::MatMulOp::inferSparsity() {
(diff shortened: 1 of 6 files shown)


pdamme (approved):
Thanks @psomas for this kernel and related compiler parts. It looks very good to me. I just noticed that the the kernel specialization with `DenseMatrix` argument and `CSRMatrix` result could still store zeros in the result; I've quickly fixed that.

From my point of view, it's ready to be merged. To get it correctly attributed to you after squashing, could you please re-open the PR in your own name (due to the migration from GitHub, it was formally opened by @daphne-eu).

Issue #191, opened 2022-02-16
Refactor CUDA buffer mgmt


daphne-eu:
In GitLab by @corepointer on Feb 16, 2022, 17:24

At the moment this sits mainly in DenseMatrix, where it does not belong.

PR #334, opened 2022-04-06
[DAPHNE-191] Refactor CUDA buffer mgmt (aka introducing Object Meta D…

corepointer:
…ata)

This change introduces a major change how external storage buffers (CUDA memory specifically) are handled. In that regard, the following noteworthy changes are implemented:
* Factor out CUDA allocations from DenseMatrix (one of the initial motivations of issue #191)
* Introduce a mechanism to handle several storage backends and track ranges of a Structure's data.
* To make use of the mechanism, an AllocationDescriptor is passed to create() and getValues() (at the moment only DenseMatrix is supported).
* Allocation descriptors need to implement the IAllocationDescriptor interface. This decouples backend specific dependencies.
* AllocationDescriptorHost and AllocationDescriptorCUDA are implemented atm. The former is more or less a no-op for now.
* The CUDA memory allocation and data movement is moved to the CUDAContext class. It keeps track of its allocations per device. For now this does nothing but can be used to reuse allocations in the future.

### DIFF
--- a/src/api/cli/DaphneUserConfig.h
+++ b/src/api/cli/DaphneUserConfig.h
@@ -17,7 +17,7 @@
 
 #pragma once
 
-#include <runtime/local/vectorized/LoadPartitioning.h>
+#include <runtime/local/vectorized/LoadPartitioningDefs.h>
 
 #include <vector>
 #include <string>
--- a/src/runtime/local/context/CUDAContext.cpp
+++ b/src/runtime/local/context/CUDAContext.cpp
@@ -16,6 +16,8 @@
 
 #include "runtime/local/context/CUDAContext.h"
 
+size_t CUDAContext::alloc_count = 0;
+
 void CUDAContext::destroy() {
 #ifndef NDEBUG
     std::cerr << "Destroying CUDA context..." << std::endl;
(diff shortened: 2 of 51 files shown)


aristotelis96:
Thank you @corepointer for this great contribution! One thing though, on my machine when I run the following script:
`a = rand(5,5,0.0,10.0,1,-1);`
`print(a+a);`
on the local runtime: `./build/bin/daphne test.script` I get a segmentation fault.

corepointer:
Thank you, @aristotelis96, for trying out these code changes and finding the bugs and pointing me right to them :D Obviously I did not clean up the code properly and code, that I commented for testing was not put in place again. I also found a few other issues. While cleaning it up and testing it. New version coming soon :copyright:

aristotelis96:
Thanks @corepointer for fixing those issues. I found a new bug with this PR (haven't tested which commit might have introduced it though). This bug applies **only** for the vectorized engine (`--vec`). There seems to be something wrong with the transpose function. The following script should return different results. 
The result dimensions seem to be fine, however the resulted values of the matrix, are wrong.
```
a = rand(10,5,0.0,10.0,1,123);
print(t(a));
```

Issue #128, opened 2021-09-15
Kernel for CastOp from DenseMatrix of one value type to DenseMatrix of another value type


daphne-eu:
In GitLab by @pdamme on Sep 15, 2021, 20:28

**Background**

In the DAPHNE prototype, data can be represented in different physical format in-memory. We distinguish between the data type (e.g. `DenseMatrix`, `CSRMatrix`, `Frame`) and value type (e.g. `float`, `double`, `uint32_t`, `int64_t`, ...). DaphneIR's `cast`-operation can be used to change the physical representation of a data object from one combination of data/value type to another combination of data/value type.

In the runtime, these casts are performed by the `castObj`-kernel, whose general template definition resides in `src/runtime/local/kernels/CastObj.h`. However, we still lack a rich set of specializations for different source and target data/value types.

**This issue asks you to implement**

1. A partial specialization of the `castObj`-kernel for casting from `DenseMatrix<VTArg>` to `DenseMatrix<VTRes>`, i.e., for changing the value type, while retaining the data type.
4. Ideally some test cases in `test/runtime/local/CastObjTest.cpp`.

**Hints**
- Have a look at the [guidelines on implementing a built-in kernel](https://gitlab.know-center.tugraz.at/daphne/prototype/-/blob/master/doc/ImplementBuiltinKernel.md).

PR #337, opened 2022-04-07
[DAPHNE-#128] Kernel for CastOp from DenseMatrix of one value type to DenseMatrix of another value type.

akroviakov:
- Features a kernel for cast operation for DenseMatrix class
- One loop to make use of dense array storage of DenseMatrix
- Check for dimensions
- Create DenseMatrix if not already allocated

### DIFF
--- a/src/runtime/local/kernels/CastObj.h
+++ b/src/runtime/local/kernels/CastObj.h
@@ -152,5 +152,31 @@ public:
         res = DataObjectFactory::create<Frame>(cols, nullptr);
     }
 };
+template<typename VTArgTo, typename VTArgFrom>
+class CastObj<DenseMatrix<VTArgTo>, DenseMatrix<VTArgFrom>> {
+
+public:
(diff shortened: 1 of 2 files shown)


pdamme (approved):
Thanks for this contribution @akroviakov. Overall, it **looks very good to me**. There were a few minor things that I quickly fixed myself. Most importantly (please see the commit I added for the details):

- I removed the check for the shape of `res` (and the corresponding test case), since we generally haven't decided how/if to check the output shape yet (none of the other kernels does it). It's a good point, but we should address it separately and adapt all kernels accordingly one day.
- I added a code path for processing matrices which are views into another matrix. In that case, we cannot iterate over all elements in one loop, but must proceed row by row. In the simplest case, a `DenseMatrix` is backed by a contiguous array. However, a `DenseMatrix` could also be a view into a larger `DenseMatrix`. `DenseMatrix::rowSkip` represents the number of columns in the underlying array; for (column) views, this is larger than `numCols` and the part of the array *used by the view* is not contiguous anymore.

Issue #196, opened 2022-03-01
Add support for work stealing using multiple task queues in vectorized execution


daphne-eu:
In GitLab by @jonnygiger on Mar 1, 2022, 16:57

Currently all CPU tasks are added to a single queue which multiple workers can dequeue from. Creating multiple queues and distributing the tasks among the queues can reduce lock contention on the queue mutex and improve task-to-data affinity. Work stealing can be implemented to reduce load imbalance among the queues.

philipportner:
Has been implemented with #342

PR #342, opened 2022-04-22
[DAPHNE-196] Multiple task queues and work stealing

jonnygiger:
For [issue #196](https://github.com/daphne-eu/daphne/issues/196)

This is a concept to implement optional per-CPU and per-socket task queues with work stealing support. 

TODO:
- ~~Sanity checking the hardware probing~~
- ~~Better queue-to-core mapping for the per-socket option~~
- Check if CPU pinning is already implemented by a parent process (e.g. SLURM) before pinning CPUs


corepointer:
This is a friendly reminder to squash commits before merging to main.

pdamme:
This is a friendly reminder (to all of us) to agree on certain rules for merging in PRs, first of all ;) .

In general, it's great if more collaborators participate and handle PRs, that really helps to distribute the load, so thanks a lot @AhmedEleliemy. So far, we have usually squashed commits into one before merging, to keep a clean history and to make it easy for everyone to follow the most recent changes. There can be exceptions, though, for instance, if a branch is very well structured into meaningful commits (I haven't checked if this was the case here). In that case I'd vote for rebasing the branch before the merge to get the new commits on top, where everyone staying up-to-date will see them.

src/runtime/local/vectorized/Tasks.cpp
```diff
@@ -65,6 +65,11 @@ void CompiledPipelineTask<DenseMatrix<VT>>::execute(uint32_t fid, uint32_t batch
     }
 }
 
+template<typename VT>
+uint64_t CompiledPipelineTask<DenseMatrix<VT>>::getTaskSize() {
+return _data._ru;
```
corepointer, line 70: Just checking: Why is this different from the CSR version below?

src/runtime/local/vectorized/MTWrapper_dense.cpp
```diff
@@ -31,11 +31,32 @@ void MTWrapper<DenseMatrix<VT>>::executeSingleQueue(
...
 
-    // create task queue (w/o size-based blocking)
-    std::unique_ptr<TaskQueue> q = std::make_unique<BlockingTaskQueue>(len);
+    std::vector<std::unique_ptr<TaskQueue>> q;
+    std::vector<TaskQueue*> qvector;
+    if (ctx->getUserConfig().pinWorkers) {
+        for(int i=0; i<this->_numQueues; i++) {
+            // Ideally queues would be created as such:
+            // q.emplace_back(new BlockingTaskQueue(len));
+            // However I'm getting an error so they are created with make_unique and get now.
```
corepointer, line 40: The BlockingTaskQueue class contains a non copy-able/moveable member (the condition variable). This is why you have to work with (unique) pointers here.

src/runtime/local/vectorized/Tasks.cpp
corepointer: OK thx for the update. I'll have it fixed with a few other things to make the CUDA code paths fly again.

PR #347, opened 2022-05-01
[DAPHNE-#218]: colBind-kernel for CSRMatrix

### LINKED ISSUE #218: colBind-kernel for CSRMatrix
In GitLab by @pdamme on Mar 11, 2022, 12:50

The `colBind`-operation concatenates two matrices or frames `lhs` and `rhs` horizontally and returns the result as a new data object `res`. The two inputs must have the same number of rows.

**Example**

```
lhs: 1, 2,
     3, 4,

rhs: 0, 1, 0,
     1, 0, 1,

res: 1, 2, 0, 1, 0,
     3, 4, 1, 0, 1,
```

We already have specializations of the `rowBind`-kernel for `DenseMatrix` and `Frame`, but still lack one for `CSRMatrix`.

**This issue asks you to implement**

1. A partial specialization of the kernel struct for `CSRMatrix` as the data type implementation of all input and output matrices.
4. Ideally some test cases.

**Hints**
- Have a look at the [guidelines on implementing a built-in kernel](https://gitlab.know-center.tugraz.at/daphne/prototype/-/blob/master/doc/ImplementBuiltinKernel.md).

akroviakov:
- Implements column binding for CSRMatrix datatype.
- Closes #218.

### DIFF
--- a/src/runtime/local/kernels/ColBind.h
+++ b/src/runtime/local/kernels/ColBind.h
@@ -118,21 +118,21 @@ struct ColBind<CSRMatrix<VT>, CSRMatrix<VT>, CSRMatrix<VT>> {
         size_t lhsStartOffset = lhsRowOffsets[0];
         size_t rhsStartOffset = rhsRowOffsets[0];
 
-        for(size_t rowRes = 0; rowRes < res->getNumRows(); rowRes++){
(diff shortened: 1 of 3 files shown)


pdamme (approved):
Thanks, **looks very good**. I'll merge it in. I've only changed two variable names (see my latest commit).

PR #348, opened 2022-05-03
[DAPHNE-#224]: reshape-kernel for non-zero-copy cases

### LINKED ISSUE #224: reshape-kernel for non-zero-copy cases
In GitLab by @pdamme on Mar 16, 2022, 14:06

DaphneIR's `ReshapeOp` changes the shape of a data object, while retaining the data in it. For instance, the following DaphneDSL script generates the following output:

```
X = seq(1, 12, 1);
print(X);
print(reshape(X, 3, 4));
```
```
DenseMatrix(12x1, int64_t)
1
2
3
4
5
6
7
8
9
10
11
12
DenseMatrix(3x4, int64_t)
1 2 3 4
5 6 7 8
9 10 11 12
```

We already have an implementation of the `reshape`-kernel on `DenseMatrix`. However, it can only handle the zero-copy case when the input matrix is not a view into a subset of the columns of another matrix and no result data object is given. In that case, `reshape` is a pure metadata operation: we simply wrap the input matrix's underlying data buffer into a new `DenseMatrix` object with the desired shape. Consequently, the following throws an error at the moment:

```
X = reshape(seq(1, 12, 1), 3, 4);
Y = X[, 2]; // Y is a view onto the 2nd column from X
print(Y);
print(reshape(Y, 1, 3)); // should yield [3, 7, 12] as a 1x3 matrix
```
```
DenseMatrix(3x1, int64_t)
3
7
11
terminate called after throwing an instance of 'std::runtime_error'
  what():  reshape does not support cases which aren't pure meta data operations yet
```

**The task**
- Extend the existing `reshape`-kernel for `DenseMatrix` in `src/runtime/kernels/Reshape.h`, such that it can handle inputs which are a view into a column segment of another matrix.
- Extend the existing unit test cases for the `reshape`-kernel, *or* add script-level test cases.

**Hints**
- If the input matrix is a view into a column segment of another matrix (#columns different from row skip), you need to create a new matrix and copy row by row from the input to the result.
- If a result matrix is given (`res != nullptr`), you can use that data object, but must copy from input to result.

akroviakov:
- Implements reshape-kernel for views on `DenseMatrix` datatype.
- Closes #224.

### DIFF
--- a/src/runtime/local/kernels/Reshape.h
+++ b/src/runtime/local/kernels/Reshape.h
@@ -59,9 +59,20 @@ struct Reshape<DenseMatrix<VT>, DenseMatrix<VT>> {
 
         if(arg->getRowSkip() == arg->getNumCols() && res == nullptr)
             res = DataObjectFactory::create<DenseMatrix<VT>>(numRows, numCols, arg->getValuesSharedPtr());
-        else
-            // TODO Support those cases. See #224.
-            throw std::runtime_error("reshape does not support cases which aren't pure meta data operations yet");
+        else {
+            if(res == nullptr)
+                res = DataObjectFactory::create<DenseMatrix<VT>>(numRows, numCols, false);
+
+            auto resVals = res->getValues();
+            auto argVals = arg->getValues();
+            size_t numArgRows = arg->getNumRows();
+            size_t numArgCols = arg->getNumCols();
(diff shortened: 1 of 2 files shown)


pdamme (approved):
Thanks, looks perfect.

PR #349, opened 2022-05-06
[DAPHNE-#215]: Ternary/conditional operator `?:` in DaphneDSL

### LINKED ISSUE #215: Ternary/conditional operator `?:` in DaphneDSL
In GitLab by @pdamme on Mar 8, 2022, 18:39

Many script/programming languages offer a ternary/conditional operator, which evaluates a condition and, depending on the outcome, returns the value of one of the two given expressions. For instance, in `C`-like languages, this is done by
```
v = conditionExpr ? thenExpr : elseExpr;
```

Something similar (can be exactly the same syntax) would be very handy in DaphneDSL, too. Supporting this requires:
1. Adding a new kind of expression in the DaphneDSL ANTLR grammar in `src/parser/daphnedsl/DaphneDSLGrammar.g4`.
2. Adding a visitor function for this new kind of expression in `src/parser/daphnedsl/DaphneDSLVisitor.h/cpp`.
3. Adding a some script-level test cases in `test/api/cli/expressions` (to be created).

**Hints:**
- The ternary/conditional operator is essentially just syntactic sugar for an if-then-else statement. Therefore, instead of implementing a completely new visitor, the try to reuse the code in `DaphneDSLVisitor::visitIfStatement()`. This will also ensure that exactly the same restrictions hold for both constructs (e.g., regarding types of the then/else branches' outcomes).

akroviakov:
- Implements ternary `?:` operator (new expression).
- Some script level tests.
- Closes #215.


pdamme (changes requested):
Thanks, @akroviakov. **Overall, it looks good to me**. However, there are **a few points that need to be addressed before we can merge it**:

1. `visitTernExpr()` visits the `thenExpr` and `elseExpr` twice (the first time for inserting their code into the `thenBlock`/`elseBlock`, the second time to capture their return values). This will result in a duplication of their code in the IR (this should be visible if you print the IR after parsing, see `src/compiler/execution/DaphneIrExecutor.cpp`). **Please simply capture their return values into a variable when you call `visit()` the first time**.
2. The test cases seem to be incomplete. There are three DaphneDSL files, but no unit tests executing them and no text files for the reference output. Did you forget to push something? **Please complete the test cases.**
3. It's not functionally wrong, but it might confuse readers of the code: Please don't use `valueOrError()` on the `IfOp`'s result when you return it. `valueOrError()` is intended to be used with arguments of type `antlrcpp::Any`, which are returned by all the visitor functions; it should only be used on the return values of a `visit()`-call. Instead, you can **simply `static_cast<mlir::Value>()` the `IfOp`'s result**.

pdamme (approved):

pdamme:
Thanks @akroviakov, looks good to me. I've simplified the test cases a bit, though ;) .

PR #360, opened 2022-05-09
[DAPHNE-#350]: DaphneDSL: rbind checks number of rows instead the number of columns

### LINKED ISSUE #350: DaphneDSL: rbind checks number of rows instead the number of columns
```rbind``` is checking the number of rows instead of columns.

```
r = rand(1,5,0,1,1,-1);
r2 = rand(1,5,1,2,1,-1);
r3 = rand(1,5,2,3,1,-1);
print(r);
print(r2);
print(r3);
r = rbind(r,r2);
print(r);
r = rbind(r,r3);
```
Ouput:
grafik

aristotelis96:
Fixed a bug where inferNumColsFromArgs would use rows instead of columns.


pdamme (approved):
Thanks @aristotelis96, that fixes it.

PR #361, opened 2022-05-09
[DAPHNE-#356]: type casting of DenseMatrix

### LINKED ISSUE #356: DaphneDSL: Type casting matrices
The type conversion on matrices is not working. 

```
A = rand(5,5,0.0,1.0,1,-1);
print(A);
B = as.matrix.f32(A);
# Output:
# JIT session error: Symbols not found: 
#[ _cast__DenseMatrix_float__DenseMatrix_double ]
```

grafik

aristotelis96:
CastObj kernel has already specialized implementation for `DenseMatrix<ValueType>` to `DenseMatrix<ResultValueType>`, but `kernels.json` was missing the declarations.

I open this PR as a Draft in case we want to include more cases.

Closes #356.


pdamme (approved):
Thanks @aristotelis96, looks very good. I think it's indeed a good idea to add the remaining type combinations right now (taking the currently supported value types into account).

PR #362, opened 2022-05-11
[DAPHNE-#31] Kernel for diagMatrix-operation

psomas:
- Add CSR specializations for the diagMatrix kernel.
- Add sparsity inference for the diagMatrix() daphne op.

Closes https://github.com/daphne-eu/daphne/issues/31


pdamme:
Thanks, @psomas for reopening #333 in your own name.

Issue #242, opened 2022-03-28
Kernel for GroupOp (grouping by with aggregation) on Frame


daphne-eu:
In GitLab by @pdamme on Mar 28, 2022, 19:33

To support SQL `GROUP BY`, we need a kernel for DaphneIR's `GroupOp` (see `src/ir/daphneir/DaphneOps.td`). The kernel should have the following interface:

```
void group(
        Frame *& res,
        const Frame * arg,
        const char ** keyCols, size_t numKeyCols,
        const char ** aggCols, size_t numAggCols,
        mlir::daphne::GroupEnumAttr * aggFuncs, size_t numAggFuncs
);
```

The `Frame` `arg` shall be grouped by the `numKeyCols` columns (indicated by name) in `keyCols`. In addition to that, `numAggCols` aggregates shall be calculated for each group, whereby the columns to group on are indicated by `aggCols` and the corresponding aggregate functions (`COUNT`, `SUM`, ...) are indicated by `aggFuncs`. Note that `numAggCols` and `numAggFuncs` will always have the same value. Having this information replicated is an artifact of the way we currently lower variadic DaphneIR operations to kernel calls.

*For now, it's just about making it work.* We can make it efficient later. A straightforward approach would be a sort-based grouping, which sorts the input frame by the key columns using the `order`-kernel, and calculates the aggregates over the subsequent data elements in a group.

**Hints:**
- `COUNT` shall always return a `UI64` column.
- `AVG` shall always return a `FP64` column.

PR #363, opened 2022-05-12
Draft: [DAPHNE-#242] Kernel for GroupOp on Frame

Hiebl:
* Implements column-wise group aggregation on Frames
* Supports the aggregation functions count, sum, min, max and average
* Only materialzes an ordered version of the input frame and the result frame
* Reduces the input frame to key and aggregation columns before mutli-column ordering (without copying) for efficieny
* Utilizes the Order & extractCol Kernel, as well as the  DeduceType Util
* Refactors the Order Kernel to use the DeduceType Util
* Adds an optional export the group vector from the Order Kernel
* This option is less efficient in the last sort step, but overall increases efficiency in scenarios where the final group vector is reused (e.g. Group Kernel)
* Fixes a Warning in ExtractCol (redundant comparison)
* closes https://github.com/daphne-eu/daphne/issues/242

### DIFF
--- a/src/runtime/local/kernels/Group.h
+++ b/src/runtime/local/kernels/Group.h
@@ -122,10 +122,10 @@ template <> struct Group<Frame> {
         size_t numRowsArg = arg->getNumRows();
         size_t numColsRes = numKeyCols + numAggCols;
         size_t numRowsRes = numRowsArg;
-        if (arg == nullptr || keyCols == nullptr || numKeyCols == 0 || aggCols == nullptr || numAggCols == 0 || aggFuncs == nullptr || numAggFuncs == 0) {
+        if (arg == nullptr || (keyCols == nullptr && numKeyCols != 0) || (aggCols == nullptr && numAggCols != 0) || (aggFuncs == nullptr && numAggFuncs != 0))   {
             throw std::runtime_error("group-kernel called with invalid arguments");
         }
-
+        
         // convert labels to indices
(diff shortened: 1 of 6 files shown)


pdamme (changes requested):
Thanks @Hiebl for this contribution so far, and sorry for the delay. It already **looks good to me**. 

However, it would be great if you could **support the following two corner-cases before we merge it** (at the moment, both cases are explicitly caught as invalid inputs):
1. `numKeyCols == 0` should imply a pure aggregation over all rows (no grouping).
2. `numAggCols == 0` should imply a pure grouping/duplicate elimination (no aggregation).

While looking over your PR, I noticed that there were a few challenges to the overall integration of this kernel, including the lowering of DaphneIR's `GroupOp` to a kernel call. This was partly due to the fact that we didn't have a kernel with variadic *scalar* inputs before (except for `order`, where it had not been addressed yet, either). Since this was all a bit tricky, I did it myself and added a few general commits to main to pave the way. Then, I rebased your branch on main, and added two commits specific to this PR. Now all test cases succeed, including the SQL tests involving `GROUP BY`. This should be a good state for you to continue with the two points mentioned above.

Hiebl:
Added the edge cases in question (incl. Test)

pdamme (approved):
Thanks @Hiebl, **looks very good to me**. Nice that the corner cases could be added with such minimal changes :) .

Issue #240, opened 2022-03-26
Enable zero-copy for left indexing when possible


daphne-eu:
In GitLab by @pdamme on Mar 26, 2022, 21:31

Following DaphneDSL's copy-on-write semantics, the current implementations of the `insertRow`/`insertCol`-kernels underlying left indexing (like `X[100:200, 50:50] = ...;`) always create a new data object and copy over the data to exclude unwanted side-effects on other variables sharing the same underlying data. So from a functional point of view, that's fine. But it is very inefficient and in many cases, we could actually avoid this copying and modify/return the input data object. The decision can be made based on the reference counters of the input data object and those of the `std::shared_ptr`s of its underlying data buffers.

akroviakov:
Assigned to @akroviakov.

PR #365, opened 2022-05-16
[DAPHNE-#240]: Enable zero-copy for left indexing when possible

akroviakov:
- Implements the basic zero-copy mechanism for left indexing (`InsertRow()` and `InsertCol()`).
- Adds a new member function `getValuesUseCount()` in `DenseMatrix.h`.
- Script & unit tests.


pdamme (changes requested):
Thanks for this contribution, @akroviakov. Avoiding unnecessary copying in left indexing can greatly improve performance. **Overall, your implementation looks good to me.** **A few points still need to be fixed before we can merge it**, but *you don't need to do that anymore, of course* :) .

-----

There is just one point that may not have been fully clear before, and that doesn't work as expected at the moment (I modified the test cases to trigger this situation):

From `test/api/cli/indexing/left_indexing_success_5.daphne`:
```
A = fill(123, 3, 3);
B = A;
B[, 1] = [1, 2, 3]; // left indexing on B shall not affect A
print(A);
print(B);
```

With the current implementation, this test fails, as `A` and `B` are affected by the left indexing on `B`. Solving this issue is possible, but will involve the DAPHNE compiler. Interestingly, a more general problem (update-in-place) is already WIP (#498), maybe it will yield a solution for left indexing as well.

We should thoroughly address this issue, since it could otherwise cause unexpected behavior and crashes. Thus, I recommend not including this feature into the upcoming 0.2 release anymore.

PR #372, opened 2022-05-20
[DAPHNE-xyz] Sample integration of an accelerator API

corepointer:
This commit adds all the bits and pieces for the initial integration of an accelerator API (Intel OneAPI in this case)


corepointer:
Leaving this here as a draft as I want people to be able to have a look and comment if they want but I'd only merge it after the build system changes are through.

pdamme (approved):
Thanks @corepointer for this example of how to integrate a new hardware accelerator API. Besides that, the basic support for OneAPI is a great initial step towards supporting more hardware accelerators.

In general, it **looks good to me**. I fixed just a few small things (see my commit). A few minor comments, but they are not required for the merge:

**Optional changes:**
- Usually, we also add new passes to `src/ir/daphneir/Passes.td`. Nevertheless, I think that file is actually unused at the moment. And CUDA is not present there either.
- Ideally, we should mention `dpcpp` (and perhaps other OneAPI dependencies) in `GettingStarted.md`. Nevertheless, it seems like information on CUDA is missing there, too.
- The OneAPI MatMul-kernel still prints a todo message.

pdamme:
So feel free to merge it or let me know if I shall do it. But for now, I wait in case you want to apply some final changes.

corepointer:
Thank you for the review @pdamme! I'll go over it again before it can be merged.
* Passes.td - I ran into troubles there with ``daphne::createMarkCUDAOpsPass(const DaphneUserConfig& cfg)`` because of the parameter. I noticed that everything works fine if I remove that entry from Passes.td so I did. Same story for OneAPI :blush: 
* Documentation update - yes, it's in the pipeline ;-)
* The MatMul kernel not only prints that ToDo but also has other issues (include file from sample code, not using transa/transb)
* I just noticed - test case is missing too :angel:

corepointer:
Will more thoroughly integrate this for the next release -> switching milestones

Issue #210, opened 2022-03-08
File meta data in JSON


daphne-eu:
In GitLab by @pdamme on Mar 8, 2022, 17:46

DAPHNE stores some essential meta data about a file in a *meta data file* (see `src/runtime/local/io/FileMetaData.h`). Those meta data files have a very simple text-based format. However, its semantics is still somewhat cryptic. It would be great to make it more explicit and user-friendly by using a simple JSON-based format.

For instance, the following current meta data file:
```
2,2,0,si64,f64,foo,bar
```
Could become something like:
```
{
  "numRows": 2,
  "numCols": 2,
  "columns": [
    {"name": "foo", "valueType": "si64"},
    {"name": "bar", "valueType": "f64"}
  ]
}
```

A JSON parser library should soon be added via #141.

matejmoravec:
Hello. I would like to work on this issue. :-)

pdamme:
For clarification: Since `FileMetaData` can represent both matrix and frame meta data (which are slightly different), it's not so trivial how to design the JSON-based format. I propose the following:

**Schema of a meta data JSON file**

The JSON file should always contain the following keys:
- `numRows`: integer
- `numCols`: integer

The JSON file should always contain exactly one of the following keys:
- either: `valueType`: string
  Then, `FileMetaData::schema` has one element, `FileMetaData::labels` is empty, and `FileMetaData::isSingleValueType` is `true`.
- or: `schema`: array of objects (`label`: string, `valueType`: string)
  Then, `FileMetaData::schema` and `FileMetaData::labels` are filled with the provided value types/labels, and `FileMetaData::isSingleValueType` is `false`.

The JSON file may optionally contain the following key:
- `numNonZeros`: integer (default: `-1`)

**Examples**

Matrix meta data:

```json
{
  "numRows": 7,
  "numCols": 5,
  "valueType": "f64",
  "numNonZeros": 29
}
```

Frame meta data:

```json
{
  "numRows": 5,
  "numCols": 3,
  "schema": [
    {"label": "foo", "valueType": "si64"},
    {"label": "bar", "valueType": "f64"},
    {"label": "baz", "valueType": "f32"}
  ]
}
```

PR #373, opened 2022-05-20
File meta data in JSON

matejmoravec:
- Added parser for reading meta data from a simple JSON-based format
- A few unit tests for the newly added meta data parser have been added

This pull request addresses issue #210.


matejmoravec:
Just one concern: I had to add a new value `INVALID` into the enum class [`ValueTypeCode`](https://github.com/matejmoravec/daphne/blob/210-file-meta-data/src/runtime/local/datastructures/ValueTypeCode.h). The script [build.sh,](https://github.com/daphne-eu/daphne/blob/main/build.sh) now shows many warnings everywhere where the enum class is used, usually because of not checking the newly added value. Maybe any recommendation on what to do here?

The next thing is that somewhere, "metadata" is used as one word and elsewhere as two words, "meta data." Maybe this should be unified in the future...

matejmoravec:
> Just one concern: I had to add a new value `INVALID` into the enum class [`ValueTypeCode`](https://github.com/matejmoravec/daphne/blob/210-file-meta-data/src/runtime/local/datastructures/ValueTypeCode.h). The script [build.sh,](https://github.com/daphne-eu/daphne/blob/main/build.sh) now shows many warnings everywhere where the enum class is used, usually because of not checking the newly added value. Maybe any recommendation on what to do here?
> 
> The next thing is that somewhere, "metadata" is used as one word and elsewhere as two words, "meta data." Maybe this should be unified in the future...

The alternative solution is that I can set a default value from one of the existing (`SI8`, `SI32`, `SI64`, `UI8`, `UI32`, `UI64`, `F32`, or `F64`) in case the JSON format contains the unknown value for `ValueTypeCode`. In that case, I don't have to add a new value to the enum class.

corepointer:
Hi!
I did not fully test this feature, but so far it LGTM. The change in terminology (metadata vs meta data) in comments can be easily addressed any time. The issue with warnings about the newly introduced INVALID enum can be fixed by either adding that to the switch statements that use ValueTypeCode or adding a default case. My compiler (gcc 11.2.0) found five occurrences where this is needed in CastObj.h, EwBinaryObjSca.h, ReadCsvFile.h, ReadDaphne.h and WriteDaphne.h. I silenced the warnings by putting this at the end of the switch statement: ```default: throw std::runtime_error("unknown value type code");```

corepointer:
LGTM - Test suite runs fine. Merging with minor changes.

I found another warning of a switch statement in CheckEqApprox.h (noticable when running test.sh) and I beautified the inclusion of json.hpp ;-)

matejmoravec:
> LGTM - Thank you for your contribution! 👍
> 
> #### Test suite runs fine. Merging with minor changes:
> * I found another warning of a switch statement in CheckEqApprox.h (noticable when running test.sh)
> * I beautified the inclusion of json.hpp ;-)

Great! Thank you.

Issue #230, opened 2022-03-18
String as value type of DenseMatrix


daphne-eu:
In GitLab by @pdamme on Mar 18, 2022, 11:42

So far, DAPHNE supports strings only for scalars, but not for the values in a matrix or frame. However, data sets with string columns are commonplace and play an important role in both machine learning and database applications. Therefore, we must support those.

The task is to implement (C++) a specialization of `DenseMatrix` for the string value type. The major challenge is that each value could have its individual length. Instead of storing each value as a separately allocated string, a more compact and cache-friendly structure is desired. For instance, one could store the concatenation of all string values in one large buffer, and make the entries of the matrix pointers into that buffer. The data structure shall implement the typical interface of a matrix in DAPHNE, including `get`/`set`/`append` for individual values, direct access to the underlying buffers, and the creation of (zero-copy) views (row/column segments) into the data.

In addition to the data representation itself, a handful of important operations on a matrix of strings shall be implemented, whereby the operations can be chosen depending on personal interests and team size. Interesting candidates include:
- elementwise comparisons (to another matrix and/or scalar): (in)equality, lexicographical order, SQL-style `LIKE` with wildcard `%`
- dictionary encoding (assigning unique numerical codes to distinct strings in a matrix)
- elementwise concatenation
- reading/writing string columns in CSV files (extension of the existing CSV reader/writer)
- ...

akroviakov:
Assigned to @akroviakov.

pdamme:
@akroviakov: I recommend specializing for `DenseMatrix<const char *>`, since we use `const char *` (not `std::string`) to represent string scalars at the moment. But if that turns out to be difficult, we can talk about it.

PR #381, opened 2022-05-24
[DAPHNE-#230]:  String as value type of DenseMatrix

akroviakov:
- Implements `DenseMatrix` template specification for `const char*`.
- Additional argument `strBufCapacity` is introduced to allow user-defined initial strings buffer capacity.
- A small `CharBuf` struct is introduced and used for storing and managing `char *` buffer of **null-terminated** strings.
- `DenseMatrix` for strings is built around two arrays:
- - `strings` - Character array stored in `CharBuf`.
- - `values` - A `DenseMatrix` local array with `const char *` offsets in `strings`.
- `DenseMatrix` constructor sets all strings to `\0` by default to avoid issues regarding null-termination.
- Supports `set()`: memory in `strings` is shifted and `values` are adjusted according to `diff` between strings.
- Supports `append()`: any previous content gets simply overwritten.
- Supports `get()`: no changes, a pointer to null-terminated string is returned.
- Supports views: `CharBuf` is shared, so changes are propagated accordingly.
- - Note: `CharBuf` has to remember the number of cells in the original `DenseMatrix` to properly handle memory shifts when using views.
- Overruns are handled by expanding `strings` to either double the storage or, if more required, exactly to the required size, the content is copied over to the new storage, the old pointer is deleted.
- Changed asserts in `pos()` to exceptions to actually report an issue.

### DIFF
--- a/src/runtime/local/datagen/GenGivenVals.h
+++ b/src/runtime/local/datagen/GenGivenVals.h
@@ -93,6 +93,22 @@ struct GenGivenVals<DenseMatrix<VT>> {
     }
 };
 
+template<>
+struct GenGivenVals<DenseMatrix<const char*>> {
+    static DenseMatrix<const char*> * generate(size_t numRows, const std::vector<const char*> & elements, size_t minNumNonZeros = 0) {
+        const size_t numCells = elements.size();
+        assert((numCells % numRows == 0) && "number of given data elements must be divisible by given number of rows");
+        const size_t numCols = numCells / numRows;
+        auto res = DataObjectFactory::create<DenseMatrix<const char*>>(numRows, numCols, false);
+        res->prepareAppend();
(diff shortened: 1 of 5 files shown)


pdamme (changes requested):
Thanks @akroviakov. It looks good to me already. I've rebased the branch and added a minor fix. As just discussed, please just adapt the test cases such that they don't rely on `genGivenVals` and `==` on matrices, since those actually rely on the code we want to test ;) . After that, I will merge it in.

Regarding the code duplication between `DenseMatrix<VT>` and `DenseMatrix<const char *>`, we can think of improvements later.

pdamme (approved):
Thanks @akroviakov for the updated test cases. I fixed a small bug and now it looks good to me. Looking forward to operations on string matrices :) .

Issue #352, opened 2022-05-06
DaphneDSL: Integer matrix multiplication


RetoKrummenacher:
Matrix multiplication with dense matrices filled with integers (int64_t) is not working.

```
i1 = reshape(seq(1,5,1),1,5);
print(i1);
i2 = reshape(seq(1,5,1),5,1);
print(i2);
print(i1 @ i2);
```
Output:
![grafik](https://user-images.githubusercontent.com/71382164/167213807-eb231d38-6eb4-4e0f-b533-e4dede3d81a0.png)

pdamme:
Thanks for bringing this to our attention, @RetoKrummenacher. Currently, we only have matrix multiplication kernels for `f32` and `f64` value types (based on BLAS routines). We need either a general implementation of matrix multiplication for a generic value type, or specializations for typical integer value types.

PR #384, opened 2022-05-25
Integer matrix multiplication #352

KostasBitsakos:


corepointer:
This solution is not ideal with the copies and casts. This affects performance and precision. I have another version in the pipeline that should fix this. I'll honor your efforts via co-authored-by, which should give you the GitHub credit.

Issue #364, opened 2022-05-15
Type promotion in DaphneDSL


pdamme:
DaphneDSL still has several limitations regarding the combination of values of different data/value types. I'm going to address them in roughly the following steps:
- Type inference shall support type promotion (e.g., `si64` plus `f64` shall be `f64`, `matrix<si64>` plus `f32` shall be `matrix<fp32>`, etc.). For that purpose, type inference shall be based on traits for typical cases (as an alternative to the existing interface), akin to shape inference.
- A compiler pass harmonizing argument types of operations, to avoid an exponential explosion of the kernel instantiations to generate, while still being (becoming) able to process any argument type combination. For instance, `si64` plus `f64` shall cast the first argument to `f64`.
- More flexible instantiation of kernels with op codes. It should be possible to specify different type combinations (result, arguments) for different sets of op codes.

This should also fix some other issues:
- #355
- #353, #354 (at least the problem with loops)

PR #387, opened 2022-05-31
[DAPHNE-#358]: DaphneDSL: Matrix literals

### LINKED ISSUE #358: DaphneDSL: Matrix literals
Presently, it is not possible to generate a matrix with desired values. Like a magic square with predefined values at desired position. Currently, such a matrix needs to be created with a series of ```fill```, ```cbind``` and ```rbind``` operations. 

There are several possible ways to provide this:
1. Similar to Matlab by just writing values in brackets
2. Similar to R by creating a one dimensional matrix and reshaping it to the desired dimensions.
```
# matlab
m = [17 24 1 8 15, 23 57 14 16];
# R
v = c(17 2 4 1 8 15 23 5 7 14 16);
m = reshape(m,5,2)
# or with matrix command
m = matrix(17 2 4 1 8 15 23 5 7 14 16); 
```

This would be useful for example when solving given small systems of linear equations. Then the matrix can be created directly in DaphneDSL instead of importing it from a file.

akroviakov:
Implements Matrix literal for DaphneDSL.

- A light `bool` type support for `DenseMatrix` is added (for print and reshape kernels).
- Creates a `DenseMatrix` with shape (nx1) of type `int64_t`,  `double` or `bool` inside `DaphneDSLVisitor`.
- Usage example: `x = [1, 2, 3]` or `x = [1.1, 2.2, 3.3]` or `x = [true, false, true]`.
- Throws error in case of:
- - You try to create a matrix with mixed types (e.g., `x = [1, 2.2, 3]`).
- - You try to create an empty matrix: can't infer the type.
- - You use a not yet defined in `ValueTypeUtils` type (e.g., `const char*`).

Closes #358 .

### DIFF
--- a/src/ir/daphneir/DaphneOps.td
+++ b/src/ir/daphneir/DaphneOps.td
@@ -72,6 +72,11 @@ def Daphne_FillOp : Daphne_Op<"fill", [
     let results = (outs Matrix:$res);
 }
 
+def Daphne_MatrixConstantOp : Daphne_Op<"matrixConstant">{
+    let arguments = (ins UI64:$matrixAddr);
+    let results = (outs Matrix:$res);
+}
+
 def Daphne_CreateFrameOp : Daphne_Op<"createFrame", [
     SameVariadicOperandSize,
     DeclareOpInterfaceMethods<InferFrameLabelsOpInterface>,
--- a/src/parser/daphnedsl/CMakeLists.txt
+++ b/src/parser/daphnedsl/CMakeLists.txt
@@ -30,5 +30,9 @@ add_library(DaphneDSLParser STATIC
 )
 
 target_link_libraries(DaphneDSLParser PRIVATE
+        DataStructures
+)
+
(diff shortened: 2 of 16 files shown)


pdamme (approved):
Thanks a lot, @akroviakov, it **looks very good to me**. I've applied just a few minor changes for polishing. Matrix literals are a **very useful feature**. Having a basic support for them will simplify writing DaphneDSL scripts a lot :) .

I can imagine several follow-up issues relaxing some current constraints, e.g.:
- Empty matrix literals could just assume `double` as the value type. The user can always cast that to any other value type like `as.si64([])` (which will be zero-cost).
- Allowing mixed element types by infering the common value type of the matrix via type promotion (see #364).
- Allowing complex expressions for the elements of a matrix literal. All elements might become known at compile-time through constant folding, then we could create a `MatrixConstantOp` in a compiler pass. However, when not all elements are known at compile-time, we could create DaphneIR ops to calculate them at runtime (will look like a matrix literal in DaphneDSL, but be executed differently).

Issue #385, opened 2022-05-26
Connect JSON-based file meta data parser to file IO


pdamme:
Since recently (dcf90639baff66394fbfbc2267c7a1a965223796), we have a parser for a JSON-based file meta data file format. Currently, this is still disconnected from the rest of the code base. 

**What needs to be done**
1. `FileMetaData::ofFile()` should use the new `MetaDataParser` internally.
2. `FileMetaData::toFile()` should create a JSON file that can be read again by `MetaDataParser`.
3. Existing meta data files in the test cases (primarily in `test/runtime/local/io/`) need to be converted to the new JSON-based format.

**Hints**
- I recommend keeping the interface of `FileMetaData` (and not marking it deprecated). That way, we don't need to modify all places where it is used in the compiler and runtime.
- The old meta data file format does not need to be supported at all anymore and can be removed altogether.

matejmoravec:
Hello. I would like to work on this issue. :-)

PR #394, opened 2022-06-03
[DAPHNE-#385] Connected JSON-based file meta data parser to file IO

matejmoravec:
- Reading and writing meta data moved from the file [FileMetaData.h](https://github.com/daphne-eu/daphne/blob/main/src/runtime/local/io/FileMetaData.h) to [MetaDataParser.h](https://github.com/daphne-eu/daphne/blob/main/src/parser/metadata/MetaDataParser.h).
- Call-sites of FileMetaData::toFile/ofFile replaced with MetaDataParser::writeMetaData/readMetaData.
- Existing meta data files in the test cases converted to the new JSON-based format.

This pull request addresses issue #385.


matejmoravec:
Just two things:

- In the current implementation of writing the meta data, I assume that the number of values in the `schema` vector and values in the `labels` vector will always be the same. Maybe should I check this before writing meta data to the file and throw an exception if it's not true?
- The next thing is that when I read the `label` JSON key that is nested inside the `schema`, this JSON key is mandatory when reading the file with meta data. I can set it to optional if it would be more appropriate.

corepointer (approved):
LGTM - Thank you for your contribution @matejmoravec 
I will merge this with minor modifications:
* cleanups
* silencing of warnings
* another test case
* documentation

Issue #396, opened 2022-06-03
DaphneDSL: Built-in functions var() not working


RetoKrummenacher:
The built-in function var() is not working in DaphneDSL:
```
m = fill(2.0,5,2);
print(var(m)); # overall
print(var(m, 1)); # cloumn wise
print(var(m, 0)); # row wise
```

![grafik](https://user-images.githubusercontent.com/71382164/171903542-c98ba4ef-e964-4e73-a6a0-91a07ea1c88d.png)

Issue #395, opened 2022-06-03
DaphneDSL: Built-in functions stddev() only working column wise


RetoKrummenacher:
The built-in function stddev() is working only column wise:
```
m = fill(2.0,5,2);
print(stddev(m, 1)); 
```
![grafik](https://user-images.githubusercontent.com/71382164/171902759-a189dd61-8b2e-4705-853d-a8a8d831b754.png)

However, it is not working row wise or for the total matrix:
```
m = fill(2.0,5,2);
print(stddev(m, 0)); # row wise
print(stddev(m)); # overall
```
![grafik](https://user-images.githubusercontent.com/71382164/171903014-8bc978d6-6011-4e42-9c41-7ed908549429.png)

Issue #76, opened 2021-06-11
Extension of `aggCol`-kernel for more aggregation functions


daphne-eu:
In GitLab by @pdamme on Jun 11, 2021, 20:35

So far, the `aggCol`-kernel for column-wise aggregation only supports `SUM`, `MIN,` and `MAX`, which can all be implemented via binary operations on a pair of values. However, we also need `MEAN` and `STDDEV`, which are more involved to implement.

These two new op-codes should be added to `src/runtime/local/kernels/AggOpCode.h` and the kernel implementation should support them.

PR #398, opened 2022-06-03
Support for mean operation for all-aggregation and row-aggregation.

### LINKED ISSUE #393: DaphneDSL: Built-in functions sum and mean
The built-in functions sum and mean are used differently in DaphneDSL. 

Sum() for instance is usable with arguments: 0 = row wise, 1 = column wise, none = all and working fine:

```
m = fill(2.0,5,2);
print(sum(m)); # overall
print(sum(m, 0)); # row wise
print(sum(m, 1)); # column wise
```
image

However, this is different with mean() where only column wise mean is possible.
```
m = fill(2.0,5,2);
# working
print(mean(m, 1)); # cloumn wise
# not working
print(mean(m)); # overall
print(mean(m, 0)); # row wise
```
image

aristotelis96:
- Support for Aggregate-All `mean` for Dense and CSR matrices.
- Support for Aggregate-Row `mean` for Dense and CSR matrices.
- Some testcases.

Closes #393.

### DIFF
--- a/src/runtime/local/kernels/AggAll.h
+++ b/src/runtime/local/kernels/AggAll.h
@@ -59,19 +59,39 @@ struct AggAll<DenseMatrix<VT>> {
         const size_t numCols = arg->getNumCols();
         
         const VT * valuesArg = arg->getValues();
-        
-        assert(AggOpCodeUtils::isPureBinaryReduction(opCode));
-        
-        EwBinaryScaFuncPtr<VT, VT, VT> func = getEwBinaryScaFuncPtr<VT, VT, VT>(AggOpCodeUtils::getBinaryOpCode(opCode));
 
-        VT agg = AggOpCodeUtils::template getNeutral<VT>(opCode);
+        EwBinaryScaFuncPtr<VT, VT, VT> func;
+        VT agg;
+        if (AggOpCodeUtils::isPureBinaryReduction(opCode)) {
+            func = getEwBinaryScaFuncPtr<VT, VT, VT>(AggOpCodeUtils::getBinaryOpCode(opCode));
+            agg = AggOpCodeUtils::template getNeutral<VT>(opCode);
+        }
+        else {
+            // TODO Setting the function pointer yields the correct result.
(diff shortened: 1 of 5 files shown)


aristotelis96:
Note that in this implementation, result value type for `AggAll` (and result value type DenseMatrix for `AggCol`/`AggRow`) is the same value type as the argument matrix. This does not make much sense for `mean` in the case of `int` matrices (which the result will probably be `double`). This kernel could be extended to support `DenseMatrix<int> -> double` as @psomas first suggested at #76. If so, it also makes sense to implement `stddev` #395 and `var` #396  in a similar way.

aristotelis96:
Note that in this implementation, result value type for `AggAll` (and result value type DenseMatrix for `AggCol`/`AggRow`) is the same value type as the argument matrix. This does not make much sense for `mean` in the case of `int` matrices (which the result will probably be `double`). This kernel could be extended to support `DenseMatrix<int> -> double` as @psomas first suggested at #76. If so, it also makes sense to implement `stddev` #395 and `var` #396  in a similar way.

pdamme (approved):
Thanks for reacting so quickly to this issue, @aristotelis96. It **looks good to me**. I fixed just a few minor things:
- `AggAll` and `AggRow` kernels throw a `std::runtime_error` for `STDDEV`, rather than returning a wrong result.
- Bugfix in `AggRow` for `CSRMatrix`: divide by #cols, not by #rows * #cols for `MEAN`.
- Testing `AggRow` also for `CSRMatrix`.

pdamme:
I fully agree regarding the value types. The output of mean/stddev on an integer matrix should be a floating-point type. As suggested in the discussion on #76, we need to make the aggregation kernels more flexible by splitting the template parameter `VT` into `VTRes` and `VTArg`.

Note also that I'm working on type promotion (#364) in branch `364_type-promotion`. Once this is finished, it will allow us to call the aggregation kernels with the right value types from DaphneDSL (assuming the kernels support it by then ;) ).

PR #400, opened 2022-06-06
[DAPHNE-#388] order-kernel for DenseMatrix

### LINKED ISSUE #388: order-kernel for DenseMatrix
We need a specialization of the `order`-kernel for `DenseMatrix`.

At the moment, we need that to make the PCA DaphneDSL script (`scripts/algorithms/pca.daph`) work. Therefore, it is okay for now if the kernel only supports ordering by a single column (but it should throw an exception if sorting by multiple columns is requested). The multi-column case can be handled in a separate issue later.

Hiebl:
* Implements specialization of the order-kernel for DenseMatrix
* Works in the single column and multi-column case
* Slight changes to existing order methods to respect the rowSkip in multi-column matrices
* Builds on the Order Kernel DeduceType Util Refactor from #242
* closes #388


pdamme (approved):
Thanks a lot, @Hiebl. Looks good to me.

Issue #403, opened 2022-06-07
CLI args for explaining/printing the IR at various points


pdamme:
The DAPHNE compiler uses DaphneIR as the internal representation of a program, and performs various compiler passes, which optimize and lower the IR. To better understand the behavior of the DAPHNE compiler and runtime, it is very helpful to look at the IR at various stages, e.g., after property inference or after creating vectorized pipelines. We also call this feature EXPLAIN, in accordance with the well-known feature of database management systems.

To make this feature easily usable, `daphne` should have command-line arguments to turn on the printing of the IR after certain passes. We partially support this already. For instance, the [`DaphneUserConfig`](/daphne-eu/daphne/blob/main/src/api/cli/DaphneUserConfig.h) already contains various `explain_...` flags, and the [`DaphneIrExecutor`](/daphne-eu/daphne/blob/main/src/compiler/execution/DaphneIrExecutor.cpp) queries these flags to decide if a `PrintIRPass` shall be added after the respective stages. However, the flags in the `DaphneUserConfig` still need to be connected to the command-line arguments parsed in [`daphne.cpp`](/daphne-eu/daphne/blob/main/src/api/cli/daphne.cpp). So far, `daphne` only supports `--explain-kernels`, which maps to `DaphneUserConfig::explain_kernels`, and prints the IR after lowering to kernel calls.

**Task**
- This task is to ensure that the printing of the IR after all currently relevant DAPHNE compiler passes can be triggered by command-line arguments.
- In particular, we should support the passes already specified in `DaphneUserConfig`: `kernels`, `llvm`, `parsing`, `property_inference`, `sql`, `vectorized`, `obj_ref_mgnt`.
- There are two good options to implement this feature, please choose one of them:
  1. Add additional separate arguments like `--explain-kernels`, i.e., `--explain-llvm` etc. This should be quite straightforward, but will require the user to type the prefix `--explain` unnecessarily often when the IR shall be printed at multiple stages.
  2. Introduce a single argument `--explain`, which can be used like `--explain pass1,pass2,pass3` or `--explain pass1 pass2 pass3`, i.e., with a comma-separated or space-separated list. For the individual passes, only the values defined above shall be allowed. Since this feature is only about printing the IR (not about configuring the order of passes), the order of this list as well as potential duplicates would be irrelevant. [`cl::bits`](https://llvm.org/docs/CommandLine.html#collecting-options-as-a-set-of-flags) might be useful here.

**Hints**
[...]

PR #404, opened 2022-06-12
[DAPHNE-#403]: CLI args for explaining/printing the IR at various points

philipportner:
Implements simple mapping from CLI arguments to [DaphneUserConfig](https://github.com/daphne-eu/daphne/blob/main/src/api/cli/DaphneUserConfig.h)s various `explain_...` members using a `llvm::cl::list`.

- adds the `ExplainArgs` enum containing all the available `--explain` arguments

### DIFF
--- a/src/api/cli/daphne.cpp
+++ b/src/api/cli/daphne.cpp
@@ -155,12 +155,14 @@ main(int argc, char** argv)
     llvm::cl::list<Explains> explainArgList(
         "explain", cat(daphneOptions),
         llvm::cl::desc("Available explainations:"),
-        llvm::cl::values(clEnumVal(kernels, "KERNELS"), clEnumVal(llvm, "LLVM"),
-                         clEnumVal(parsing, "PARSING"),
-                         clEnumVal(property_inference, "PROPERT INFERENCE"),
-                         clEnumVal(sql, "SQL"),
-                         clEnumVal(vectorized, "VECTORIZED"),
-                         clEnumVal(obj_ref_mgnt, "OBJ REF MGNT")),
+        llvm::cl::values(
+            clEnumVal(kernels, "Show DaphneIR after kernel lowering"),
+            clEnumVal(llvm, "Show DaphneIR after llvm lowering"),
+            clEnumVal(parsing, "Show DaphneIR after parsing"),
+            clEnumVal(property_inference, "Show DaphneIR after property inference"),
+            clEnumVal(sql, "Show DaphneIR after SQL parsing"),
+            clEnumVal(vectorized, "Show DaphneIR after vectorization"),
+            clEnumVal(obj_ref_mgnt, "Show DaphneIR after managing object references")),
         CommaSeparated);
 
     llvm::cl::list<string> scriptArgs1(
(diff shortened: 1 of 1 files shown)


philipportner:
I tried using `llvm::cl::bits` as you suggested @pdamme but this resulted in 2 bits being always set, e.g., `.getBits()` would return `0011 0000` even if no argument was passed.
I tried the example from the documentation and this changed the behaviour in a strange way, the `.getBits()` of our arguments object would be random and the `.getBits()` of the example object would be `0011 0000`.

So instead of the list declaration I had this:
```
llvm::cl::bits<Explains> explainArgBits(
    "explain", llvm::cl::desc("Available explainations:"),
    llvm::cl::values(clEnumVal(kernels, "KERNELS"), clEnumVal(llvm, "LLVM"),
                     clEnumVal(parsing, "PARSING"),
                     clEnumVal(property_inference, "PROPERT INFERENCE"),
                     clEnumVal(sql, "SQL"),
                     clEnumVal(vectorized, "VECTORIZED"),
                     clEnumVal(obj_ref_mgnt, "OBJ REF MGNT")),
    CommaSeparated);
```
and instead of the loop:
```
std::cout << "explainArgBits after parsing: " << explainArgBits.getBits() << std::endl;
user_config.explain_kernels = explainArgBits.isSet(kernels);
user_config.explain_llvm = explainArgBits.isSet(llvm);
user_config.explain_parsing = explainArgBits.isSet(parsing);
user_config.explain_property_inference = explainArgBits.isSet(property_inference);
user_config.explain_sql = explainArgBits.isSet(sql);
user_config.explain_obj_ref_mgnt = explainArgBits.isSet(obj_ref_mgnt);
```

I wasn't able to figure out what caused this strange behaviour, I would have prefered using `cl::bits` to not have the `for`-loop but it seems more reliable.

pdamme (approved):
Thanks a lot, @philipportner. This feature is super useful for debugging the system and for communicating its internals to other people. **Your implementation looks very good to me.** And no worries about the `cl::bits`, using that was just a suggestion, but your approach is totally fine, too.

Just a small thing before we merge it: We often apply some polishing to finalize a PR. Could you allow others to push commits to your PR, please? Or, could you apply the following change yourself (more expressive help message and stages in the same order as they are applied by the DAPHNE compiler)?

```c++
llvm::cl::list<ExplainArgs> explainArgList(
        "explain", cat(daphneOptions),
        llvm::cl::desc("Show DaphneIR after certain compiler passes (separate multiple values by comma, the order is irrelevant)"),
        llvm::cl::values(
                clEnumVal(parsing, "Show DaphneIR after parsing"),
                clEnumVal(sql, "Show DaphneIR after SQL parsing"),
                clEnumVal(property_inference, "Show DaphneIR after property inference"),
                clEnumVal(vectorized, "Show DaphneIR after vectorization"),
                clEnumVal(obj_ref_mgnt, "Show DaphneIR after managing object references"),
                clEnumVal(kernels, "Show DaphneIR after kernel lowering"),
                clEnumVal(llvm, "Show DaphneIR after llvm lowering")
        ),
        CommaSeparated
);
```

Other than that, it's all fine.

pdamme:
Great, thanks!

Issue #200, opened 2022-03-01
Transposition-aware matrix multiplication


daphne-eu:
In GitLab by @pdamme on Mar 1, 2022, 19:24

Matrix multiplications are at the heart of most ML algorithms and can be very expensive in terms of runtime. There are highly optimized routines (e.g., BLAS) for executing matrix multiplications. In many cases, one or both inputs are transposed matrices. Thus, linear algebra libraries like BLAS can efficiently process matrices without materializing the transposed representation. However, DAPHNE's MatMulOp is currently not aware of whether its inputs are transposed. Thus, an expression like `C = t(A) @ B;` would first transpose A before it calculates the matrix multiplication.

The task is to modify the existing MatMulOp such that it has two boolean flags indicating if the left/right-hand side input is transposed. The DaphneDSL parser shall initially create MatMulOps assuming non-transposed inputs, while a new compiler pass shall identify if an input to a MatMulOp is the result of a TransposeOp and rewrite the program accordingly. Finally, the runtime kernels of the MatMulOp must pass this information on transposition to the BLAS kernels they call internally. Implementation in C++.

PR #406, opened 2022-06-17
[DAPHNE-#200] Transposition-aware matrix multiplication

Hiebl:
* Adds a Canonicalization Pass for Transposition-aware matrix multiplication
* Refactors the MatMul-Kernel to utilize transposition information and
* Adds Boolean Flag and Script Level Tests for MatMul


pdamme (changes requested):
Hi @Hiebl, thanks for this contribution! It already **looks good to me** as the core of the required steps (extension of `MatMulOp` and its kernel, canonicalizer for `MatMulOp`, adaptation of the DaphneDSL parser, etc.) has been implemented very well.

As you've already mentioned, there are still **a few limitations, and they should be fixed** before we merge it. However, this shouldn't cause too much effort.

***Regarding "nested" transpose:*** You decided to pass `transa` and `transb`, which indicate if the `lhs` and `rhs` inputs are transposed, as *arguments*. This makes sense, since it integrates well with the lowering to kernel calls. However, it makes the extraction of this information in the compiler slightly more difficult. But we already have some examples for this in the codebase, see, e.g., `DaphneDSLBuiltins::createRowOrColAggOp`. It should be possible in a similar way here. **Please add that to your implementation**. As you've mentioned, passing `transa` and `transb` as *attributes* would also have been possible, but it would have required changes to the way `MatMulOp` is lowered to a kernel call, since we do not automatically pass all attributes as runtime parameters to kernels. Both alternatives would have been fine.

***Regarding the failing test case `kmeans.daphne`:*** In general, when one of the script-level test cases fails, I recommend executing the respective DaphneDSL script directly (using the same arguments as the test case). So in this case:
```bash
build/bin/daphne test/api/cli/algorithms/kmeans.daphne r=100 c=5 f=20 i=10
```
It prints the error:
```
Pass error: the type/shape of a variable must not be changed within the body of a for-loop
```
That's strange and shouldn't happen, since actually, the shape does not change in the for-loop in this algorithm. You can have a closer look at the IR by (okay, due to some limitations we need to comment out the check in InferencePass.cpp lines 236-240 to make that work...):
```bash
build/bin/daphne --explain property_inference test/api/cli/algorithms/kmeans.daphne r=100 c=5 f=20 i=10
```
We can see that the infered shape after the `MatMulOp` is wrong:
```
    %11 = "daphne.constant"() {value = true} : () -> i1
    %12 = "daphne.constant"() {value = false} : () -> i1
...
        %20 = "daphne.matMul"(%13, %arg1, %12, %11) : (!daphne.Matrix<100x20xf64:sp[1.000000e+00]>, !daphne.Matrix<5x20xf64>, i1, i1) -> !daphne.Matrix<100x20xf64>
```
The compiler infers 100x20, but 100x5 would be correct. That's because the shape inference of `MatMulOp` is determined by the traits `NumRowsFromIthArg<0>, NumColsFromIthArg<1>` (see `DaphneOps.td` and `DaphneInferShapeOpInterface.h`). This is no longer valid, since the inputs could be transposed. **Please fix this** by not using shape inference traits but implementing the shape inference interface for `MatMulOp` (see `CartesianOp` for a simple example, and `DaphneInferShapeOpInferface.cpp` line 155ff). Furthermore, `MatMulOp::createOpsOutputSizes` (in `DaphneVectorizableOpInterface.cpp` also needs to be adapted).

***Regarding the failing test case `runMatMult.daphne`:*** The error message indicated that something was expected to be a `VectorSplit`, but wasn't. `MatMulOp` is a vectorizable operation, i.e., we can split its inputs and process them in parallel. Each vectorizable operation must provide some information relevant for creating vectorized pipelines by implementing `VectorizableOpInterface`. In particular, we need to specify if/how the new arguments `transa` and `transb` can be split. **I've already fixed this (see my commit) and the test passes now.** However, I needed to fix a bug in the existing `LowerToLLVMPass` to make that work.

Note that I have rebased your branch on the main branch.

pdamme (changes requested):
Thanks for the update, @Hiebl. It looks very good to me now. I just added some minor polishing. Regarding the DAPHNE compiler part and the CPU/host kernel, it is **absolutely ready to be merged**, and you clearly pass the AMLS project with that.

However, there is **one thing that should be addressed** before we can really merge it in: In `kernels.json` you removed the instantiation of the CUDA kernels for matrix multiplication, but we still need them to work. It would be great if you could:
1. Undo that change in `kernels.json`.
2. Adapt the respective CUDA kernel (`src/runtime/local/kernels/CUDA/MatMul.h/cpp`) accordingly (and potentially its test cases). This should not require being an expert in CUDA. The kernel is based on cuBLAS, which also seems to support the concept of specifying if the inputs are transposed (looking at lines like `cusparseOperation_t opA = CUSPARSE_OPERATION_NON_TRANSPOSE;`).

@corepointer: Is there a hint you could give @Hiebl? You don't need to read the entire code in this PR, the most important part for you to know is that the `MatMul`-kernel has two additional `bool` parameters specifying if the lhs/rhs input is transposed. Maybe you could tell @Hiebl how to run the CUDA test cases, possibly even on alpha where CUDA is set up already (feel free to give him access there).

Issue #198, opened 2022-03-01
Second-order functions in DaphneDSL: map()


daphne-eu:
In GitLab by @pdamme on Mar 1, 2022, 19:19

While DaphneDSL provides a multitude of elementwise unary and binary operations on matrices and frames, users could still have very specific needs that are not fulfilled yet. Thus, we would like to offer a map() function (as typically known from functional languages), which takes a matrix and a user-defined function (UDF) as input and applies the UDF to each element of the matrix.

This task will touch multiple layers of the DAPHNE system, starting from the DaphneDSL parser, where it must be possible to pass a UDF as an argument, over the internal representation and compiler, where a new map-operation must be created along with appropriate lowering passes, down to the runtime, where a generic map-kernel must be provided for the actual execution. Implementation in C++.

This task can be scaled depending on the team size, from only elementwise unary functions on dense matrices to unary/binary functions and row/column-wise aggregation on dense/sparse matrices and frames.

PR #407, opened 2022-06-17
WIP: 198 second order function map

sigmaeon:
# Implementation for #198
This PR  implements the *map*-operation as described in #198 for the most basic case. 
That is, an element-wise application of a UDF over a dense matrix.

### DIFF
--- a/src/compiler/lowering/SpecializeGenericFunctionsPass.cpp
+++ b/src/compiler/lowering/SpecializeGenericFunctionsPass.cpp
@@ -179,16 +179,17 @@ namespace {
 
         /**
          * @brief Try to reuse an existing specialization for the given template function
-         * @param callOp The call operation
-         * @param templateFunction The template function called by the `callOp`
+         * @param operandTypes Operand types of the operation
+         * @param templateFunction The template function called by the operation
          * @return either an existing and matching `FuncOp`, `nullptr` otherwise
          */
-        FuncOp tryReuseExistingSpecialization(daphne::GenericCallOp callOp, FuncOp templateFunction) {
+        // TODO_198 Replace callOp by result of callOp->getOperandTypes() to also use it for mapOp
+        FuncOp tryReuseExistingSpecialization(TypeRange operandTypes, FuncOp templateFunction) {
             auto eqIt = specializedVersions.equal_range(templateFunction.sym_name().str());
             for(auto it = eqIt.first ; it != eqIt.second ; ++it) {
                 auto specializedFunc = it->second;
 
-                if(callTypesMatchFunctionTypes(specializedFunc.getType(), callOp->getOperandTypes())) {
+                if(callTypesMatchFunctionTypes(specializedFunc.getType(), operandTypes)) {
                     // reuse existing specialized function
                     return specializedFunc;
                 }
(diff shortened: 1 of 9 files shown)


src/parser/daphnedsl/DaphneDSLVisitor.cpp
```diff
@@ -643,52 +644,137 @@ antlrcpp::Any DaphneDSLVisitor::visitParanthesesExpr(DaphneDSLGrammarParser::Par
...
+    // Create DaphneIR operation for the built-in function.
+    return builtins.build(loc, func, args);
+}
+
+
+antlrcpp::Any DaphneDSLVisitor::visitCallExpr(DaphneDSLGrammarParser::CallExprContext * ctx) {
+    std::string func = ctx->func->getText();
+    mlir::Location loc = utils.getLoc(ctx->start);
+ 
+    if (func == "map")
```
sigmaeon, line 760: As discussed, we need to give function `map` an extra treatment here since the second parameter is a UDF identifier and `visitIdentifierExpr` only searches the `symbolTable` which does not include UDFs.

src/runtime/local/kernels/Map.h
```diff
@@ -0,0 +1,76 @@
...
+
+#include <cassert>
+
+// ****************************************************************************
+// Struct for partial template specialization
+// ****************************************************************************
+
+template<class DTRes, class DTArg>
+struct Map {
+    static void apply(DTRes *& res, const DTArg * arg , void* func, DCTX(ctx)) = delete;
```
sigmaeon, line 34: We could have a more specialized function pointer here i.e. `(DTRes::VT)(*func)(DTArg::VT)`. 
Problem is that this is currently not supported by *kernels.json*.

src/compiler/lowering/SpecializeGenericFunctionsPass.cpp
```diff
@@ -179,40 +179,56 @@ namespace {
...
+                specializedFunc = createSpecializedFunction(calledFunction, specializedTypes);
+            }
+            return specializedFunc;
+        }
+
+        /**
+         * @brief Recursively specializes all functions within a `FuncOp` based on calls to the functions
+         * @param function The `FuncOp` to scan for function specializations
+         */
         void specializeCallsInFunction(FuncOp function) {
```
sigmaeon, line 222: @pdamme: To resolve the question left open in our discussion: `specializeCallsInFunction` does not introduce any casts nor does any mechanism for calls to UDFs. Hence, since `y` is of type `si64`  the following would also not compile:

```
def testfunc(i: ui64)
{
    x = i + 2;
    return x;
}

y = 10;
z = testfunc(y);

```

src/parser/daphnedsl/DaphneDSLVisitor.cpp
```diff
@@ -643,52 +644,137 @@ antlrcpp::Any DaphneDSLVisitor::visitParanthesesExpr(DaphneDSLGrammarParser::Par
...
+    return std::nullopt;
+}
+
+antlrcpp::Any DaphneDSLVisitor::handleMapOpCall(DaphneDSLGrammarParser::CallExprContext * ctx) {
+    std::string func = ctx->func->getText();
+    assert(func == "map" && "Called 'handleMapOpCall' for another function");
+    mlir::Location loc = utils.getLoc(ctx->start);
+    
+    if (ctx->expr().size() != 2) {
+      throw std::runtime_error(
```
sigmaeon, line 726: Note: The exception text is a copy of the exception thrown in `checkArgsExact`.

src/compiler/lowering/SpecializeGenericFunctionsPass.cpp
```diff
@@ -223,6 +239,46 @@ namespace {
                     specializeCallsInFunction(calledFunction);
                 }
             });
+
+            // Specialize all functions called by MapOp
+            function.walk([&](daphne::MapOp mapOp) {
+                auto calledFunction = functions[mapOp.func().str()];
```
sigmaeon, line 245: `functions` contains `FuncOp` by their symbol names (i.e. the ones created by the parser). To look for other `FuncOps`with the same identifier (e.g. for the purpose mentioned in the remark below), one would need to implement an appropriate search through the keys of `functions`.
The problem then is, that there might be different `FuncOp`s that fit the specialization and one would need to decide which one to prefer.

pdamme:
Hi @sigmaeon, sorry for the long delay! But now I've finally had a close look at your PR. Your contribution **looks very good to me**, and I can confirm that it **works very well**. It **adds a very valuable feature to DaphneDSL**.

I have locally brought your branch up-to-date with the main branch, fixed a few conflicts that arised, applied some minor polishing, and added a few script-level test cases (to verify that `map()` works end-to-end from parser over compiler to runtime). With these additions, the PR would be ready to be merged. **Unfortunately, I don't have permissions to push to your branch. I would appreciate if you could grant this permission** (you should see a checkbox on the PR). Then the PR would be marked as "merged". Otherwise, I would just squash your contributions and my additions and commit them on the main branch in your name, but then this PR would be marked as "closed", which doesn't sound nice.

If you get a chance, please let me know your preferred option in the next days. Otherwise, it's also fine and I will just go the second route. In any case, you'll get the credit (in the git history) for this contribution.

Thanks again for contributing this feature!

src/parser/daphnedsl/DaphneDSLVisitor.cpp
pdamme: The current implementation is totally fine :) .

src/runtime/local/kernels/Map.h
pdamme: That's a good point. I copied your comment into the source code (still locally with me), where it is more likely to be found again. From my point of view, it's not urgent, but maybe we will address it at some point.

src/compiler/lowering/SpecializeGenericFunctionsPass.cpp
pdamme: Thanks for the information. We will probably rethink a few characteristics of UDFs in DAPHNE anyway, maybe we will add automatic casts of function parameters if they have a fully specified type. That behavior would be consistent with built-in functions. For now, it's fine as it is.

src/parser/daphnedsl/DaphneDSLVisitor.cpp
pdamme: Totally fine!

sigmaeon:
Hi @pdamme, thanks for the review! I've enabled edits by maintainers on this PR - pushing to the branch should work now.

pdamme:
Thanks for the quick response. Unfortunately, I still get the error `remote: Permission to sigmaeon/daphne.git denied to pdamme.` when trying to push to your branch. Could you double-check the permission, please? Otherwise, I will just use the second approach mentioned above :) .

pdamme (approved):
Thanks again for the very good work! After bringing it up-to-date with the current main branch and applying some polishing (as mentioned above), I approve this contribution.

PR #408, opened 2022-06-21
[DAPHNE-#55]: Casts in DaphneDSL

### LINKED ISSUE #55: Casts in DaphneDSL
In GitLab by @pdamme on Jun 11, 2021, 17:20

We need to be able to cast/convert between data objects and scalars of different types in DaphneDSL.

For instance:
- between different data objects (e.g. matrix and frame)
- between different value types of the same data type (e.g. matrix of float and matrix of int32)
- between data objects and scalars (e.g. making a scalar a 1x1 matrix, or making a 1x1 matrix a scalar)
- between scalars of different types

DaphneDSL needs a general syntax for that. It could be as simple as having built-in functions like `as.matrix(...)`. In any case, the result type of the DaphneIR `CastOp` to create must be reflected somehow, since it cannot be inferred from the arguments, of course.

akroviakov:
Implements following casts in DaphneDSL:
* **DataType specified**
* *  allows casts to some data type (`matrix` | `frame` | `scalar` ) in form of e.g., `as.matrix(x)`.
* **ValueType specified**
* *  allows casts to some value type in form of e.g., `as.si64(x)`.
* **DataType and ValueType specified**
* * allows casts to some data type of some value type in form of e.g., `as.matrix<si64>(x)`.

Minor notes:
- Changed `func == "frame"` in  `src/parser/daphnedsl/DaphneDSLBuiltins.cpp` to `func == "createFrame"`
- Added data types in `src/parser/daphnedsl/DaphneDSLGrammar.g4`, now it is `('matrix' | 'scalar' | 'frame') `

Limitations: 
- As discussed in #55 , the cast implementations of types "ValueType specified" and "DataType and ValueType specified" for `frame` are postponed.

Closes #55.


pdamme (approved):
Thanks for this feature, @akroviakov. Overall, it looks good to me. I have applied some smaller fixes. For instance, as casts to frames of particular column types are not supported by the kernels, the parser already throws an error on such attempts.

Not supporting that yet is okay for now. I've created #438 as a follow-up issue.

Other than that, I see only one fundamental issue that we need to fix later: Currently, underspecified casts assume that the data/value type of the input is known at parse-time. That will not always be the case. I opened #439 as a follow-up issue.

Issue #214, opened 2022-03-08
Import mechanism in DaphneDSL


daphne-eu:
In GitLab by @pdamme on Mar 8, 2022, 18:20

Like many other script/programming languages, DaphneDSL should have some kind of import/include mechanism enabling users to employ code in other files. This is necessary for both, user files and DaphneDSL standard lib files.

Open questions include how exactly the file to include shall be specified and how exactly it will be included.

pdamme:
This issue has been closed a while ago with #327.

pdamme:
This issue has been closed a while ago with #327.

pdamme:
We need this feature to be able to reuse DaphneDSL code in other DaphneDSL scripts. For instance, we want to create a DAPHNE standard library of high-level data analysis and ML functions, which are in turn implemented as functions in DaphneDSL. Then, users should be able to simply import those library functions and use them in their own DaphneDSL scripts.

It could look like this:
```
import "kmeans.daphne";
in = readMatrix("someFile.csv");
print(kmeans(in, 5, 20));
```

The particulars of this feature are yet to be designed. Here is some inspiration:
- Importing a DaphneDSL script file should have roughly the same effect as pasting its contents literally.
- An exception to this could be that function and variable names could be prefixed with the file name, e.g., `foo()` in `bar.daphne` could become `bar.foo()` in the importing script, to avoid name clashes.
- It would be okay to allow import statements only in the beginning of a DaphneDSL script.
- Parser errors should always indicate the actual file where the problem was encountered. So if `main.daphne` imports `bar.daphne` and there is a syntax error in `bar.daphne`, then the error message should report `bar.daphne`.

akroviakov:
Assigned to @akroviakov .

akroviakov:
_`foo()` in `bar.daphne` could become `bar.foo()` in the importing script, to avoid name clashes._
Identifier rule does not allow `.` in the name, could `_` be used as a delimeter for concatenating names with the prefix or must `.` be used?

_Importing a DaphneDSL script file should have roughly the same effect as pasting its contents literally._
The example with kmeans (`test/api/cli/algorithms/kmeans.daphne`) requires command line arguments, so I assume a function (i.e., `def km(r:si64,c:si64,f:si64,i:si64)`) was meant or am I wrong?

akroviakov:
I'm trying to write unit test cases that use information from `UserConfig.json`.
An illustrative test case for one file (something similar to `ConfigTest.cpp`):
[...]

PR #410, opened 2022-07-06
[DAPHNE-#214]:  Import mechanism in DaphneDSL

akroviakov:
Implements simple import mechanism in DaphneDSL.
-----------------------------------
Supports imports from absolute and relative paths, default directories and libraries.
All imported variables and functions are prefixed with a filename or alias.

Example usage: 
```
import "bar.daphne";
import "foo.daphne" as "utils";

print(bar.x);
print(utils.x);
```
-------------------------------------

`UserConfig.json` now has a new field `daphnedsl_import_paths`, which maps e.g., library names to a list of paths, see example:
```
    "daphnedsl_import_paths": 
    {
        "default_dirs": ["test/api/cli/import/sandbox", "some/other/path"],
        "algorithms": ["test/api/cli/import/sandbox/algos"]
    }
```
NOTE: `default_dirs` can hold many paths and it will look for the **one** specified file in each, whereas any other library names have a list consisting of **one** directory,  from which **all** files will be imported (can be easily extended to multiple directories).
Example:
```
import "a.daphne";
import "algorithms";

print(a.x);
print(algorithms.kmeans.someVar);
```
The first import will first check if the relative path exists, then it will look for it relative to paths in `default_dirs`. If the specified file exists for more than one relative path, an error will be thrown.
The second import goes to `algorithms` directory from `UserConfig` and imports all files from it.

Paths from `UserConfig` get to `DaphneDSLVisitor` from `daphne.cpp` via `DaphneUserConfig`. 

-----------------------------------
Variable name collision resolution:
Whenever we stumble upon equal prefixes (e.g., files with the same name in different directories), a parent directory of the file where conflict is detected is prepended before prefix.
Example:
```
import "somedir/a.daphne";
import "otherdir/a.daphne";

print(a.x);
print(otherdir.a.x);
```
NOTE: the parent directory may be prepended even though you never specified it (e.g., the import script is in the same directory as the original script).
Example:
```
import "somedir/a.daphne";
import "a.daphne";

print(a.x);
print(otherdir.a.x);
```

Libraries and aliases:
Currently, the following example is allowed:
```
import "algorithms";
import "sandbox/b.daphne" as "algorithms"; 

print(algorithms.x);
print(algorithms.kmeans1.someVar);
```
Even though both prefixes will begin with `algorithms.`, the entire library content's prefix is extended with filenames. It is up to user to not confuse yourself.

-----------------------------------
Cascade imports:
Any variables/functions imported into the script we are currently importing will be discarded.
Example import scheme: `A<-B<-C`. A imports B, B imports C. B uses some vars/functions from C, but A doesn't "see" any of C's content.

---------------
Minor issue fixed: 
`compareDaphneToRefSimple()` in `test/api/cli/Utils.h` was passing arguments to `compareDaphneToRef()` in a slightly wrong order: expected second argument should be script name, not `args...`.

-------------------

Closes https://github.com/daphne-eu/daphne/issues/214.


corepointer (commented):
LGTM - Thank you for your contribution @akroviakov 

I made a few edits before merging to main. Mostly minor optimizations concerning const refs. I also ported your documentation from the comment section of this PR to a markdown file in the doc directory.

corepointer:
One minor thing: The default paths point to some sub dir of test. This doesn't look nice imho as tests are not so much of a concern for people who want to benefit from a meaningful default. Nevertheless, I left the defaults as they are because the test cases seem to depend on it. I'll file an issue to remind us to fix it at some point.

Issue #389, opened 2022-06-01
order-kernel: return indexes instead of ordered data object


pdamme:
DaphneIR's `OrderOp` and the corresponding kernel have an additional parameter `returnIdxs`. If `returnIdxs` is `false`, the kernel shall return the sorted data object; if `returnIdxs` is `true` it shall return the permutation of the rows that establishes the order (e.g., as a single-column `DenseMatrix`). For instance, the *i*-th value in that vector is *j*, if the sorting would move row *j* of the input to row *i* of the output.

On a feature branch (see #363), this has already been implemented partially, but it needs to be connected to the `returnIdxs` parameter. Besides that, we need to find a good way to integrate it into the overall compilation...

PR #411, opened 2022-07-10
[DAPHNE-#389] order-kernel: return indexes instead of ordered data object

Hiebl:
* implements returning the index matrix from the OrderOp if a boolean flag is set
* shape inference adaptation (to include the case mentioned above)
* builds on #388


pdamme (approved):
Thanks a lot, @Hiebl. It looks good to me. I just added a few things to make sure that we can use the new feature in DaphneDSL.

Issue #412, opened 2022-07-14
Matrix Chain Optimization


Hiebl:
Matrix multiplications are at the heart of most ML algorithms and can be very expensive in terms of runtime. Computational complexity is heavily influneced by how the product is parenthesized. Because the operation is associative, the result obtained will be the same for all possible parenthesized matrix products, if the order of matrices is not change.

For example, for a matrix `X` with dimensions 1000x1000 and a vector 'V' with dimensions 1000x1 the matrix multiplication `Y = (t(X) @ X) @ V;` would result in 2002 MFLOPs of calculation, while `Y = t(X) @ (X @ V);` would result in only 4 MFLOPs of calculation being necessary.

The DaphneDSL parser shall initially create MatMulOps, while a new compiler pass shall identify matrix chains, perform the textbook dynamic programming algorithm to find the minimum cost for the full sequence and rewrite the matrix chain accordingly.

PR #413, opened 2022-07-16
[DAPHNE-#412] Matrix Multiplication Chain Optimization

Hiebl:
* Implements an algorithm to identify matrix chains, find the minimum cost MatMul-Sequence and rewrites the matrix chain accordingly
* Removes redundant transpose during the TranspositionAwareMatMul optimization
* Builds on #200


Hiebl:
TODO:
* programmatically check if we can actually remove the "redundant" operators or if they have use somewhere else 
* extensive testing

pdamme:
Thanks @Hiebl, first of all, for this PR. I hope I will soon find the time to look over it.

Regarding the removal of redundant operations: I guess you refer to the `TransposeOp`s that can become unused due to transposition-aware matrix multiplication. I noticed in #406 that these `TransposeOp`s stay in the code (unsurprisingly) and even get executed at runtime. Of course, it would be great to avoid that and get rid of them, *if they don't have other uses*. I think rather than including such a feature in any specific pass, we should have a dedicated pass removing unused operations (which don't have side-effects, see `NoSideEffects` trait) aka *dead-code elimination*. It seems like MLIR's [CSE pass](https://mlir.llvm.org/docs/Passes/#-cse-eliminate-common-sub-expressions) includes that. Maybe that could be useful to you.

pdamme:
@Hiebl Thanks again for this contribution. I'm going to finalize this feature, so no further action is required from your side. However, *if you could point me to the concrete approach for matrix multiplication chain optimization you implemeted, that would be helpful*, since I'm having some difficulties in mapping the implementation to what I know about matrix multiplication chain optimization.

PR #418, opened 2022-08-09
[DAPHNE-#415]: Kernels for DenseMatrix of strings

### LINKED ISSUE #415: Kernels for DenseMatrix of strings
With the introduction of `DenseMatrix` for strings, we need to support kernels for strings, as well as add new string-specific kernels. Specializing templates for strings may result in _possibly severe_ code duplication. To address this issue, a conditional compilation (of a string-specific branch) within templates is preferred.

One of the main drawbacks in this regard of the current "naive" implementation is the usage of `const char*` to store strings, which requires heap allocation (and subsequent deletion) of temporaries before they can be `set()` in the result `DenseMatrix`.

**Objectives:**
1. Introduce elementwise string-specific kernels:
 1.1. Unary: `ewLower`, `ewUpper`
 1.2. Binary: `ewConcat`, `ewLike`

2. Extend existing kernels (not all of them, just try to find a good way to handle different classes of kernels) to support strings:
2.1. Output of new strings: `fill`, `matrixConstant`
2.2. Only output of existing strings: `transpose`
2.3. No output of strings: elementwise binary comparisons (lexicographical order)
2.4. Output of single string: aggregations

akroviakov:
Implements following operations for `DenseMatrix<const char*>`:
- Element-wise: min/max, uppercase/lowercase, comparisons, concatenation.
- DSL: supports matrix of strings literals; now you can also use `fill` to create a strings matrix.
- `AggRow` and `AggCol` can now be used with strings matrix.
- `Transpose` can be applied on strings matrix, the strings layout in memory is changed as well to keep strings continuous in memory.
- Primitive `Like` kernel.

Implements following operations for `DenseMatrix<const char*>`:
- Element-wise: min/max, uppercase/lowercase, comparisons, concatenation.
- DSL: supports matrix of strings literals; now you can also use `fill` to create a strings matrix.
- `AggRow` and `AggCol` can now be used with strings matrix.
- `Transpose` can be applied on strings matrix, the strings layout in memory is changed as well to keep strings continuous in memory.
- Primitive `Like` kernel (not ewBinary here, see #415 ).

Notes:
As can be seen, supporting strings relies on ` if constexpr(std::is_same_v<VT, const char*>)` kind of compilation branching, which is in many cases much better than duplicating lots of code in template specializations when the difference is one line, for example:
```c
// Fill.h
   ...
        for(size_t r = 0; r < numRows; r++) {
            for(size_t c = 0; c < numCols; c++){
                if constexpr (std::is_same_v<VT, const char*>)
                    res->set(r,c, arg);
                else
                    valuesRes[c] = arg;
            }
            valuesRes += res->getRowSkip();
        }
```
However, such branching can become more complex (see `EwBinarySca.h`). There, we have to repeat `MAKE_CASE` for the same `BinaryOpCode` (MIN/MAX), since we can now have different `VT` for `res` and `lhs/rhs`. 
For example, we want to do `MAKE_CASE(BinaryOpCode::EQ)` on `const char*` matrices and get `bool`/`int` matrix as a return, this is ok, but if we leave `MAKE_CASE(BinaryOpCode::MAX)` in the "general" switch statement, then the compiler will try to compile `MAX` op for `lhs/rhs` of type `const char*` and `res` of type `int`, which makes no sense or is, in other words, **ill-formed**. That's why we now have to condition on the types for proper instantiations. 
So in short, the extensibility in `EwBinarySca.h` on high level can be achieved as simple as conditionally add/constrain switch cases for template instantiations during compile time and have the division of the switch statement based on return/input types. 

Another example of complex branching is `AggRow.h`, where we have `IDXMIN` and `IDXMAX` opcodes implemented in-place before a general binary operations are used. Again, with the introduction of non-numeric `const char*` value type for `DenseMatrix`, we now have to take care and not instantiate templates that do not make sense. As in the example above, `IDXMIN` would expect string matrix and return numeric matrix. This is ok for `IDXMIN` and `IDXMAX` opcodes, but this combination of value types results in ill-formed code, so we branch.

In cases we don't want the code to compile past a certain point (e.g., we are done with calculations and know that the following code is irrelevant for our value type), we can simply do the following:
```c
//AggRow.h
            if constexpr(std::is_same_v<VTArg, const char*>)
                return;
            else{
                // The op-code is either MEAN or STDDEV
               ...
            }

```
__________________________
Suggestion for the following work: consider using `string_view` (non-owning read only access into a null-terminated area in memory) for accessing strings in `CharBuf`, this way we have quick info on string length (instead of current `strlen()` each time we need it), and probably more important, we get access to `std` libraries for our strings, without disrupting continuity of the memory (as it could be the case with `std::string`). 
Closes #415 .

### DIFF
--- a/src/parser/daphnedsl/DaphneDSLVisitor.cpp
+++ b/src/parser/daphnedsl/DaphneDSLVisitor.cpp
@@ -18,7 +18,6 @@
 #include <ir/daphneir/Daphne.h>
 #include <parser/daphnedsl/DaphneDSLVisitor.h>
 #include <parser/ScopedSymbolTable.h>
-#include <runtime/local/datastructures/DenseMatrix.h>
 
 #include "antlr4-runtime.h"
 #include "DaphneDSLGrammarLexer.h"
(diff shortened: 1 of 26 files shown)


pdamme:
I also want to mention **a problem I see with the current solution**. This is no criticism on this PR, just something we need to further think about, and it's not string-specific either.

Some of the existing kernels, which have not been adapted to string matrices yet, cannot be instantiated for `const char *` value type (due to compiler errors); and that's good. However, some of those would not result in compiler errors. For instance, you managed to instantiate the `Reshape`-kernel for `const char *` in `kernels.json` without adapting the kernel code. The problem is that the reshaped result matrix will contain pointers to strings in the input matrix's string buffer, *without explicitly sharing it*. So when the input matrix is cleared away by gargabe collection, the pointers in the result matrix will be dangling. (I don't see a way to trigger that in DaphneDSL right now, since the kernels on string matrices are not fully available there yet, but I think this will become an issue.)

The actual problem is that we cannot prevent anyone from trying to instantiate a generic kernel template for some value type it does not support, algorithmically. The same could happen for, say, complex numbers or other value types we do not support yet.

Anyhow, this is for later. For now it's great to have some decisive operations on string matrices!

pdamme:
Thanks also for **your suggestion regarding `std::string_view`**, @akroviakov. This is a very good point. @corepointer recently also asked why we don't use that. There are two main reasons why we use the old-fashioned `const char *` for passing around strings in DAPHNE: (1) `const char *` is very low-level and, thus, easy to handle between MLIR and C++, and (2) `std::string` (and surely also `std::string_view`) have a certain storage overhead.

Nevertheless, I totally see the point that `string_view` has some great advantages (you mentioned `len`, and it also supports `==` etc. which could simplify elementwise unary ops), and might not be so bad in terms of overhead. We should definitely think about using `string_view`, but I recommend doing so at some point after this PR was merged.

pdamme:
@akroviakov: I've had a look at your latest two commits. It looks largely good to me. There are a few things I'd polish myself. But if you still have some time, it would be great if you could do the following:
- reuse `EwBinarySca`-kernel for `LIKE` in constant folding for `EwLikeOp` instead of rewriting the code
- make `EwLikeOp` an `EwBinaryOp` (or `EwCmpOp`) (see my comment on #415)
- integerate `like` as another option for `cmpExpr` in the DaphneDSL parser (no separate `likeExpr` required)

If you don't have the time anymore, it's also fine, then I'll finalize these things.

Issue #420, opened 2022-08-25
Error in SQLParser.cpp


AhmedEleliemy:
I am trying to compile daphne. However, I am getting the error below from SQLParser.cpp
@pdamme @corepointer any suggestions?? 

In file included from /daphne/src/parser/sql/SQLVisitor.h:19,
                 from /daphne/src/parser/sql/SQLParser.cpp:19:
/daphne/src/parser/ParserUtils.h: In member function 'mlir::Value ParserUtils::valueOrError(antlrcpp::Any)':
/daphne/src/parser/ParserUtils.h:209:14: error: 'using Any = class std::any' {aka 'class std::any'} has no member named 'is'
  209 |         if(a.is<mlir::Value>())
      |              ^~
/daphne/src/parser/ParserUtils.h:209:28: error: expected primary-expression before '>' token
  209 |         if(a.is<mlir::Value>())
      |                            ^
/daphne/src/parser/ParserUtils.h:209:30: error: expected primary-expression before ')' token
  209 |         if(a.is<mlir::Value>())
      |                              ^
/daphne/src/parser/ParserUtils.h:210:22: error: 'using Any = class std::any' {aka 'class std::any'} has no member named 'as'
  210 |             return a.as<mlir::Value>();
      |                      ^~
/daphne/src/parser/ParserUtils.h:210:36: error: expected primary-expression before '>' token
  210 |             return a.as<mlir::Value>();
      |                                    ^
/daphne/src/parser/ParserUtils.h:210:38: error: expected primary-expression before ')' token
  210 |             return a.as<mlir::Value>();
      |                                      ^
/daphne/src/parser/sql/SQLParser.cpp: In member function 'mlir::Value SQLParser::parseStreamFrame(mlir::OpBuilder&, std::istream&, const string&)':
/daphne/src/parser/sql/SQLParser.cpp:52:14: error: 'using Any = class std::any' {aka 'class std::any'} has no member named 'is'
   52 |         if(a.is<mlir::Value>()){
      |              ^~
/daphne/src/parser/sql/SQLParser.cpp:52:28: error: expected primary-expression before '>' token
   52 |         if(a.is<mlir::Value>()){
      |                            ^
/daphne/src/parser/sql/SQLParser.cpp:52:30: error: expected primary-expression before ')' token
   52 |         if(a.is<mlir::Value>()){
      |                              ^
[...]

PR #422, opened 2022-08-25
[DAPHNE-#420]: ANTLR Any Conversion.

DerSchmidt:
ANTLR version 4.9.3 raised an issue with both parsers (daphne and sql).
Both had instances where antlrcpp::Any was implicitly cast to another type, like mlir::Value and mlir::Type.

This should close Issue #420

### DIFF
--- a/src/parser/ParserUtils.h
+++ b/src/parser/ParserUtils.h
@@ -30,74 +30,74 @@
 
 /**
  * @brief General utilities for parsing to DaphneIR.
- * 
- * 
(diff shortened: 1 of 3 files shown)


corepointer:
Thx for this fix. I'll give it a quick test run.

corepointer:
LGTM - thx for this fix. Tried to add some more to it to go directly to 4.10.1 but that didn't work out so well quickly. So merging this in directly.

Issue #397, opened 2022-06-03
Vectorized pipelines with individual result types


pdamme:
At the moment, we require all results of a vectorized pipeline to have the same data type and value type. We should relax this requirement to allow each result to be represented in the most suitable way.

One approach to achieve this is to refactor `MTWrapper` and all related components for the superclass `Structure` (or at least `Matrix`, since we don't support `Frame`s in the vectorized engine yet) as the output type. Special behavior required for a particular data type could be implemented by the data type itself (and be accessible via the `Structure`/`Matrix` interface). This would also facilitate extensibility with new data types.

Issue #390, opened 2022-06-02
Kernel for ctable operation


pdamme:
This task is to implement a kernel for DaphneIR's `CTableOp`. The contingency table of two *(n x 1)* matrices *lhs* and *rhs* is a matrix *res* of shape (*max(lhs) + 1*, *max(rhs) + 1*), and *res[i, j] = |{ k | A[k] = i and B[k] = j, 0 ≤ k ≤ n-1 }|*.

Example:
```
lhs: 2
     4
     5
     4

rhs: 0
     3
     1
     3

res: 0 0 0 0
     0 0 0 0
     1 0 0 0
     0 0 0 0
     0 0 0 2
     0 1 0 0
```

Note that this operation is inspired by SystemDS's `table()` built-in function (see https://apache.github.io/systemds/site/dml-language-reference). However, its definition is slightly different here, since we start counting rows/columns at zero (not at one).

We currently need this kernel to make the PCA DaphneDSL script in `scripts/algorithms/pca.daph` run.

akroviakov:
Yes sorry, my mistake, I've misinterpreted the brackets in the definition.

PR #431, opened 2022-10-06
Distributed runtime documentation.

aristotelis96:
Added two .md files, both related to the use and development of the distributed runtime.
- One documentation file for the Daphne user, how to deploy distributed nodes (workers) and execute a script on the distributed runtime.
- One documentation file for the Daphne developer, explaining the basics on how to extend the existing infrastructure of the distributed runtime.

### DIFF
--- a/doc/DistributedRuntime.md
+++ b/doc/DistributedRuntime.md
@@ -1,0 +1,79 @@
+<!--
+Copyright 2021 The DAPHNE Consortium
+
+Licensed under the Apache License, Version 2.0 (the "License");>
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+# Running DAPHNE on the distributed runtime
+
+## Background
+
+Daphne supports execution in a distributed fashion. Similar to the local vectorized engine, the distributed runtime
+uses multiple distributed nodes (workers) that work on their local data, while a main node, the coordinator, is responsible
+for transferring the data and code to be executed. The user is required to start the workers, either manually or using an 
+HPC tool as SLURM (scripts that start the workers locally or remotely, natively or not, can be found [here](https://github.com/daphne-eu/daphne/tree/main/deploy)). 
+<!-- TODO: add link to documentation. -->
+
+##  Scope
+
+This document focuses on:
+- how to start distributed workers
+- executing Daphne scripts on the distributed runtime
+
+
+## Build the daphne prototype
+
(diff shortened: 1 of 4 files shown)


pdamme:
Thanks @aristotelis96, any documentation is highly welcome! I will give it a read by Tue.

aleszamuda:
Hi, I have updated the issue #437 description in order to explain how documentation for `deploy/` could help addressing the points 1-4 in @pdamme comment https://github.com/daphne-eu/daphne/pull/431#pullrequestreview-1136148570.

aristotelis96:
Thanks @pdamme for reviewing this PR. You are right `--target DistributedWorker` should be mentioned when building the worker, this is now fixed. Regarding your comments:
1. Added. I've also added a "Limitations" section specifing that workers might end up using all their memory, when used for multiple scripts.
2. I agree, this sounds good. Added.
4. @aleszamuda is working on this. After this is completed we can reference a ".md" file here, instead of a directory.
5. Added.
6. Added. @AhmedEleliemy is currently working on MPI support (PR #436). This PR also includes implementation of a new cli argument that enables the user to select a specifing distributed backend. Once #436 is merged this section will be updated.

pdamme:
Thanks @aristotelis96 for the updates. Looks good to me.

I think we could mention more limitations (also to clarify that DAPHNE's distribution capabilities are not fully developed yet), e.g.:
- focus on `DenseMatrix` as the data type (no sparse matrices, no frames)
- focus on `f64` as the value type (no various other value types)

Regarding my initial point 3: Would it be possible to run one of the scripts in `scripts/algorithms/` on the distributed runtime?

aristotelis96:
Thank you for your comments @pdamme. Please note that I've rebased and force-pushed. Adding these limitations is very important. I've also added a few more:

- Distributed runtime and vectorized engine do not support  pipelines with no outputs.
- Distributed runtime and vectorized engine do not support results with individual result types #397.
-  Distributed runtime does not support duplicate pipeline inputs. 

Regarding `scripts/algorithms`, unfortunately none of them can be run on the distributed runtime at the moment. Actually, distributed runtime depends on the vectorized engine (for pipeline creation) and I could not execute any of these scripts with the `--vec` flag:

- components: operand # 4 does not dominate this use (also duplicate pipeline-input error for the distributed runtime)
- gnmf: vectorizedPipelineOp without outputs not supported at the moment!
- lmDS: error: operand # 4 does not dominate this use
- multilogRef: vectorizedPipelineOp without outputs not supported at the moment!
- pca: "tableOp" is missing (however ctable is implemented (#390), are `ctable` and `table` the same?)

pdamme:
Thanks, looks good to me.

I wasn't aware that there are still those problems with the scripts. I thought we had fixed them for all but 1 or 2 scripts... If I find the time, I'll have a look at them.

Regarding the limitations, it's right to mention them, but maybe we should call them "Current Limitations" and clarify that we are going to fix them at some point.

aristotelis96:
Thank you @pdamme. I've renamed the section to "Current limitations" and noted that these will be fixed in the future. 
I've also rebased to main branch.

pdamme (approved):
Thanks, looks good to me!

PR #442, opened 2022-10-15
SQL tutorial

DerSchmidt:
### A Simple SQL Tutorial for release.

**Changes:**
- Made the ";" at the end of SQL Queries Optional to make it more user-friendly! 
- Added a written Tutorial
- Added 3 Examples of varying complexity
- Added CSV and Meta files for one Example and .dict text files to see the encoding of Strings


pdamme:
Thanks @DerSchmidt for this initial write-up. I fixed some minor issues in the text.

I also fixed the problems with the examples as far as I got. Please have a look at my commits for the details. Nevertheless, the examples still don't run smoothly. Note that you should be in directory `daphne/` when you execute the examples, e.g., `/build/bin/daphne doc/tutorial/sqlExample1.daph`.

**Before we can merge it, the following points must be fixed:**
1. Example 2 crashes with a `pass error:`.
2. Example 3 seems to run forever. Maybe that's due to the cross-product? I tried to rephrase the query to use "JOIN ... ON", but couldn't make it pass the SQL parser.
3. Please add the DAPHNE license header to all new files.

**Optionally, it would be great if you could:**
1. Paste the examples into the tutorial file an elaborate a little bit on them (the comments in example 1 could be enough). Just so that the reader can read something without switching to other files. Ideally, create an "Example" section/headline at the end of the document.

pdamme (changes requested):
In other words: Requesting changes ;) .

pdamme:
I figured out how to use `JOIN` correctly:
```
result = sql(
	"SELECT l.l_orderkey,
		sum(l.l_extendedprice * (l.l_discount * -1 + 1)) as revenue,
		o.o_shippriority
	FROM customer c JOIN orders o ON c.c_custkey = o.o_custkey JOIN lineitem l ON o.o_orderkey = l.l_orderkey
	WHERE c.c_mktsegment = 0
		AND o.o_orderdate < 9204
		AND l.l_shipdate > 9204
	GROUP BY l.l_orderkey,
		o.o_orderdate,
		o.o_shippriority;
		"
);
```

However, the query result is empty. It turns out, even the joins of customer and order as well as orders and lineitem are empty:

```
result = sql("SELECT c.c_custkey, o.o_orderkey FROM customer c JOIN orders o ON c.c_custkey = o.o_custkey");
```
```
result = sql("SELECT l.l_orderkey, o.o_orderkey FROM lineitem l JOIN orders o ON l.l_orderkey = o.o_orderkey");
```

Note that the initial frames contain the right data.

But maybe I'm doing something wrong. I'd say it's good that we found those problems. We should take some more time to find out what is going wrong and to solve the issues. For now, I suggest we remove the example 3 and go only with examples 1 and 2, ao that we can include them with the v0.1 release.

pdamme (approved):
I've addressed the points myself now, since we would like to include this tutorial in the v0.1 release.

- Fortunately, for some reason, example 2 now went smoothly (maybe I did something wrong yesterday...).
- As mentioned above, I couldn't make example 3 work, so I removed it for now. But don't worry, we can easily retrieve it from the history and keep working on it. Would be great to include it in the end.

Anyway, thanks again @DerSchmidt for preparing the tutorial! I think it is already a very good starting point that will be useful to people. We just need to keep extending it ;) .

pdamme:
Sorry, seems like my last commit on your PR didn't contain all changes I made... I've added 33c5c0f4ce04b7f4ab5edba831ed4f96f0282e20 on main in your name after the merge.

PR #444, opened 2022-10-16
Release scripts

corepointer:
Lacking documentation but already ready for a first review I'd say. 
This PR not only adds scripts to prepare artifacts for release but also contain a few other additions/improvements (contained in their respective commits):
* Proper use of bin and lib directories (this one affects documentation and deployment scripts)
* The test script now understands the feature flags and passes them on to the build script (not the most sophisticated solution as these flags need to be updated twice if changed)
* OpenBLAS compatibility is improved to support a wider range of x86_64 processors). The max threads property is taken from the machine where the release is created (so atm it's set to 112 cores). This is open for debate if we should provide a fixed value there and what this value should be. According to the OpenBLAS documentation this setting affects internal structures.

The full pipeline is untested as I copy the created files back to my computer for signing (don't have gpg agent forwarding set up).
For the final merge some things like DAPHNE_REPO_URL need to be changed of course.

At them moment I use these scripts like this:
1) **Create the artifacts:** ``time ./release.sh --version 0.1-testing-rc04 --githash `git rev-parse --short HEAD` ``
2) **Copy back to my machine:** ``rsync -havP <hostname>:./code/daphne-eu/daphne/build/daphne-0.1*rc04*.* .``
3) **Signing and checksumming:** ``~/code/daphne/prototype/release.sh --artifact ./daphne-0.1-testing-rc04-bin.tgz --gpgkey B28F8F4D``

The release.sh script will call pack.sh which will call build.sh and test.sh. Only if testing completes successfully, the artifact, a gzipped tar archive (format open for discussion) is created. The command after ``--githash`` fetches the git hash of the current commit. The script checks out this git hash and restores the current commit after successful completion. This is a bit of a shortcoming as you have to issue a manual ``git checkout -`` if the script fails and terminates early.

Note to self: don't write all the documentation in the PR description :-P


pdamme:
Thanks @corepointer! Overall, it **looks very good** to me. Not only the release-related scripts, but also the other improvements.

**Things we need to finalize before the merge:**
- Currently, the built executables and libraries reside in different locations in a build from source (`/build/...`) and in a release (`/bin/...` and `/lib/...`). At least the locations of the executables are used many times thorughout the documentation, deployment scripts, and source code. I highly recommend using the same locations in both cases. I personally like the `/bin/` and `/lib/` approach. Why don't we adopt that generally? I think we would just need to (1) adapt the `CMakeLists.txt`s, and (2) the `--clean` in `build.sh`.
- Depending on how we resolve the first point, we need to update all occurrences of the old paths thoughout the code base.

**Optional improvements:**
- Likewise, the user configuration file `UserConfig.json` resides in different locations. I recommend generally moving it from `/src/api/cli/UserConfig.json` to `/UserConfig.json` (even in the normal repo).
- `pack.sh` and `release.sh` both have a variable `unknown_options`, but it doesn't seem to have an effect.

Please let me know if/which of those points you'd like to address, or if I shall try it.

corepointer:
Thx for the review @pdamme !

Writing this message while the test suite runs before I push changes ;-) 
* Adjusted bin/lib output. Also redirected the test executables to the bin directory btw. 
* The unknown_options thing is a copy&paste error from build.sh. I pulled over the printing of these from build.sh now.

Suggestions:
* For UserConfig.json I recommend a conf directory. Or even etc? :-D
* We could move all the shell scripts (it has grown to quite a number) to a directory as well while we're at it. Maybe even to bin (I know, they're not binary but still ;-) )
* What about an env.sh that sets up PATH, LD_LIBRARY_PATH and DAPHNE_ROOT (and potentially other stuff) ?
* In the style of SystemDS builtin functions, we could wrap our algorithms in a function body and use a default import from DAPHNE_ROOT/scripts/{algorithms,builtins}. That would boost usability imho.
* At them moment I adjusted the output of libs and executables to what we need. In the long run we should make proper use of cmake install and package features. But that is more a v0.2 thing ;-)

pdamme:
Sounds very good. In general, I would not over-engineer things right now (we want to be done at some point...). We will see what we really need later. So:

- For UserConfig.json, I'm fine with both individual file or `conf`-directory (I just don't know what else we will ever put there...).
- Moving shell scripts: IMHO: simple in theory, might need adjustments in various places (missing them could cause errors), no clear benefit, and not directly related to release automation. So I suggest leaving that for later.
- env.sh: IMHO also not required right now :)
- Wrapping DaphneDSL scripts in functions for (re)usability: definitely! But not now, please. Preparing a simple-to-use library of high-level ML and other primitives in DaphneDSL is on our agenda, but it's a task of its own and should be done thoroughly.
- Thanks and agreeing :)

Finally, I think the release scripts are good as they are (with the points above addressed), for now. We can still improve them for the next releases.

pdamme:
Thanks for finalizing it. Good point about the config dir, we can do that after the release :) .

PR #451, opened 2022-10-20
Fpga operators

ratusz:
GEMV added to the SGEMM kernel, bitstreams removed from main repo


ratusz:
conflict is not important - the file name(example script) is just changed

PR #460, opened 2022-11-16
459: enable CI

auge:
- prepares an ubuntu base image with required dependencies
- builds the daphne system
- runs test suite
- stores created artifacts

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,63 @@
+name: Build DAPHNE
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    env:
+      # fix usage of tput in build script:
+      TERM: dumb
+      DEBIAN_FRONTEND: noninteractive
+
+    steps:
+    - 
+      name: Get latest CMake and Ninja
+      # Using 'latest' branch, the most recent CMake and ninja are installed.
+      uses: lukka/get-cmake@latest
+    -
+      name: "Install build dependencies"
+      run: |
+        set -e
+        pwd
+        echo Installing ...
+        sudo apt-get update
+        sudo apt-get install -y clang default-jdk git libssl-dev lld pkg-config python3 python3-pip gfortran uuid-dev libboost-dev unzip wget moreutils
+        sudo python3 -m pip install -U numpy
+        echo "---------"
+        echo "clang: $(clang --version)"
+        echo "cmake: $(cmake --version)"
+        echo "ninja: $(ninja --version)"
+        python3 -m pip show numpy
+
+    -
+      name: Checkout
+      uses: actions/checkout@v3
+    -
+      name: Building
+      run: |
+        set -e
+        ./build.sh -nf --clean --yes | ts
+        ./build.sh -nf | ts
+    -
+      name: Testing
+      run: ./test.sh | ts
+    -
+      name: "List generated files"
+      run: |
+        ls -la bin
+        ls -la lib
+    -
+      name: 'Upload Artifact'
+      uses: actions/upload-artifact@v3
+      with:
+        name: daphne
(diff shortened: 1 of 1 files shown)


auge:
please comment on whether we should enable only for main branch - would need a small change in the lines 5&6:

```
  push:
    branches: [ "main" ]
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,55 @@
+name: Build DAPHNE
+
+on:
+  workflow_dispatch:
+  push:
+
+jobs:
+  build:
+    runs-on: ubuntu-22.04
```
corepointer, line 9: We should let the automated testing happen on 20.04 imho as we test the newer versions on our dev machines anyway.

.github/workflows/main.yml
```diff
@@ -0,0 +1,55 @@
...
+      name: "Install build dependencies"
+      run: |
+        set -e
+        pwd
+        echo Installing ...
+        sudo apt-get update
+        sudo apt-get install -y cmake clang default-jdk git libssl-dev lld pkg-config python3 python3-pip gfortran uuid-dev libboost-dev unzip wget moreutils
+        sudo wget -qO /usr/local/bin/ninja.gz https://github.com/ninja-build/ninja/releases/latest/download/ninja-linux.zip
+        sudo gunzip /usr/local/bin/ninja.gz
+        sudo chmod a+x /usr/local/bin/ninja
```
corepointer, line 26: What's wrong with the packaged version of ninja?

.github/workflows/main.yml
```diff
@@ -0,0 +1,55 @@
...
+      run: |
+        ls -la bin
+    -
+      name: 'Upload Artifact'
+      uses: actions/upload-artifact@v3
+      with:
+        name: daphne
+        path: |
+          bin/daphne
+          bin/DistributedWorker
```
corepointer, line 55: What is this list and upload for (sorry - not an expert on Github Workflows/Actions) ?

corepointer (commented):
Thank you for taking care of this so promptly!
I think this is a valuable addition. As resources for the free runners are limited I suggest we run this only on additions to the main branch (PR and direct push)

.github/workflows/main.yml
auge: ok, I can modify this ...

.github/workflows/main.yml
auge: hm, I could not found a packaged version in apt nor snap ??

.github/workflows/main.yml
auge: list was a debugging step (are binaries generated)
upload is for storing the generated binaries (without the test_runner) for downloading - using Github to build the binary instead of locally ... (accessible for contributors only, for public, a release must be created)

.github/workflows/main.yml
corepointer: It's called "ninja-build" ;-)

.github/workflows/main.yml
corepointer: OK this sounds nice :) Please include the lib directory as well to have a fully working DAPHNE build. Btw does this work with wild cards (e.g., bin/* lib/*) ?

auge: ok I see. I assumed you statically link.
the star would not be possible.
I can provide an update tomorrow, have to leave for an appointment now ;)

auge:
@corepointer : any blocking points?

corepointer:
Sorry for the late reply. I'll come back to this after the deliverables are finalized.

corepointer:
Thx @auge for starting work on the github actions. We'll need caching (of the thirdparty dependency build step) or run this from a prebuilt container imho as one run takes 100+ minutes which will quickly deplete our [quota](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions). Same goes for the generated artifacts.

I merged this now to see how it will work out on the main branch. In case the issues become real we can disable the workflow again.

PR #463, opened 2022-12-05
Eigen op 391

KostasBitsakos:
The EigenOp complete.

Returns two matrixes  as results, one for the Eigen Values and one for the Eigen Vectors.
The res1 matrix for eigenvalues and the res2 matrix for eigenvectors
Column k of the returned matrix res2 is an eigenvector corresponding to eigenvalue number k as returned by eigenvalues(). 
The eigenvectors are normalized to have (Euclidean) norm equal to one. 

Also includes the download and build commands for the cpp Eigen library, using the cmake tool.

### DIFF
--- a/src/runtime/local/kernels/EigenCal.h
+++ b/src/runtime/local/kernels/EigenCal.h
@@ -1,0 +1,148 @@
+/*
+ * Copyright 2021 The DAPHNE Consortium
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <runtime/local/context/DaphneContext.h>
+#include <runtime/local/datastructures/CSRMatrix.h>
+#include <runtime/local/datastructures/DataObjectFactory.h>
+#include <runtime/local/datastructures/DenseMatrix.h>
+#include <runtime/local/kernels/IsSymmetric.h>
+#include <Eigen/Eigenvalues>
+#include <Eigen/Dense>
+
+#include <iostream>
+#include <cassert>
+#include <cstddef>
+
+
+// ****************************************************************************
+// Struct for partial template specialization
+// res1 matrix for eigenvalues, res2 matrix for eigenvectors
+// Column k of the returned matrix res2 is an eigenvector corresponding to eigenvalue number k as returned by eigenvalues(). 
+// The eigenvectors are normalized to have (Euclidean) norm equal to one. 
+// ****************************************************************************
+
+template<class DTRes1, class DTRes2, class VTArg>
+struct EigenCal {
+    static void apply(DTRes1 *& res1, DTRes2 *& res2, const VTArg * inMat, DCTX(ctx)) = delete;
+};
+
+// ****************************************************************************
+// Convenience function
+// ****************************************************************************
+template<class DTRes1, class DTRes2, class VTArg>
+void eigenCal(DTRes1 *& res1, DTRes2 *& res2, const VTArg * inMat, DCTX(ctx)) {
+    EigenCal<DTRes1, DTRes2, VTArg>::apply(res1,res2, inMat, ctx);
+}
+
+// ****************************************************************************
+// (Partial) template specializations for different data/value types
+// ****************************************************************************
+
+// ----------------------------------------------------------------------------
+// DenseMatrix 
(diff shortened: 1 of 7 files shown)


test/runtime/local/kernels/EigenCalTest.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include <cstdint>
+
+
+template<class DT>
+void checkEigenCal(const DT * inMat, const DT * exp1, const DT * exp2) {
+    DT * res1 = nullptr;
+    DT * res2 = nullptr;
+    eigenCal<DT,DT, DT>(res1, res2, inMat, nullptr);
+    CHECK(*res1 == *exp1);
+    CHECK(*res2 == *exp2);
```
corepointer, line 38: These checks fail due to precision problems. Use something like this (taken from SolveTest.cpp):
    CHECK(checkEqApprox(res, exp, 1e-6, nullptr));

corepointer (changes requested):
Thank you @KostasBitsakos  for your efforts on the initial version.

I've cleaned it up a little, fixed the dependency installation and rebased on the current state of the main branch (rebasing needs force pushing so you will get conflicts that you can circumvent by either hard resetting your local branch to the state on github or moving your local copy out of the way and then checking out the branch again).

Besides addressing the inline comments from the code, please add the following:
* There's a isSymmetric() function somewhere in the code base (in runtime/local/kernels iirc) that you can use to check for symmetric inputs and gracefully fail with an informative message if your EigenCal kernel is used in another way.
* Add the necessary bits and pieces  to use the kernel from DaphneDSL. Ideally with another test case.

Regards, Mark

src/runtime/local/kernels/EigenCal.h
```diff
@@ -0,0 +1,157 @@
...
+        const auto nr = static_cast<size_t>(inMat->getNumRows());
+        const auto nc = static_cast<size_t>(inMat->getNumCols());
+        Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic> inputMatrix(nr, nc);
+       
+       std::tuple<int, int>{1, -1};
+        for(size_t r=0; r<nr; r++){
+            for(size_t c=0; c<nc; c++){
+                inputMatrix(r,c)=inMat->get(r,c);
+            }
+        }
```
corepointer, line 80: Please try to avoid these copy loops. Eigen can work on array data. Maybe this helps to get you started:
https://stackoverflow.com/questions/28722899/creating-an-eigen-matrix-from-an-array-with-row-major-order

src/runtime/local/kernels/EigenCal.h
```diff
@@ -0,0 +1,157 @@
...
+                    res1->set(r,c,s.eigenvalues()[r].real());
+    }
+    }
+    for(size_t r = 0; r < eigenVectorsrows; r++) {
+                for(size_t c = 0; c < eigenVectorscols; c++) {
+                    res2->set(r,c,eigenVectors.coeff(r, c));
+    }
+    }
+    }
+};
```
corepointer, line 107: Likewise the output does not need to be copied. There are DenseMatrix constructors taking chucks of memory and I'm sure there's a way to get the data pointer out of an Eigen matrix.

corepointer: Thinking twice about it it'd be better to allocate the DenseMatrix and pass its data pointer to an Eigen matrix. Otherwise your memory for the output would be held by a local variable that gets deallocated at the end of your function.

corepointer:
Hi @KostasBitsakos!
I rebased this branch onto the latest changes in main, resolved some conflicts and removed that one extra commit that added temporary files from the test cases. Once the mentioned issues from the review are resolved, we can merge it in.
Regards, Mark

corepointer:
There was a minor error on main. I used merging instead of rebasing to pull this here. This should avoid conflicts and can be fixed with the final merge to main anyway.

And please don't worry about the failing checks (I cancelled the last one running because it would fail anyway). They fail because the Eigen library is not in the Docker image that is used for the test compile yet.

KostasBitsakos:
Hi Mark,

I did a hard reset on my local branch, to your last commit.
As you mentioned the tests are now failing.
That's an issue cause I cannot work without testing my results.
Also I need to change some things on my testing code.
Shall I work on the b0da452 commit maybe and then we can merge them
together?

Regards,
Constantinos

Στις Παρ 17 Φεβ 2023 στις 11:53 π.μ., ο/η Mark ***@***.***>
έγραψε:

corepointer:
I meant the GitHub Action test is failing and that can be ignored (what you see if you click on "Checks" in this pull request). Your own tests *should* work fine. Please share more details and error messages of your failing tests.

corepointer:
Any updates on the compilation issues? If it's still blocking you from fixing the remaining issues (see comments from my code review above) you could use the daphneeu/github-action docker container where all the dependencies (including Eigen) are already installed. There you'd run the build script with ```./build.sh --no-deps --installPrefix /usr/local``` to avoid building the dependencies.
Hope that helps :)

corepointer (approved):

corepointer:
LGTM - thx for your efforts @KostasBitsakos. I've merged this with minor cleanup. Please add documentation and a script level test case in another commit.

Issue #465, opened 2023-01-24
Abstractions for distributed backends (MPI, gRPC, ...)


AhmedEleliemy:
This issue will serve as a place to continue the discussions of wp4/wp5 on abstractions of (data partitioning, placement, and serialization) which are needed for the distributed backends
at the moment, the distributed backed with gRPC (already merged to the main branch) and the [MPI fork](https://github.com/AhmedEleliemy/daphne) has the following issues
1. a lot of code snippets are repeated, thus, can be abstracted for better code readability and maintainability 
2. certain things are considered static (how the data is partitioned across workers), and thus, could be abstracted  in more generic ways

AhmedEleliemy:
An initial suggestion is as follows
### Serialization/deserialization
serializeStructure(void * DataBuffer, DT * Object, size_t* BufferLength)
deserializeStructure(void * DataBuffer, DT * Object, size_t BufferLength) 
### Data Partition/Placement
replicateData(void * Identifiers, size_t identifiersCount, Range * Ranges, DT * Object)
paritionData (enum WayofParition, void * Identifiers, enum method, Range * Ranges, DT * Object)
placeData(void *Identifiers, Range * Ranges, size_t * identifiersCounter, DT * Object)

we also need to discuss the following
- How the suggested abstractions should be implemented? Code style (return parameters vs referenced parameters, abstraction naming ... ) Design style (relay on Utility classes, e.g., Load Partitioning to offer these abstractions, ...)
- When does DAPHNE take the partitioning decisions? Compile/runtime
- How does DAPHNE account for cache size information?

pdamme:
The serialization aspect of this issue should be addressed, now that #467 has been merged.

Issue #103, opened 2021-08-12
(De)Serialization of data objects


daphne-eu:
In GitLab by @pdamme on Aug 12, 2021, 11:41

We need a way to serialize our in-memory data objects (`DenseMatrix`, `CSRMatrix`, `Frame`) to byte streams and to deserialize such byte streams to get back the data objects again.

This is required at multiple points, e.g.:
- For the data transfer between nodes in the distributed run-time.
- For complex IO handling, e.g. in connection to computational storage (even though that might also use its own (de)serialization routines)
- For persistence using simple Daphne-specific binary file formats.

PR #467, opened 2023-01-25
Initial implementation of DaphneSerializer

aristotelis96:
This PR is about implementing Daphne's serializer.

- Added `DaphneSerializer`.
- `DaphneSerializer` provides 3 static functions, `save`, `load` and `length`, used for serializing Daphne objects into byte streams (and vice versa).
- Current support for DenseMatrix and CSRMatrix (Frame not implemented).
- Updated ReadDaphne and WriteDaphne to use DaphneSerializer.
- Updated distributed kernels and distributed gRPC worker to serialize objects with DaphneSerializer and use protobufs to transfer bytes only, instead of complex protobuf messages.
- Removed `ProtoDataConverter`.

Issue for discussion:
- DaphneSerializer is a templated class, with static methods. Distributed kernels need to downcast from `Structure*` to `Dense/CSR<int/double>` matrices before serializing an object, leading to an if-else chain using `dynamic_cast`. I believe that this could be avoided with virtual functions, i.e. instead of serializing an object in a seperate (static) templated class, we could provide virtual functions for Dense/CSR/Frames that serialize an object. Feel free to comment and provide alternative ideas/suggestions.

Related issues #465.


pdamme:
Thanks for this contribution @aristotelis96. I will look into it and will provide feedback today or tomorrow.

pdamme (changes requested):
Hi again, **factoring out the (de)serialization logic to a central component is a good idea**, since it allows us to use it from mutliple places, e.g., file I/O in the local runtime and data transfer in the distributed runtime.

**Regarding the problem of templates you mentioned:**

As far as I see it, this could be solved by forwarding the template parameters of the kernels to `DaphneSerializer`. For instance, the `readDaphne`/`writeDaphne` kernels have a template parameter for the data type (including the value type) `DT`. Their partial specializations for `DenseMatrix`/`CSRMatrix` have a template parameter for the value type `VT` and, hence, use `DaphneSerializer<DenseMatrix<VT>>`/`DaphneSerializer<CSRMatrix<VT>>`.

I think the same trick should work for the `broadcast`/`distribute`/`distributedCollect` kernels. Each of them gets the data type (including the value type) `DT` as a template parameter. So we know the type of the data object at C++ compile time and don't need to determine it at runtime via if/else and dynamic casts; just use `DaphneSerializer<DT>`.

A remark on `distributedCollect`: The data we receive could be represented in another format, since each block of the DAPHNE binary format can have its individual representation. However, this should not be an issue. The (de)serialization logic should be (made) general enough to cope with that case (instead of erroring out).

**Regarding additional copying:**

Unfortunately, the current implementation introduces additional copying into the critical IO/transfer path. For instance, `readDaphne` used to create a large enough matrix and to populate it as it read the file step by step. Now, the entire file is first read into a memory-resident buffer before the deserialization from the buffer into the matrix starts, i.e., we create two copies of the data, thereby increasing runtime and memory footprint. This should be improved, since one reason for using our own serialization instead of protobuf is to increase performance by avoiding protobuf's inefficiencies.

In particular, extra copies are currently done for serialization (`writeDaphne`, `broadcast`, and `distribute` first serialize the data object into a new buffer before they write it to a file or pass it to protobuf) and deserialization (`readDaphne` was mentioned above; `distributedCollect` creates a new data object per slice and copies its contents into the actual target object).

I think, `DaphneSerializer::save()` should be able to pass data to some consumer, instead of fully materializing it in memory. Likewise, `DaphneSerializer::load()` should be able to read the data from some producer. Maybe this could be achieved by providing some logic (e.g., as a template parameter) for interacting with the producer/consumer. Furthermore, it would be great if deserialization supported writing to a (view into an) existing data object, but that could be left for later.

**Please let me know what you think.** In case we quickly need the factored out serialization for #465, I'm open for postponing these fixes, but I think they could have an impact on the serialization API.

pdamme:
FYI: This PR contributes to #103. We should mention it in the commit message once we merge this PR.

aristotelis96:
Thanks @pdamme for your review and feedback. 

**Regarding the problem of templates:**

Distributed runtime kernels (distribute, broadcast, etc.) are called from `DistributedWrapper`, which is initialized and called from `DistributedPipeline.h` kernel. Distributed kernels are templated, however `DistributedWrapper` only uses templates for the pipeline output, whereas each pipeline input is treated as a `Structure` pointer. Therefore the template parameter passed to the distributed kernel (and by extension the serializer) would be `Structure` (instead of `DenseMatrix`/`CSRMatrix`). For now it is not clear to me how we could provide the template parameter for the pipeline inputs, especially since we want to use `Structure` pointers in case there are different data types of pipeline inputs. 

An option would be to create a specialization of `DaphneSerializer` for `Structure` where we could _hide_ the `dynamic_cast` chain there, but this shifts/hides the issue instead of actually solving it. 
Let me know what you think.

aristotelis96:
I've updated DaphneSerializer and included a few additional implementations used for (de)serializing in chunks. One implementation for in order serialization using iterators and one for out of order.

- `DaphneDe/SerializerChunks` : Used for (de)serializing in order using an iterator. Example use:
```c++
auto ser = DaphneSerializerChunks<DT>(mat, chunkSize);
size_t idx = 0;            
for (auto it = ser.begin(); it != ser.end(); ++it){ 
std::copy(it->second.begin(), it->second.begin() + it->first, tempBuff.begin() + idx);
    idx += it->first;
}
DT* newMat = nullptr;
idx = 0;
auto deser = DaphneDeserializerChunks<DT>(&newMat, chunkSize);
for (auto it = deser.begin(); it != deser.end(); ++it){
    size_t chnck = std::min(size_t(chunkSize), tempBuff.capacity() - idx);
    std::copy(tempBuff.begin() + idx, tempBuff.begin() + idx + chnck, it->second.begin());
    it->first = chnck;
    idx += chnck;                
}
``` 
- `DaphneDe/SerializerOutOfOrderChunks`. Used for (de)serialiazing in chunks, out of order. The chunks produced are identical to the previous implementation but contain an additional header used to identify the index of the chunk. Example use:
```c++
std::vector<std::vector<char>> message;

DaphneSerializerOutOfOrderChunks<DT> serializer(mat, chunkSize);
while (serializer.HasNextChunk())
{
  std::vector<char> bufferTmp(200);
  serializer.SerializeNextChunk(bufferTmp);
  // Push to vector of messages
  message.push_back(bufferTmp);
}

// Shuffle vector of messages
std::shuffle(std::begin(message), std::end(message), rng);

// Deserialize
DaphneDeserializerOutOfOrderChunks<DT> deserializer;
size_t i = 0;
DT *res = nullptr;
// We want to iterate using HasNextChunk() method
while (deserializer.HasNextChunk())
{
  // Deserialize message
  res = deserializer.DeserializeNextChunk(message[i++]);
}
```

aristotelis96:
Updated `ReadDaphne` and `WriteDaphne` kernels to serialiaze an object and read/write in chunks, instead of using one big buffer.
**TODO**: In both `Read/WriteDaphne` kernels specializations for `DenseMatrix` and `CSRMatrix` are the same. Could we provide a single implementation for both of them? Instead of `delete`ing the default implementation we could implement that, however I am not sure if we want to allow calling `Read/WriteDaphne` with other objects (e.g. `Matrix`, `Structure`).

pdamme (approved):
Thanks @aristotelis96 for promptly addressing the comments. The PR is **ready to merge** now. The additional source code documentation will be helpful. Thanks also for extending the test cases!

A little downside of making `DEFAULT_SERIALIZATION_BUFFER_SIZE` is that it is now defined in multiple places, but that can be addressed after the merge.

I will squash the commits in this PR to create a single commit by @aristotelis96, co-authored by @aristotelis96 and @psomas (since @psomas had also contributed to it, especially in the beginning). Would that be okay?

aristotelis96:
Yes, that sounds good to me. Thank you @pdamme 
for reviewing this PR!

Issue #466, opened 2023-01-24
Behavior of build.sh --clean


corepointer:
Our build script has two types of cleaning the source tree:
* **--clean**: removes the DAPHNE build directories **and** the dependencies
* **--cleanAll**: everything that --clean does plus downloaded packages and extracted 3rd party sources

Imho this would be more practical:
* --clean: remove bin, build, lib from DAPHNE source tree
* --cleanAll: all from --clean plus compiled 3rd party deps
* --cleanDownloadCache: (name is just a suggestion) remove the downloaded packages and extracted sources

The idea originated from how often I needed one of these three in the past.

pdamme:
Hi @corepointer, I think this is a very good idea, especially since it allows removing the DAPHNE-specific artifacts without removing anything from the third-party dependencies.

I also agree with what these three steps should remove. Regarding the wording, I suggest:

`--clean`: removes `bin`, `build`, `lib`
`--cleanDeps`: removes `thirdparty/build`, `thirdparty/installed`, `thirdparty/*.install.success`, `thirdparty/llvm-last-built-commit.txt`
`--cleanCache` (or `--cleanDownloads`): removes `thirdparty/download-cache`, `thirdparty/sources`, `thirdparty/*.download.success`, and `thirdparty/llvm-project` (except for `thirdparty/llvm-project/.git`)

`--cleanAll`: combines `--clean`, `--cleanDeps`, and `--cleanCache`

The first three options remove disjoint sets of directories/files. Then, to my mind, `--cleanAll` should really remove *all* of them.

corepointer:
Thx for your comments @pdamme  I'll make it so ;-)

PR #471, opened 2023-02-05
Build script improvements

corepointer:
This PR contains several minor and major improvements to our infamous build script. Some of it is prep work for building daphne containers.

Commit messages speak for themselves but here's a quick summary:

* The clean parameters are changed according to #466 
* The prefix where third party dependencies are installed can be set with --installPrefix <path>
* The LLVM/MLIR dependency is now installed into installPrefix and the use of buildPrefix removed from the main cmake config invocation (at the bottom of build.sh)
* Third party dependency compilation can be omitted via --no-deps flag

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,7 +33,7 @@ jobs:
         path: ./thirdparty
         key: ${{ runner.os }}-${{ env.cache-name }}-${{ hashFiles('build.sh') }}
         restore-keys: |
-          ${{ runner.os }}-${{ env.cache-name }}-
+          ${{ runner.os }}-${{ env.cache-name }}-${{ hashFiles('build.sh') }}
 
     - if: ${{ steps.cache-deps.outputs.cache-hit == 'true' }}
       name: Print cache hit result
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -177,5 +177,6 @@ if(USE_FPGAOPENCL)
 endif() 
 add_subdirectory(src/util)
 
+add_dependencies(CompilerUtils MLIRDaphneTransformsIncGen)
 
 add_subdirectory(test)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -129,20 +129,9 @@ if(USE_CUDA AND CMAKE_CUDA_COMPILER)
     endif()
 endif()
 
-option(USE_ARROW "Whether to activate compilation of Arrow/Parquet features" OFF)
-if(USE_ARROW)
-    find_package(Arrow CONFIG REQUIRED
-        PATHS thirdparty/installed/lib/cmake/arrow
-        NO_DEFAULT_PATH
-    )
-    find_package(Parquet CONFIG REQUIRED
-        PATHS thirdparty/installed/lib/cmake/arrow
-        NO_DEFAULT_PATH
-    )
-    link_libraries(arrow_shared parquet_shared)
-    add_definitions(-DUSE_ARROW)
-    message(STATUS "Arrow/Parquet enabled")
-endif()
+
+find_package(Arrow REQUIRED)
+find_package(Parquet REQUIRED)
 
 option(USE_FPGAOPENCL "Whether to activate compilation of FPGA OpenCL features" OFF)
 if(USE_FPGAOPENCL)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,15 +74,15 @@ set(BLA_VENDOR OpenBLAS)
 set(BLA_STATIC ON)
 find_package(BLAS)
 
+SET(ANTLR_VERSION 4.9.2 CACHE STRING "The version of ANTLR (parsing library).")
+message(STATUS "ANTLR_VERSION: ${ANTLR_VERSION}")
+
 # specify multiple paths in CMAKE_PREFIX_PATH separated by semicolon to add multiple include dirs
 # to make compile/exec in container plus finding includes in local IDE work, specify both, local third party sources
 # prefix and /usr/local e.g.: -DCMAKE_PREFIX_PATH="/usr/local;${PROJECT_SOURCE_DIR}/thirdparty/installed"
-
-message(STATUS "ANTLR_VERSION: ${ANTLR_VERSION}")
 foreach(path ${CMAKE_PREFIX_PATH})
     message(STATUS "CMAKE_PREFIX_PATH:  ${path}")
     include_directories(${path}/include)
-    include_directories(${path}/include/openblas)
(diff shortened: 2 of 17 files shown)


auge:
Hi,
where do you have the image `daphneeu/github-action:latest` available?
* is the source on github (account daphne-eu) ?
* is the image on docker hub ?

corepointer:
> Hi, where do you have the image `daphneeu/github-action:latest` available?
> 
>     * is the source on github (account daphne-eu) ?
> 
>     * is the image on docker hub ?

I set up the daphneeu account on Docker Hub (no "-" allowed, otherwise it would have been daphne-eu there too). We'll change that account to an organization. 

The image can be pulled from Docker Hub. This was a first test. I'll push some more images soon and add the source to our Github repo as well.

pdamme (changes requested):
Hi @corepointer, thanks for these useful improvements of the build script and documentation. **Overall, it looks good to me**, however, here are a few comments, some of which should be addressed:

**Required, IMHO:**
- The top-level `CMakeLists.txt` still hardcodes the default third-party installation prefix for arrow (optional dependency), so setting `--installPrefix` won't have an effect here.
- `GettingStarted.md` says that DAPHNE can be built on all Ubuntu versions in the range 20.04 to 22.10. However, on my Ubuntu 22.04 system, it does not work out of the box. Have you tried it with 22.04? I can provide more details if needed, but we don't need to make it work right now, but should document it accurately.
- Minor: Please consistenly use an indentation of 4 spaces in `build.sh`.

**Optional:**
- While it is not strictly an error, I noticed a somewhat unintuitive behavior. Assume we have a fully built repo (DAPHNE and dependencies), using the default install prefix. Now, invoking `./build.sh --installPrefix "$(pwd)/thirdparty/my_installed"` has no effect. Since all the install success token files are present, nothing is built anew and nothing will be in the new install prefix `$(pwd)/thirdparty/my_installed`. So we need to do `./build.sh --cleanDeps` first to get rid of the old built/installed third-party deps. Now, invoking `./build.sh --installPrefix "$(pwd)/thirdparty/my_installed"` causes a rebuild of the third-party deps, but many of the variables/paths in cmake were *cached* so they still point to the old/default install prefix. This even causes the build to fail (for antlr in my case, but I think the problem is more general). 
  ```
  -- CMAKE_PREFIX_PATH:  /daphne/thirdparty/my_installed
  -- ANTLR_JAR_FOUND: /daphne/thirdparty/my_installed/share/antlr4/antlr-4.9.2-complete.jar
  -- ANTLR4_JAR_LOCATION is /daphne/thirdparty/installed/share/antlr4/antlr-4.9.2-complete.jar
  CMake Error at thirdparty/my_installed/lib/cmake/antlr4-generator/antlr4-generator-config.cmake:156 (message):
    Unable to find antlr tool.
    ANTLR4_JAR_LOCATION:/daphne/thirdparty/installed/share/antlr4/antlr-4.9.2-complete.jar
  ```
  So we need to do `./build.sh --clean` as well to get rid of the cached variables. That may all be fine, since one may not change the install prefix very often. But at least we should document that, i.e., before invoking `build.sh` with another install prefix, one should at least do `./build.sh --clean && ./build.sh --cleanDeps`.
- MLIR: You also build the target `install`. We had discussed this a while ago in #359. It adds more targets to build (~700), which takes some more time. You provided a clear justification in the commit message, and I'm fine with it if it simplifies things. I'm just wondering why it helps, because we never used any of the outputs of the install target before.

**I have also pushed a few minor fixes** to your branch. Feel free to handle them as you wish (e.g., integrate into your existing commits).

corepointer:
Thx for your comments @pdamme 
* The Arrow paths hard coded in CMakeLists.txt might explain troubles I had with adding that to the Docker image. I'll have a closer look at it. I was actually considering removing the --arrow flag and just compiling it anyway. I'll open a separate issue reasoning about it.
* Yes I had DAPHNE running at times when Ubuntu 22.04 was the current version. But I didn't test it recently. If you share more details about your troubles in another issue we'll figure it out.
* I had in the back of my head that we use 2 spaces for indentation in scripts. I'll let my formatter have a go at it ;-)
* Yes, the cached paths are a problem. I think documenting that caches need to be cleared if the installPrefix is changed is a solution we can live with (to not over engineer the already quite elaborate(to use another wording ;-)) build script).
*  Imho we use quite a lot of the LLVM and MLIR libraries, binaries and include files in DAPHNE. Removing the reference to buildPrefix tremendously reduces the size of the Docker image with prebuilt deps (saving 6GB). As I mentioned somewhere, many of the changes are related to better handling of the Docker images and also the github action. So having those extra 700 compile steps in the dependency compilation should pay of as people won't have to compile dependencies over and over again if they either use the containers or have the dependencies compiled once into a directory that can be moved out of the way when cleaning the source tree.

pdamme (changes requested):
Hi @corepointer, thanks for the fixes and additions. Quickly glancing over it, it looks good to me. There do not seem to be so many changes to the commits that existed before you incorporated my feedback (except for some docu). The two latest commits on reformatting `build.sh` and on improving the Arrow integration were added.

I pushed a minor fix removing `USE_ARROW` from the `ReadParquet` test case.

For testing, I did the following:
- clone freshly and `./build.sh` and `./test.sh`: success
- `./build.sh --clean` and `./build.sh`: success
- `./build.sh --cleanDeps`: success (and expected files removed)
- `./build.sh --cleanCache`: no prompt, but some error and then the script continues doing unrelated things...
  ```
  root@3c7b6b5a37fa:/daphne# ./build.sh --cleanCache
  -- Cleanup of downloaded/extracted artifacts of the third party dependencies
  ./build.sh: line 225: `/daphne/thirdparty/sources': not a valid identifier
  [DAPHNE]..Applying 0000-antlr-silence-compiler-warnings.patch.............................................................................................................................[10.02.23 20:27:56]
  patching file runtime/src/atn/ATNState.h
  Reversed (or previously applied) patch detected!  Skipping patch.
  1 out of 1 hunk ignored -- saving rejects to file runtime/src/atn/ATNState.h.rej
  patching file runtime/src/support/Any.h
  Reversed (or previously applied) patch detected!  Skipping patch.
  1 out of 1 hunk ignored -- saving rejects to file runtime/src/support/Any.h.rej
  [DAPHNE]..Build Antlr v4.9.2..............................................................................................................................................................[10.02.23 20:27:56]
  [...]
  ```

**Please double check what's going wrong here.**

pdamme (approved):

pdamme (changes requested):
Sorry, approved by accident, but there are still problems to fix.

pdamme:
Anyway, the required points from my first review have been fixes (we can talk about Ubuntu 22.04 later), and the optional points are also fine (Arrow, and I see the point about the MLIR install target and agree with you).

PR #476, opened 2023-02-23
Duplicate inputs in pipeline input array.

### LINKED ISSUE #461: Duplicate objects in pipeline input array
In the current state of daphne, vectorized pipelines (and by extension distributed pipelines) are created with duplicate objects in the input array. For example the following code:

    x = rand(10, 10, 1.0, 10.0, 1.0, -1);
    print(x+x);

when executed with the `--vec` flag creates a vectorized pipeline with two arguments (two times the same object) instead of one argument:
```
%9 = "daphne.vectorizedPipeline"(%8, %8, %6, %6, %7) ( {
    ^bb0(%arg0: !daphne.Matrix<?x10xf64:sp[1.000000e+00]>, %arg1: !daphne.Matrix<?x10xf64:sp[1.000000e+00]>):  // no predecessors
      %10 = "daphne.call_kernel"(%arg0, %arg1, %7) {callee = "_ewAdd__DenseMatrix_double__DenseMatrix_double__DenseMatrix_double"} : (!daphne.Matrix<?x10xf64:sp[1.000000e+00]>, !daphne.Matrix<?x10xf64:sp[1.000000e+00]>, !daphne.DaphneContext) -> !daphne.Matrix<?x?xf64:sp[1.000000e+00]>
      "daphne.call_kernel"(%arg1, %7) {callee = "_decRef__Structure"} : (!daphne.Matrix<?x10xf64:sp[1.000000e+00]>, !daphne.DaphneContext) -> ()
      "daphne.call_kernel"(%arg0, %7) {callee = "_decRef__Structure"} : (!daphne.Matrix<?x10xf64:sp[1.000000e+00]>, !daphne.DaphneContext) -> ()
      "daphne.return"(%10) : (!daphne.Matrix<?x?xf64:sp[1.000000e+00]>) -> ()
    },  {
    }) {combines = [1], operand_segment_sizes = dense<[2, 1, 1, 1]> : vector<4xi32>, splits = [1, 1]} : (!daphne.Matrix<10x10xf64:sp[1.000000e+00]>, !daphne.Matrix<10x10xf64:sp[1.000000e+00]>, index, index, !daphne.DaphneContext) -> !daphne.Matrix<10x10xf64:sp[1.000000e+00]>
```
notice in the second line that there are two `!daphne.Matrix` objects. 

There are also cases where a pipeline might need a whole matrix and that same matrix splitted by rows/cols, at the same pipeline. In that case vectorization pass adds the same object twice in the input array and `VectorSplit` provides information on how to split (or not) the correnspoding matrix. 

This issue is about updating the vectorized (and distributed) pipelines in order to contain each input only once.

A few hints provided by @pdamme:

- We should search for repetitions of inputs in VectorizedPipelineOp (see DaphneOps.td, DaphneVectorizableOpInterface.td). If a `mlir::Value` is repeated, we need to check the corresponding `VectorSplits`. If they are the same, we can simply replace all uses of the one operand in the pipeline body by the other operand and remove that one operand. If they are not the same (i.e., we have broadcast (`NONE`) and split by row (`ROWS`)), then we must ensure two things: (1) the corresponding VectorSplit of the retained operand must be NONE (broadcast), and (2) all uses (within the pipeline body) of the operand that was split-by-row must be replaced by the result of an ExtractRowOp, which must be inserted at the beginning of the pipeline body. The reason is that for operands that are split-by-row, the operations in the pipeline body expect a row slice, not the entire data object. We also need to make sure that the ExtractRowOp knows which row range to extract, this information must be passed from the runtime in some way (maybe via a new input to the pipeline).
- One approach to integrate this would be to adapt VectorizeComputationsPass accordingly. However, two inputs could become the same after that pass due to other rewrites (e.g., something like "x + t(t(x))" could become "x + x" after the vectorization pass). Thus, I suggest another approach: Implementing a canonicalization for VectorizedPipelineOp (and similarly for DistributedPipelineOp).

aristotelis96:
Both vectorized and distributed pipelines could include an input multiple times in the pipeline input array.

Updated canonicalization pass for VectorizedPipelineOp in order to scan for duplicate inputs and in case both inputs have the same `VectorSplits` entry, one of them is removed.

Closes #461.

### DIFF
--- a/src/ir/daphneir/DaphneDialect.cpp
+++ b/src/ir/daphneir/DaphneDialect.cpp
@@ -44,6 +44,7 @@
 #include <llvm/ADT/BitVector.h>
 #include <llvm/ADT/APInt.h>
 #include <llvm/ADT/APSInt.h>
+#include <llvm/ADT/DenseMap.h>
 
 void mlir::daphne::DaphneDialect::initialize()
 {
@@ -410,6 +411,33 @@ mlir::OpFoldResult mlir::daphne::ConstantOp::fold(FoldAdaptor adaptor)
 mlir::LogicalResult mlir::daphne::VectorizedPipelineOp::canonicalize(mlir::daphne::VectorizedPipelineOp op,
                                                                      mlir::PatternRewriter &rewriter)
 {
+    // // Find duplicate inputs
+    std::vector<Attribute> vSplitsAttrs;
+    for (auto & split : op.getSplits())
+        vSplitsAttrs.push_back(split);
+    auto currentSize = op.getInputs().size();
+    
+    DenseMap<Value, size_t> inputMap;
+
+    for (size_t i = 0; i < currentSize; i++) {
+        const auto& input = op.getInputs()[i];
+        const auto& split = op.getSplits()[i].cast<daphne::VectorSplitAttr>().getValue();
+
+        if (inputMap.count(input) == 0) {
+            inputMap[input] = i;
+        } else {
+            size_t j = inputMap[input];
+            if (op.getSplits()[j].cast<daphne::VectorSplitAttr>().getValue() == split) {
+                op.getBody().getArgument(i).replaceAllUsesWith(op.getBody().getArgument(j));
+                op.getBody().eraseArgument(i);
+                op.getInputsMutable().erase(i);
+                vSplitsAttrs.erase(vSplitsAttrs.begin() + i);
+                currentSize--;
+                i--;
+            }
+        }
+    }
+
     std::vector<Value> resultsToReplace;
     std::vector<Value> outRows;
     std::vector<Value> outCols;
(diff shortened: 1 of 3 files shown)


aristotelis96:
Thank you @pdamme for reviewing this PR.

- Regarding the testcases: Distributed runtime always checks for duplicate inputs in the pipeline array and throws an exception accordingly. Therefore a script level test case is enough to check that the canonicalization pass works. In the case of vectorized execution though duplicate inputs do not cause any errors, even if we have duplicate inputs. We could test the generated IRs by using the `--explain=kernels` flag, however such test case will need to be updated each time there are changes to the IR. Testing just the distributed case might be enough though, since vectorization always happens before creating distributed pipelines. Also, could we create a test case in `test/ir` to check that the canonicalization pass removes duplicates from the pipelines? I'm not sure which is the best approach here, let me know what you think.
- Oops, broadcast is indeed unrelated, will fix this :)

pdamme:
Hi @aristotelis96, you are right, the local vectorized execution was already able to handle duplicate inputs (I remembered that incorrectly). So yes, a script-level test case for the distributed execution with duplicate inputs should be sufficient.

For now, I wouldn't add any checks of the IR, since these would indeed be a bit fragile. Furthermore, the IR is not that crucial here, as long as the DaphneDSL script works, to my mind ;) .

aristotelis96:
Updates: 
- Two of the existing distributed script-level test cases where written in a way to bypass the limitation of duplicate inputs. Both of them have been updated and now use each use duplicate inputs.
- Removed unreleated changes.

pdamme (approved):
Thanks for updating the test cases, **LGTM**.

Issue #132, opened 2021-09-15
Frame column labels after group-join


daphne-eu:
In GitLab by @pdamme on Sep 15, 2021, 20:46

**Background**

The `groupJoin`-operation performs an equi-join of two frames (on the columns labeled `lhsOn` and `rhsOn`) followed by a grouping by the join column including an aggregation (summation) on the column labeled `rhsAgg`. The columns of the output frame are named after the input columns, i.e., `lhsOn` and `rhsAgg`. To reflect that a summation was performed, it would be nice to label the second output column `"SUM(" + rhsAgg + ")"`.

**This issue asks you to**

1. Adapt the implementation of the `groupJoin`-kernel such that the output aggregate column is labeled `"SUM(...)"`.
2. Adapt the frame label inference in `src/ir/daphneir/DaphneInferFrameLabelsOpInterface.cpp` accordingly.
3. Adapt the test cases in `test/runtime/local/GroupJoinTest.cpp`.

PR #478, opened 2023-03-03
Draft: [DAPHNE-#132] Frame column labels after group-join

DamianDinoiu:
Hello,

My name is Damian Dinoiu and I will join the project for my master's thesis.

Here is an implementation for my first-good issue: [DAPHNE-#132] Frame column labels after group-join

- Modified the labeling for the second column of the group-join operation to reflect the summation.
- Adapted the frame label inference for consistency.
- Modified the test case to reflect the changes.

Thanks, Damian.


pdamme (approved):
Hi @DamianDinoiu, thanks a lot, **it looks very good to me**. Congrats on your first contribution to DAPHNE, we're looking forward to your upcoming work in the system!

Issue #423, opened 2022-08-26
Timing - Profiling support


aristotelis96:
Both the end-user and developers of DAPHNE should be able to get timing statistics regarding various steps of execution and get insight and understanding on where most time is spent. A few examples could be:

- Time spent compiling and executing a DSL script.
- Time spent executing dividual pipelines.
- Time spent executing individual kernels (potentially per pipeline).

This issue is to further discuss the best way to approach this. Some examples for discussion:

- Architectural design.
- Libraries or frameworks that could be used.
- Where we should measure time (e.g., Before/after calling a kernel etc.).
- Ways to inform the user about the result (e.g., output file, printing, etc.) and what additional data should be provided (e.g., pipeline code, kernel name, etc.).

jonnygiger:
When it comes to outputting the result to a file, Clang has a very clean implementation of this worth taking a look at:

[https://github.com/aras-p/llvm-project-20170507/pull/2/files#diff-8289d7b6556cc50e792f2c015a05b3efbeba5eacf8e5319b39cb1f082f1b71b5](https://github.com/aras-p/llvm-project-20170507/pull/2/files#diff-8289d7b6556cc50e792f2c015a05b3efbeba5eacf8e5319b39cb1f082f1b71b5)

pdamme:
Timing for the high-level steps of DAPHNE has been addressed in 004e67bda6c4efd0cae43f258625c7609ad7b578, which introduced the `--timing` command-line argument.

Moreover, this issue is related to #227, which was addressed in d4cf91031dcebe2467ca9b07c9e651c77f3917ff by introducing the `--statistics` command-line argument.

This timing information is not used by system-internal components (e.g., scheduling) yet, but I think, we could close this issue nevertheless and create more specific follow-up issues when necessary.

So I will close this issue now, but feel free to re-open if you disagree.

PR #479, opened 2023-03-05
Add profiling support

psomas:
This is an initial attempt at adding profiling support to Daphne. As a first draft, I've added support for the [PAPI library](https://icl.utk.edu/papi/).

There's a profiling compiler pass that adds profiling 'tracepoints' at the beginning of each block, which call the Start/StopProfiling kernels.  This is more of a placeholder and we should decide exactly how / where this should be enabled. This can be exposed at the DSL / IR level via a pair of (start/stop) profiling operations.

The profiling kernels only support the HL (high-level) PAPI API at the moment, but we can switch to the low-level API if needed. PAPI also supports various plugins (e.g. for CUDA) which we can consider using / enabling.

Finally, we should also discuss whether PAPI is good enough or if we'd like to consider / evaluate other profiling tools / frameworks.

The documentation on how to use the new profiling infrastructure is also missing at the moment. You can try it by passing the   _--enable-profiling_ CLI option to Daphne and settings the relevant PAPI HL environmental variables, e.g.:
`PAPI_EVENTS="perf::CYCLES,perf::INSTRUCTIONS,perf::CACHE-REFERENCES,perf::CACHE MISSES,perf::BRANCHES,perf::BRANCH-MISSES" PAPI_REPORT=1 ./daphne --enable-profiling script.daph`

This PR partly addresses #423

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -397,6 +397,7 @@ arrowVersion=11.0.0
 openMPIVersion=4.1.5
 eigenVersion=3.4.0
 spdlogVersion=1.11.0
+papiVersion=7.0.1
 
 #******************************************************************************
 # Set some prefixes, paths and dirs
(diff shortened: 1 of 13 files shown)


corepointer:
Is this ready to be merged now @psomas? You could confirm by removing the WIP in the PR title. Don't worry about the failing test - that's because the github-action docker image does not have libpapi.so (only libpapi.a). I'll fix that when merging.

corepointer:
LGTM - thx for your contribution @psomas. I added a few minor corrections and cleanups. Please add documentation about this (e.g., the text you have written in this PR description and possibly more) to the documentation section (doc sub directory) in another commit.

PR #480, opened 2023-03-05
[BUGFIX] Use the static versions of Arrow and Parquet libs

psomas:
With our current build system, if we use a shared thirdparty library, the resulting binary (executable or shared library) depends on the absolute path of the shared library. This means that it might not be possible to run Daphne on a host other than the one we built it on.

### DIFF
--- a/src/runtime/distributed/worker/CMakeLists.txt
+++ b/src/runtime/distributed/worker/CMakeLists.txt
@@ -40,8 +40,8 @@ set(LIBS
         CallData
         Proto
         DaphneMetaDataParser
-        Arrow::arrow_shared
-        Parquet::parquet_shared
+        Arrow::arrow_static
+        Parquet::parquet_static
         )
 
 add_library(WorkerImpl ${SOURCES})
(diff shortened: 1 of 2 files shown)


corepointer:
I can not confirm this. Maybe you did not properly set up your linker environment (LD_LIBRARY_PATH) ?

psomas:
Yes, sorry, it's not the absolute path, setting the LD_LIBRARY_PATH works, but we should maybe document it. We could consider "installing" / copying the thirdparty shared libraries we link against under lib/, so that having bin/ and lib/ is enough to run Daphne (now the thirdparty/ folder is also a runtime dependency instead of a build-only dep). Although, if the static versions are small enough, we could switch to them in any case.

corepointer:
The only place I ever documented it was in the download page of release 0.1 :blush: When using the interactive container all dependencies are in /usr/local/lib - that is conveniently one of the default locations that is searched anyway ;-)

I can see the benefits of linking statically and I'm not opposing it right away. But linking Arrow/Parquet would probably increase libAllKernels.so's size to more than double (libAllKernels.so is 40mb, libarrow.a is 55mb and libparquet.a is 9mb). Next up in the latest dependency additions will be Eigen, PAPI, OpenMPI and spdlog. As you can see this quickly gets out of hand imho, so documenting the linker setup might be more scalable.

psomas:
k, closing this PR then. I'll also probably modify the profiling PR to link PAPI dynamically instead of statically.

PR #489, opened 2023-03-30
Added load partioning for distributed runtime.

aristotelis96:
Abstracted duplicate code for different distributed runtime communication frameworks.
Added `distributed/coordinator/scheduling/LoadPartitioningDistributed.h` class which is used to:
- Split ranges based on some policy. For now only broadcast and simple distribution is supported.
- Sets the metadata of the matrix being distributed/broadcasted.
- Contains a static method which sets the metadata of the outputs during the computation step.
- This class can be extended in order to support more complex distribution strategies.

### DIFF
--- a/src/runtime/distributed/coordinator/kernels/Broadcast.h
+++ b/src/runtime/distributed/coordinator/kernels/Broadcast.h
@@ -26,6 +26,7 @@
 #include <runtime/local/datastructures/Range.h>
 #include <runtime/distributed/worker/WorkerImpl.h>
 #include <runtime/distributed/proto/DistributedGRPCCaller.h>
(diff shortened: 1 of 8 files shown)


corepointer:
Thanks for your contribution @aristotelis96 ! In the future, please add a reference to the issue/PR number to the first line of the commit message (e.g., "[DAPHNE-#489] Added load partitioning for distributed runtime"). That helps to reference any discussion that might relate to the change.  Furthermore, if you merge/push from your local copy, then adding "Closes #489" in the last line of the commit message automatically changes the status of the open issue/PR to closed or merged.
Best regards, Mark

PR #496, opened 2023-04-06
[DAPHNE-#492] Creating Subframe of Frame with zero rows 

### LINKED ISSUE #492: Creating Subframe of Frame with zero rows
The constructor, which creates a Frame around the sub-frame of another Frame, throws an error, if the Frame has no rows, even if the new Frame should also have zero rows.

Code example:
`f = createFrame(
    [1, 10, 20],
    "temp");

registerView("sample", f);

k = sql("select sample.temp from sample
 where sample.temp <= 0;");

print(k);`

The compiler first creates a filter row Kernel, which returns a frame with zero rows. Then it calls Kernel to extract the column temp from that frame, which throws the aforementioned error.

tomschw:
Fixed problem:

- Crash when trying to use the Frame constructor which creates a sub-frame without copying data, when using a source Frame with zero rows.

Changes:

- Add test which reproduces the problem.
- Fix by excluding the assertions of the constructor when source Frame has zero rows and the expected boundaries of the new frame are both zero.

Fixes #492.


pdamme (approved):
Thanks for this fix, @tomschw. It **looks very good to me**. Congrats on your first contribution to DAPHNE!

Issue #505, opened 2023-04-20
ANTL-jar not found, when path or version changes changes 


EricMier:
In CMake the location of the ANTL-jar file, once found, is stored in a **cached** CMake variable, see linked line:

https://github.com/daphne-eu/daphne/blob/71ffea1179f4dbdedb621eb4ecf0f5db267e7472/src/parser/CMakeLists.txt#L40

This leads to an error, if the location of the jar-file changes and/or if we decide to upgrade to a newer version. This will not occur in a freshly cloned environment, but in existing projects. The way it is defined currently, does not override the variable **ANTLR4_JAR_LOCATION**, if the old file-path does not exist. You need to force set the cached variable to update the path.

PR #506, opened 2023-04-20
Fix: Force set the cached variable ANTLR4_JAR_LOCATION, if old path not exists

EricMier:
See Issue #505


corepointer (approved):
Thx for catching this one @EricMier I'll merge it right away :)

Issue #507, opened 2023-04-20
Path changes of dependencies can lead to errors on already built projects


EricMier:
Currently we check if a dependency is already downloaded/installed via a flag-file. But this does not tackle path changes of dependencies on existing projects.

https://github.com/daphne-eu/daphne/blob/71ffea1179f4dbdedb621eb4ecf0f5db267e7472/build.sh#L572-L583

The build script checks only if the flag exists and does not do a fresh install in the updated path and therefore leads to errors while compiling assuming non existing files.

My suggestion is to add a virtual version to each dependency, which we increase each time a path updates. Alternatively we check the assumed paths, but that could lead into more overhead.

corepointer:
Despite the build system getting increasingly complex, I think it's a nice idea :)

corepointer:
This issue should have been closed automatically when we merged #508

PR #508, opened 2023-04-20
Dev-issue-507-dependency-path-changes

EricMier:
Added a new internal version variable for updating thirdparty dependencies.

Issue:
- when downloading and installing thirdparty dependencies, a internal flag is used to check, if a dependency was successfully downloaded/installed to increase stability of the build process
- this leads to issues on existing projects, if there are path changes on a newer version of DAPHNE
- since the success-flag is there, the dependency is not downloaded/installed to the new paths, which leads to errors when building the DAPHNE-target

Solution:
- new internal variable to keep track of path changes
- if paths specific for a dependecy change, this version is increased
- runs on existing project will detect this change and redownload/-install automatically
- (temporary solution, might be handled by cmake later)

Solves #507

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -14,13 +14,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
+# This is the main build script of Daphne. It downloads and builds all third-party dependencies and then builds Daphne.
+# It is required if you freshly cloned the repository or if you want to update the third-party dependencies.
+# It is not required if you only want to build Daphne from your IDE for example after you already built it once.
+# In this case you need to adapt the cmake settings according to the cmake call in section 9.
+#
+# For more information about the build script and contribution guidelines please refer to the doc/development/BuildingDaphne.md
+
+
 # Stop immediately if any command fails.
 set -e
 
 build_ts_begin=$(date +%s%N)
 
 #******************************************************************************
-# Help message
+# #1 Help message
 #******************************************************************************
 
 function printHelp {
@@ -53,7 +62,7 @@ function printHelp {
 }
 
 #******************************************************************************
-# Build message helper
+# #2 Build message helper
 #******************************************************************************
 # Daphne Colors
 daphne_red_fg='\e[38;2;247;1;70m'
@@ -204,7 +213,7 @@ function printLogo() {
 }
 
 #******************************************************************************
-# Clean build directories
+# #3 Clean build directories
 #******************************************************************************
 
 function clean() {
@@ -294,14 +303,7 @@ function cleanDeps {
 
     local dirs=("${buildPrefix}" "${installPrefix}")
     local files=(
-        "${thirdpartyPath}/absl_v"*".install.success"
-        "${thirdpartyPath}/antlr_v"*".install.success"
-        "${thirdpartyPath}/catch2_v"*".install.success"
-        "${thirdpartyPath}/grpc_v"*".install.success"
-        "${thirdpartyPath}/nlohmannjson_v"*".install.success"
-        "${thirdpartyPath}/openBlas_v"*".install.success"
-        "${thirdpartyPath}/llvm_v"*".install.success"
-        "${thirdpartyPath}/arrow_v"*".install.success"
+        "${thirdpartyFlagsDir}/"*".install."*".success"
         "${llvmCommitFilePath}")
 
     clean dirs files
@@ -313,11 +315,7 @@ function cleanCache {
     local dirs=("${sourcePrefix}" "${cacheDir}")
 
     local files=(
-        "${thirdpartyPath}/absl_v"*".download.success"
-        "${thirdpartyPath}/antlr_v"*".download.success"
-        "${thirdpartyPath}/grpc_v"*".download.success"
-        "${thirdpartyPath}/openBlas_v"*".download.success"
-        "${thirdpartyPath}/arrow_v"*".download.success"
+        "${thirdpartyFlagsDir}/"*".download."*".success"
     )
 
     clean dirs files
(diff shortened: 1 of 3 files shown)


build.sh
```diff
@@ -313,20 +313,16 @@ function cleanCache {
...
 
     clean dirs files
 }
 
 function cleanAll {
     cd "$projectRoot"
-    message="This will delete the DAPHNE build output and everyting in $thirdpartyPath and reset this directory to its \
-last state in git."
+    message="This will delete the DAPHNE build output and everyting in $thirdpartyPath and reset this directory to its"
+    message+=" last state in git.\n"
```
EricMier, line 325: Just because with multiline string I couldn't collapse the function in VSCode, which bothered me somehow :D

build.sh
```diff
@@ -313,20 +313,16 @@ function cleanCache {
     local dirs=("${sourcePrefix}" "${cacheDir}")
 
     local files=(
-        "${thirdpartyPath}/absl_v"*".download.success"
-        "${thirdpartyPath}/antlr_v"*".download.success"
-        "${thirdpartyPath}/grpc_v"*".download.success"
-        "${thirdpartyPath}/openBlas_v"*".download.success"
-        "${thirdpartyPath}/arrow_v"*".download.success"
+        "${thirdpartyFlags}/"*".download."*".success"
```
corepointer, line 316: Can this be included in "dirs" now?

build.sh
corepointer: If it puts your mind at rest so be it :D

build.sh
```diff
@@ -679,19 +693,21 @@ if [ $WITH_DEPS -gt 0 ]; then
     # abseil (compiled separately to apply a patch)
     #------------------------------------------------------------------------------
     abslPath=$sourcePrefix/abseil-cpp
-    if ! is_dependency_downloaded "absl_v${abslVersion}"; then
+    dep_absl=("absl_v${abslVersion}" "v1")
```
corepointer, line 696: What about putting all the "v1" in a central place like the version numbers? But it's not really important to change it. Just thinking out lout =)

corepointer (approved):
LGTM but I did not test-run it tbh.
Did I get that right: The sequence number v1 should be incremented if something (paths?) change but the version number does not? That was somehow related to the ANTLR problems? I wonder if it helps anything if the one causing the issue does not realize it right away and therefore does not increment that number.

build.sh
EricMier: No, this we only want to reset the downloaded files in clearCache() if I understood it correctly. If we delete the dir, all dependencies have to be built again.

build.sh
EricMier: I also thought about that, but I don't think its that benefitial. Since the changes related to a dependency are probably done, where this (virtual) version is defined. My hope is to reduce the probability to overlook this necessary change, when working on a dependency.

EricMier:
@corepointer yes you got that right. Basically an addition to force existing projects to update their dependencies, in case something fundamental (like its path) is changing.

It was not related to the ANTLR problem from my previous PR, because if the ANTLR_JAR_LOCATION is already in my local CMake-Cache, it wouldn't be overriden anyway.

If someone don't realize the necessity to change that version number when causing potential problems, we still can point that out in their PR :)

build.sh
corepointer: Ah yes, sorry my bad. But I found another little glitch to nitpick about :D The install.success files in cleanDeps() still need this modification iirc. I'll fix that one when merging tomorrow morning.

pdamme (changes requested):
Thanks @EricMier for proposing this fix.

**I agree with your general motivation**: Whenever we change something about how a third-party dependency is built (e.g., paths, static/dynamic, potentially compiler flags), but not the version itself, then an attempt to rebuild DAPHNE in an existing clone can fail. This is inconvenient for developers. To my mind, it is not a critical problem, because there are simple solutions to it, such as:
- clean and build (usually a good approach when a build fails, especially after pull)
- notify all developers via dev mailing list about the change, so they are not surprised (admittedly, we need to settle on a mailing list first, plus people might forget or oversee it)

Still, yes, automating the adaptation to changes would be more convenient. So all fine. However, I see your approach rather as a workaround. **My main concern is that we are re-inventing typical features of build-tools like `make`/`cmake`**, which are good at noticing that something has changed and triggering the necessary steps to rebuild. Achieving this effect by managing an additional virtual version number manually is a bit unintuitive and therefore hard to understand. I wonder if it would be better to handle our third-party dependencies in `cmake` in some way.

However, **since employing `cmake` in the right way might take some effort, I would be fine with accepting the contribution as *a temporary solution***.

**Required changes before a merge:**

**1. Proper documentation.**
- Please update `doc/development/BuildingDaphne.md` (section "Adding a dependency").
- Please update the doc strings of `dependency_install_success`  / `dependency_download_success` and `is_dependency_installed`  / `is_dependency_downloaded` in `build.sh` to reflect that they have two parameters now.
- Please provide a few bullet points as the description of the PR so we can use them for the commit message (mention all things that have changed).

**2. Make sure it builds when old build artifacts are present.**
- When I try to build DAPHNE with the updated build script on top of existing build artifacts, I run into problems with OpenBLAS. In detail, after the message `[DAPHNE]..Get OpenBlas version 0.3.19`, OpenBLAS asks me the following kind of question for dozens of files:
  ```
  replace /daphne/thirdparty/sources/OpenBLAS-0.3.19/.drone.yml? [y]es, [n]o, [A]ll, [N]one, [r]ename:
  ```
- This should work out of the box. Otherwise, the initial goal of the fix is somewhat contradicted.
- Can you reproduce this problem?

corepointer:
I added some code to "migrate" over to the new flag files. This should work around the issue 2) of @pdamme. Normally one would run --cleanCache and get rid of the unpacked source dirs to not run into the issue you're having.

corepointer:
This is a friendly reminder about the pending change request @EricMier  ;-)

EricMier:
Finally found time to incorporate your requested changes. Sorry for the long wait. Also updated to current main. Sorry for the long wait.

corepointer (approved):
LGTM - Thank you for finishing this up @EricMier. I made a few minor adjustments (e.g., you documented an error I introduced - there's no "--with-deps", only a "--no-deps") and squashed the PR into two commits as the little fix for the sql script doesn't seem to match the topic of the PR. 
Will merge after a final run of the build script on a fresh clone completes.

PR #509, opened 2023-04-25
[DAPHNE-#493] SQLParser: Run Aggregation without grouping

### LINKED ISSUE #493: SQLParser: Run Aggregation without grouping
Currently it is not possible to run a aggregation over a whole frame without grouping over some column in the frame when using SQL.

Not working code:
```
f = createFrame(
    [1, 10, 20],
    "a");

registerView("sample", f);

k = sql("select sum(sample.a) from sample;");

print(k);
```

Current workaround:
```
f = createFrame(
    [1, 10, 20],
    [0, 0, 0],
    "a", "temp");

registerView("sample", f);

k = sql("select sum(sample.a) from sample GROUP BY sample.temp;");

print(k);
```

tomschw:
Fixed problem:

- To use aggregations in a SQL query we currently have to use some kind of grouping, aggregation over a whole column is therefore only possible with a workaround.

Changes:

- Add test which reproduces the problem.
- Implement functionality by using corresponding already existing operations (AggAll or for Count the NumRow Operation).
- Change the name of the temporary column that gets created when converting to int matrix.

Fixes #493.


pdamme:
Thanks @tomschw for this fix. I will have a close look by the end of the week.

PR #536, opened 2023-05-11
build: Correct Apache Arrow DL URL.

psomas:
Apache Arrow 11.0.0 is now served from archive.apache.org, not dlcdn.apache.org, correct the DL URL accordingly.


pdamme:
Thanks for this fix @psomas. This was indeed a problem. However, it has already been fixed on main by cb3071c2a30c66ab8516af6d1c58016a058312e2. Therefore, I'm closing this pull request.

PR #539, opened 2023-05-23
Eigen op 391

KostasBitsakos:
Add of a Documentation and some changes in order for the Eigen Kernel to work with the DaphneOps.


corepointer:
There are duplicate commits in this PR and some other unrelated changes. I suggest adding the relevant changes to a fresh branch from main.

PR #543, opened 2023-06-12
[DAPHNE-#494] SQLParser: Group by without aggregation clause not working

### LINKED ISSUE #494: SQLParser: Group by without aggregation clause not working
Currently, when running an SQL query which is just grouping some columns and outputting them, it throws the following error:

```daphne: /daphne/thirdparty/installed/include/mlir/IR/TypeRange.h:140: mlir::Type mlir::ValueTypeRange<ValueRangeT>::operator[](size_t) const [with ValueRangeT = mlir::OperandRange; size_t = long unsigned int]: Assertion `index < size() && "invalid index into type range"' failed.```

Sample code to reproduce error:
```
f = createFrame(
    [1, 10, 10, 20],
    "a");

registerView("sample", f);

k = sql("select sample.a from sample GROUP BY sample.a;");

print(k);
```

tomschw:
Fixed problem:

- Grouping in SQL statements is only possible when having an aggregation in the outputs

Changes:

- Add tests to reproduce the problem.
- Fix the shape inference, which had problems, when no aggregation columns where used.
- Create a workaround in the Kernel lowering for the group op, when no aggregations are used.

Other possible solutions instead of the current workaround:

- Split up the GroupOp into a GroupOp with and without aggregations, handled by the SQLParser.
- Create a NoneOp as an aggregation function which occurs, when no aggregations where used (would need additional work in the shape inference and in the Group-Kernel).

As the Kernel lowering should be reworked in the future anyway, the current workaround should work fine for the moment.

Fixes #494.


pdamme (approved):
Thanks @tomschw, **this contribution fixes the bug**. I polished it just a bit.

There was indeed also a bug in the shape inference of `GroupOp`, but the problem there was not to check if the list of aggregation columns is empty. The actual problem was the use of `inferNumColsFromArgs()`, a shape inference helper function not suitable for this case. Instead, one should just take the `size()` of the list of key columns and aggregation columns. I adjusted it accordingly.

PR #550, opened 2023-06-23
Feature/docs mvp

m-birke:
Hello,

introducing a HTML page of the markown documentation under doc/ . Hosted by github pages it will be accessible later under https://daphne-eu.github.io/daphne . Since on my fork right now, it can currently be seen here https://m-birke.github.io/daphne. 

Points to discuss: 

- from which state should documentation be build/deployed ? I would actually propose a released state, would a release branch then b okay for you guys? Or since we don't have a 1.0 release just always from the main branch?
- if deploying a released state, version information should be visible in docs. Can be accomplished by using the git describe functionality, put another way using the last git annotated tag as version. Any thoughts to that?


auge:
If removing the copyright header of individual files is a problem, they could be put back using comments in markdown: https://stackoverflow.com/questions/4823468/comments-in-markdown#20885980

corepointer:
First of all thanks for all the hard work on the documentation!

In the current state of the project where many things are in flux I think it's not so good to depend on releases with the update of documentation. That would delay any documentation by half a year or more if we look at our current release habits.

pdamme (approved):
Thanks @m-birke for this contribution! The generated HTML pages make the documentation much more readable (and searchable), which is a great step forward for the project. Thanks also for tidying up all the markdown files so thoroughly! Overall, this contribution looks very good to me. (I have just added some minor fixes.)

**From my point of view, we could merge it as it is.** Nevertheless, there are a few points that would be great to address, but that could be done after the merge:
1. It would be great if we could keep the licence headers in the markdown files. I acknowledge that they may cause problems with the HTML generation, but maybe there is a workaround. On the one hand, HTML comments seem to be fine for the HTML generator in general, since several of the markdown files contain HTML comments. On the other hand, maybe the approach mentioned by @auge can be used.
2. Though you did a great job escaping the occurrences of `<` and `>` in the markdown files, some of them still don't render correctly in HTML (for instance in `BuildingDaphne.md`). It would be great if you could double-check them. However, as this point is not strictly related to the generation of the HTML docs, it can be addressed after the merge.
3. The file `/doc/DaphneDSLBuiltins.md` was essentially moved to `doc/DaphneDSL/Builtins.md` (and changed). However, in the history it shows as delete+create, unlike other moved+changed files in this PR. If it doesn't casue too much inconvenience, it would be great to address this point to keep the history of that file.

**Please let me know if you want to address any of these points before the merge.** Otherwise, I would merge it in and any additional improvements can be done afterwards.

-----

**Regarding the code version** to build the HTML documentation for, I largely agree with @corepointer: it is currently most sensible to reflect the latest version of the documentation. At the same time, I appreciate that users of a binary release may want to browse the documentation for that version. I think a great solution would be to provide both (docs of each release *plus* docs of the current main branch). Some other repos/systems solve this by a drop-down for the version (see, e.g., (https://docs.python.org/3/)). Such a feature could be added after the merge.

auge:
> 

git actually tracks content, not files. so different git clients may show it differently.
e.g. in TortoiseGit, the files are shown as renamed (and git anyway tracks it correctly ...)

image

m-birke:
> Thanks @m-birke for this contribution! The generated HTML pages make the documentation much more readable (and searchable), which is a great step forward for the project. Thanks also for tidying up all the markdown files so thoroughly! Overall, this contribution looks very good to me. (I have just added some minor fixes.)
> 
> **From my point of view, we could merge it as it is.** Nevertheless, there are a few points that would be great to address, but that could be done after the merge:
> 
> 1. It would be great if we could keep the licence headers in the markdown files. I acknowledge that they may cause problems with the HTML generation, but maybe there is a workaround. On the one hand, HTML comments seem to be fine for the HTML generator in general, since several of the markdown files contain HTML comments. On the other hand, maybe the approach mentioned by @auge can be used.
> 2. Though you did a great job escaping the occurrences of `<` and `>` in the markdown files, some of them still don't render correctly in HTML (for instance in `BuildingDaphne.md`). It would be great if you could double-check them. However, as this point is not strictly related to the generation of the HTML docs, it can be addressed after the merge.
> 3. The file `/doc/DaphneDSLBuiltins.md` was essentially moved to `doc/DaphneDSL/Builtins.md` (and changed). However, in the history it shows as delete+create, unlike other moved+changed files in this PR. If it doesn't casue too much inconvenience, it would be great to address this point to keep the history of that file.
> 
> **Please let me know if you want to address any of these points before the merge.** Otherwise, I would merge it in and any additional improvements can be done afterwards.
> 
> **Regarding the code version** to build the HTML documentation for, I largely agree with @corepointer: it is currently most sensible to reflect the latest version of the documentation. At the same time, I appreciate that users of a binary release may want to browse the documentation for that version. I think a great solution would be to provide both (docs of each release _plus_ docs of the current main branch). Some other repos/systems solve this by a drop-down for the version (see, e.g., (https://docs.python.org/3/)). Such a feature could be added after the merge.

@pdamme thanks for the feedback. I addressed point 1 (license headers work with another gh workflow) and 2 (change syntax a bit) so far.

m-birke:
@pdamme License headers are in again. Angle brackets work. No conflicts. Merge please :)

pdamme:
> git actually tracks content, not files. so different git clients may show it differently. e.g. in TortoiseGit, the files are shown as renamed (and git anyway tracks it correctly ...)

@auge Yes, you are right! Thanks.

pdamme:
All three points mentioned in my initial review have been addressed. **The PR looks very good**, and we could merge it from my point of view. However, **the action for building the docs has failed**. Could you have a look at that again please, @m-birke?

corepointer:
> All three points mentioned in my initial review have been addressed. **The PR looks very good**, and we could merge it from my point of view. However, **the action for building the docs has failed**. Could you have a look at that again please, @m-birke?

This seems to be a permission issue (initial setup for deployment required). I'll look into it.

corepointer:
I followed the documentation of that github action to set up the GITHUB_TOKEN authentication method. The source branch is now main, so I guess the action will only complete successfully *after* merging to main?

pdamme:
Okay, sounds good. We could just try it by merging and revert if it doesn't work ;) . Shall I?

corepointer:
> Okay, sounds good. We could just try it by merging and revert if it doesn't work ;) . Shall I?

Even though this PR does not show any conflicts, I ran into a number of them while squashing. I hope I resolved them well :see_no_evil: But I'm through now and will do the merge.

PR #561, opened 2023-07-03
docs: Document the PAPI profling support

### LINKED ISSUE #552: Documentation for Profiling
I recently noticed some files and the command-line option for profiling support.
Depending on the extent of its functionality I would like to use it for benchmarks.
As I think it would also be useful for other users, documentation with the following information for new users would be very helpful:

- How to enable it
- What can it do and what can't it do (For example does it support calculating the runtime of each Kernel, does it support multiple runs etc.)
- Does it only support part of DAPHNE or everything?

psomas:
Closes #552.

### DIFF
--- a/doc/Profiling.md
+++ b/doc/Profiling.md
@@ -1,0 +1,40 @@
+<!--
+Copyright 2023 The DAPHNE Consortium
+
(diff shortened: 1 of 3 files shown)


corepointer:
LGTM, thanks @psomas for taking care of this.

PR #563, opened 2023-07-04
[DAPHNE-#554] Use hwloc to query CPU topology

### LINKED ISSUE #554: non-portable cpu query in get_topology()
The method get_topology(...) in MTWrapper.h is parsing the output of ``/proc/cpuinfo`` which seems to be tailored to x86-64 processors as this fails on ARM.

A proper fix for this was already discussed to be the use of the hwloc library [1]. Until that solution is in place there will be a quick-fix that makes ARM run (filling the vectors to be filled by that method with generic IDs).

[1] https://www.open-mpi.org/projects/hwloc/

psomas:
Closes #554

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -404,6 +404,7 @@ openMPIVersion=4.1.5
 eigenVersion=3.4.0
 spdlogVersion=1.11.0
 papiVersion=7.0.1
+hwlocVersion=2.9.3
 
 #******************************************************************************
 # #6 Set some prefixes, paths and dirs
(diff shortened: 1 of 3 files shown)


corepointer:
Tests fail because of the missing library in the github action docker image. Is this the only reason why this PR is in draft state?

psomas:
I've only run this on a single-node x86 server. I'll do some tests on the NUMA ARM server we have in the lab, and if everything works out, I'll mark it ready.

The topology discovery could be improved upon but we can merge this PR and open a new issue about expanding the hwloc / topology discovery support.

psomas:
It seems to also work on our dual-socket ARM server, so I think it can be merged. I've also created #569 to track any progress with further integrating hwloc into the runtime / vectorized engine.

Issue #562, opened 2023-07-04
LLVM lowering function with variable number of operands


tomschw:
I've found a bug in the LowerToLLVMPass. In my case I wanted to use the Group Kernel twice: once for a grouping aggregation over certain columns and once without any aggregations. Both of these have a different number of inputs regarding the Kernel: with aggregation we have 9 inputs, without we have 7.

The problem occurs due to the "registering" of llvm.funcs. As it first finds in my sample script a  Group By with the aggregations it creates the following llvm func:
`llvm.func @_group__Frame__Frame__char_variadic__size_t__char_variadic__size_t__GroupEnum_variadic__size_t(!llvm.ptr<ptr<i1>>, !llvm.ptr<i1>, !llvm.ptr<ptr<i8>>, i64, !llvm.ptr<ptr<i8>>, i64, !llvm.ptr<i32>, i64, !llvm.ptr<i1>)`
This works for the first call of the kernel. But the second time it tries to call the Kernel the following way:
`llvm.call @_group__Frame__Frame__char_variadic__size_t__char_variadic__size_t__GroupEnum_variadic__size_t(%103, %92, %96, %98, %100, %101, %43) : (!llvm.ptr<ptr<i1>>, !llvm.ptr<i1>, !llvm.ptr<ptr<i8>>, i64, !llvm.ptr<i32>, i64, !llvm.ptr<i1>) -> ()`
As the number of inputs don't match, it throws an error.

The problem can also easily be reproduced in the other direction (including the function with 7 operands) with the following script:
```
f = createFrame(
    [  0,  1,  2,  3, 3,  4,  5,  6,  7,  8,  9],
    "a");

registerView("f", f);

res = sql("SELECT f.a FROM f GROUP BY f.a;");

print(res);

res2 = sql("SELECT sum(f.a) FROM f GROUP BY f.a;");

print(res2);
```

philipportner:
A fix was implemented as part of the following PR: https://github.com/daphne-eu/daphne/pull/564. Thanks again @tomschw !

PR #564, opened 2023-07-04
SQLParser: Introduce DISTINCT clause

tomschw:
New feature:

- Allow the use of DISTINCT clause in SQL statements to only return unique rows
- Allow the aggregation word in SQL commands to be any type of lower/upper case

Changes:

- Introduce  DISTINCT in the SQLParser
- Allow the use of the Group Kernel with "*" (group over all columns in the input frame) (not yet in SQL CLI)

Update:
- Fixes issue #562. Therefore Group by aggregations are now compatible with DISTINCT


philipportner:
Thank you @tomschw for your contribution! LGTM, I merged your PR and I'll close the fixed issue.

PR #570, opened 2023-07-07
[doc] fix md flaws and add newely added docs to html page

m-birke:
Hey @corepointer 

this fixes some stuff which went wrong during conflictr solving (it was not much)

Additionally, newely added docs are added.

@pdamme @corepointer it is documented how to write and add docs (/doc/development/writeDocs.md). It would be nice to raise attention for this of the other developers. I wont fix the md documents all the time ;)


corepointer:
Thx for fixing this and sorry for the trouble. Only changes applied were a rebase on main and a more elaborate commit message. We'll add the pointer to the documentation documentation in the next documentation reminders ;-)

PR #571, opened 2023-07-10
Pass errors are not only logged, but also printed again.

pdamme:
- For a long time, pass errors, which happen inside the DAPHNE compiler, used to be printed to std::cerr.
- Since a recent commit (4a8f003cf5b499bec4ecbb138441af8dfa2413f0), they are logged using the recently introduced logger, instead.
- Not printing the error messages to std::cerr anymore has at least two significant downsides:
  - Users don't see the error messages anymore. Instead, DaphneDSL scripts may just silently fail.
  - The error output captured by our script-level test cases is useless, since it is empty.
- This commit adds the printing of the error messages to std::cerr again, but also retains the logging of the errors.

### DIFF
--- a/src/api/internal/daphne_internal.cpp
+++ b/src/api/internal/daphne_internal.cpp
@@ -516,14 +516,17 @@ int startDAPHNE(int argc, const char** argv, DaphneLibResult* daphneLibRes, int 
     }
     catch(std::runtime_error & re) {
         spdlog::error("Pass std::runtime_error: {}", re.what());
+        cerr << "Pass error: " << re.what() << endl;
         return StatusCode::PASS_ERROR;
     }
     catch(std::exception & e) {
         spdlog::error("Pass std::exception: {}", e.what());
+        cerr << "Pass error: " << e.what() << endl;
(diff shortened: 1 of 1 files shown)


pdamme:
@corepointer Please let me know if there was a good reason to remove the printing of the error messages to `std::cerr`. I assume it was just a mishap, so I'm merging this myself now :) .

corepointer:
It was a partial mishap :-P I replaced std::cerr with spdlog::error on purpose but did a bad job in maintaining the expected behavior.

The ultimate goal of our logging facility is to remove these uncontrollable print outs (cout/cerr). I know some of our test cases rely on console output and the default logger configs that I set when merging it in are by far not in a solid state, but since the test cases completed without error when I tested I thought it's good enough for now. 

Some things that come to my mind that would improve the situation:
* Better defaults than log level OFF
* a separate named logger that prints to std err and maybe can even avoid being muted by log_level_limit for exceptions and/or script errors
* document when to require std err output

pdamme:
Fair enough! For those who are interested: See #575 for more on that topic.

PR #574, opened 2023-07-15
522 DaphneLib complex control flow and lazy evaluated functions

lachezar-n:
This pull request consists of changes concerning only DaphneLib and only the Python source code.
It implements context methods for creating structures like `if-then-else`, `for-loops`, `while-loops`, and `do-while-loops`, applying logical operators and defined user-defined functions. 
It allows manipulating only matrices except for the user-defined functions (that can manipulate all kinds of objects) and the logical operators (that handle only scalars). It allows also manipulating multiple objects in the same structure/function, but only a single object can be computed at the end.
The branch contains also Pytests for unit-testing the context methods. (run `pytest test/api/python/pytest` for simple execution).


pdamme:
Thanks for this contribution, @lachezar-n. Support for complex control flow structures in DaphneLib is a super useful feature in DaphneLib, which will allow users of DAPHNE to implement more complex integrated data analysis pipelines in Python. It's also a great timing, since our use case partners are about to start working with DaphneLib.

I will have a close look into your code and provide feedback in the next days.

pdamme (changes requested):
Hi @lachezar-n! Overall, **your contribution looks very good to me**. The control flow structures work as expected (except for user-defined functions, see below), there is a strong set of test cases, and at least a minimal documentation. **Very good work!**

I've rebased your branch on main to get the recent bug fixes regarding nested control flow in DaphneDSL. With that and a few minor fixes I've applied to your code, all your test cases pass successfully, including the ones on nested control flow in DaphneLib. Feel free to have a look at my changes in the commit I've added to your branch.

**For passing the AMLS project, please reply to my two questions and please address the first required change (comments), see below.**

*Feel free to let me know if you would like to address the required change regarding the test cases (otherwise, I will do that) or any of the optional changes (otherwise, they will become follow-up issues for later). These further improvements would strengthen your contribution to DAPHNE and would be highly appreciated, but they are completely optional for you.*

-----

**Questions:**
1. Is there a specific reason why you added spaces around the operator symbols in `consts.py`, `matrix.py`, and `scalar.py` (e.g., `"+"` to `" + "`)? Is that a fix for the problem regarding the `-`-operator and DaphneDSL parsing? If so the spaces should rather be inserted in a central place in the DaphneDSL code generation.
2. In `test_for_loops.py`: `test_invalid_callback_outputs()`: Why is `output[0].compute()` considered wrong?

**Required changes before we can merge this PR:**
1. Please add **a few comments** in `__init__()` and `code_line()` of the new `operation_node` subclasses (`cond` etc.), to help people understand what they do.
2. Thanks for the good set of **test cases**. While using *pytest* would be fine in general, it is a bit suboptimal in the context of the overall DAPHNE codebase, since we use *catch2* for all test cases (unit tests for C++ code and script-level tests for DaphneDSL and DaphneLib code). It would be great to stick with a single test framework to keep creating and running tests straightforward for all developers. Furthermore, it would be preferable to avoid adding *pytest* as an additional dependency. Thus, your test cases should be converted to the form used for the existing DaphneLib test cases. Essentially, this means splitting each of your test cases into two scripts, one that uses only DaphneLib and one that uses numpy (or DaphneDSL). Furthermore, these test cases must be added in `test/api/python/DaphneLibTest.cpp`, such that the outputs of the DaphneDSL and the numpy/DaphneDSL scripts are compared (for a unified output format, all prints could happen in DAPHNE).

**Optional additions:**
1. **User-defined functions do not work as expected yet.** Your current solution decorates ordinary Python functions to ensure that the result is a tuple, and you call functions the normal way in Python. However, just like using Python control flow structures, this approach does not produce function calls in DaphneDSL, but inserts the function body directly at the call site. As an example, consider the following DaphneLib/Python code:
   ```python
   from api.python.context.daphne_context import DaphneContext
   
   dctx = DaphneContext()
   
   @dctx.function
   def foo(X):
       return (X * 2 + 1).sqrt()
   
   A = dctx.fill(1.1, 2, 2)
   A2 = foo(A)
   
   B = dctx.fill(3.3, 2, 2)
   B2 = foo(B)
   
   C = A2[0] + B2[0]
   
   print(C.compute())
   ```
   It results in the following DaphneDSL script:
   ```text
   V0=fill(1.1,2,2);
   V1=V0 * 2;
   V2=V1 + 1;
   V3=sqrt(V2);
   V4=fill(3.3,2,2);
   V5=V4 * 2;
   V6=V5 + 1;
   V7=sqrt(V6);
   V8=V3 + V7;
   saveDaphneLibResult(V8);
   ```
   But we would expect something like:
   ```text
   def foo(X) {
       V0=X*2;
       V1=V0+1;
       V2=sqrt(V1);
       return V2; 
   }
   V0=fill(1.1,2,2);
   V3=foo(V0);
   V4=fill(3.3,2,2);
   V7=foo(V0);
   V8=V3 + V7;
   saveDaphneLibResult(V8);
   ```
   I think, we need something like the following API in DaphneLib:
   - `dctx.def([X], body_fn)`: generates a DaphneDSL user-defined function
   - `dctx.call(funcName, [X])`: generates a call to the user-defined function with the given name
2. **Support for changing scalars in DaphneLib control flow structures.**
   You say it only works with matrices (in fact, frames seem to work as well). Supporting scalars as well would greatly improve the value of your contribution.
3. **Little improvements to `cond`.**
   - The `else_fn` could be made optional.
   - I think the condition does not need to be a callable, any DaphneLib DAG node should be fine. This could slightly simplify using if-then-else constructs.

lachezar-n:
Hi @pdamme! Thank you for your detailed feedback!

First, about the questions:
1. Adding spaces around the operators like ` + ` was indeed a temporary fix due to the bug with using `-` undated of ` - `. It was just more convenient for me at this moment to change all of the operators to have consistent syntax.  However, I did not come to the point of compiling the project again after the bug was fixed (since this takes several hours on my machine) and because of this, I did not change back the operators before the PR.
2. The function `test_invalid_callback_outputs()` tests a case where the user defines a body callback for the for-loop with a wrong number of outputs. The test implements an example where the defined function has 2 outputs, like the number of inputs. Since the last input is reserved for the iterator, the required number of outputs is 1 (or in general 1 less than the number of inputs). However, an error is raised first when the function computes is called, since in Python there is no way of inferring the number to outputs of a function before its execution. And since the compute function can be only called on a node object, it is being called here on the first node `output[0]`.

Regarding the required changes:
1. I will try to add the mentioned comments asap.
2. I did try adding tests to the existing test framework, but even copying existing tests and just renaming it was failing for me and since it was not much rest time left before my exam I decided to go for this workaround. I think that with some more time, I will figure out the problem I encountered on the first try, and I will bring test cases to the existing framework. However, this would not be immediately but in a few weeks, once I am done with my exams. 

Regarding the other changes:
1. My idea with the current implementation of the user-defined functions was that creating functions in DaphneDSL would add overhead or at least would not bring any performance since these functions would be created every time in the temporary DAPHNE file and then executed immediately. Now, that it is clear what was the target for this functionality I can implement it easily, I think!
2. I followed pretty much the concept of TensorFlow, but I can and will make the appropriate fixes so manipulating scalars is fully supported as well.
3. Yup, this is a good idea and I am currently not sure why I have not implemented it yet. I will do it!
I will aim to fix them together with fixing the tests!

pdamme:
Hi @lachezar-n, thanks for the response!

*Regarding the questions:*

1. Ok, I see. The bug in the DaphneDSL parser regarding spaces around the `-`-operator hasn't been fixed yet, so spaces are still required in the DaphneDSL code generated by DaphneLib. However, rather than hardcoding the spaces for all operators in several places, I suggest either doing that only for `-` (since it should be the only affected operator) or adding the insertion of spaces in a central place where it automatically applies to all operators.
    
    By the way: Normally, only the first build of DAPHNE takes a bit longer, building DAPHNE anew after incremental changes should only take a few seconds to minutes.
3. Thanks, now I got it. Sounds plausible.

*Regarding the requires/other changes:*

Happy to hear that you would be willing to help out with a few of these points. Of course, please focus on your exams first. I personally will probably not find the time to continue with this pull request before around August 16 anyway. I will convert one of your test cases to the form we currently use in DAPHNE to provide an example. Again, thanks for any assistance with the remaining points. In case your time plans change, feel free to let me know and I will do the rest myself (don't feel obliged).

lachezar-n:
Hey @pdamme, I will fix the mentioned points in the next days. However, I am not too sure about the user-defined functions logic and I want to ask you first about that:
My current idea is that DaphneLib would generate these user-defined function and similarly to the lazy evaluation, these definitions would be added to the temporary daphne script in case these functions are actually called. In the simplest example this temporary script would look like:
‘def fib(n) -> {
    if (n <= 0)
        return 0;
    if (n <= 1)
        return 1;
    return fib(n - 1) + fib(n - 2);
}
fib(b);
‘
Is that what you also have in mind?
And is it okay that functions would allow returning of an single value since returning multiple values is still not (at least on my personal tests) supported by DAPHNE DSL?

pdamme:
Hi @lachezar-n, thanks a lot for the additional contributions!

Regarding UDFs: Yes, exactly, the DaphneDSL code for the function definitions should be generated before the actual main program. If it is simple to do that only for the functions that are actually called, great. If it is not simple, then it would also be okay to generate the DaphneDSL code for all functions that were created by the DaphneLib user. The DAPHNE compiler will take care of eliminating unused functions in DaphneDSL anyway (not 100% sure right now, if that is already supported, but if not, it will be at some point).

DaphneDSL UDFs with multiple return values have been implemented in a branch (0085e39853e5c2d6e8ccda9b0d236beafa08ed4e), but not merged yet. I will try to merge that commit by Friday. Otherwise, it would be okay for now to support only UDFs with a single return value.

pdamme:
JFYI: DaphneDSL supports UDFs with multiple return values now, on main (see 6db6d80d3d7a7ecf92e49fa1a3e40dc62066055e).

lachezar-n:
Hey @pdamme, I did the base implementation for the UDF, which works in the base cases. 
However, while testing (currently still with `pytest`) I encountered some unexpected behaviour of DaphneDSL regarding functions:
The following example:
```
def increase_if(L0) {
    ret = L0;
    if (sum(L0) > 100) {
        ret = ret + 1;
    }
    return ret;
}
```
leads to: `[error]: Exception in /home/ubuntu/daphne/src/compiler/inference/InferencePass.cpp:165: 
type inference returned an unknown result type for some op, but partial inference is not allowed at this point: scf.if`

The same issue occurred when I tried with different loop statements. The issue is fixed by giving `matrix<f64>` type to the arguments and the return value. I first thought of implementing function definitions that require giving types, but then I saw that another error occurred when using `matrix<f32>`. 
Other curious (for me) examples with  the same error are:
```
def foo(L0: f64) -> f64 {
    return L0;
}
ret = foo(3); // not 3.0
```
and 
```
def foo(L0: f32) -> f32 {
    return L0;
}
ret = foo(3.0);
```
These lead to: `[error]: Parser error: No function definition of `foo` found with matching types`.

At this point, I am not really sure how to approach this problem so the solution covers all the cases. Do you have any suggestions for that? And maybe first another question: Are these behaviours expected and desired overall?

pdamme:
Hi @lachezar-n, thanks for keeping up the good work, that’s highly appreciated! I read the current state of your code today. All the points mentioned above have been addressed very well. Regarding the UDFs, I particularly like that we can indeed use ordinary Python functions (definition+decorator and calls).

**Regarding the problems you reported related to UDFs in DaphneDSL:** I can reproduce those issues. They are all related to current limitations and bugs of DaphneDSL and the DAPHNE compiler.
- The examples in your comment from August 24 shall all ultimately work in DaphneDSL. The first example is a problem related to property inference in untyped functions. The second and third example currently suffer from a missing automatic cast of the arguments to typed UDFs (one could explicitly cast the arguments, e.g., by `as.f64(3)`), but that would be just a workaround.
- On your comment from September 7: Changing the value of a frame in a loop shall work in DaphneDSL. The current limitation in your example is some missing inference logic of the `rowBind`-operation and/or type comparison in the inference logic, as far as I can tell now.

**Fixing those problems is possible, but it’s beyond the scope of this PR; I will take care of them.** For your PR, I recommend focussing on the generated DaphneDSL script, rather than on the execution/output, for those problematic cases. Indeed, the generated scripts do already look quite good and as expected. **I only noticed two cases that produce unexpected DaphneDSL code (see below). If you still have some time and interest, you could try to fix those.** If not, then I will address them. In any case, you have already made an awesome contribution to DAPHNE, thanks again!

**Besides that, I will take care of the remaining minor polishing of the code and merge it.**

------

**The two cases of unexpected generated DaphneDSL scripts:**

1. `test/api/python/user_def_func_with_condition.py`
    - The expression `fill(123.0,5,5)` should be outside the UDF.
    - In fact, `F1_V0` is used but not defined in the main scope, which makes the DaphneDSL script crash irrespective of potential problems with DaphneDSL UDFs.
    - Generated DaphneDSL script:
        ```
        def function_0(ARG_0) {
            F1_V0=fill(123.0,5,5);
            F1_V1=sum(F1_V0);
            F1_V2=F1_V1<10.0;
            if (F1_V2) {
                L2_V0=ARG_0 - 1;
                ARG_0=L2_V0;
            } else {
                L2_V0=ARG_0+1;
                ARG_0=L2_V0;
            }
            F1_V4=ARG_0;
            return F1_V4;
        }
        V0_0=function_0(F1_V0);
        print(V0_0);
        ```

2. `test/api/python/user_def_func_with_while_loop.py`
    - The condition `C2_V0>0.0` is calculated once before the while loop.
    - Thus, all iterations see the same value.
    - I would have expected something like `while(C2_V0>0.0)`.
    - The generated DaphneDSL script:
        ```
        def function_0(ARG_0) {
            C2_V0=sum(ARG_0);
            C2_V1=C2_V0>0.0;
            while (C2_V1) {
                L1_V0=ARG_0 - 1;
                ARG_0=L1_V0;
                C2_V0=sum(ARG_0);
                C2_V1=C2_V0>0.0;
            }
            F1_V1=ARG_0;
            return F1_V1;
        }
        V0=fill(3.8,5,5);
        V1_0=function_0(V0);
        print(V1_0);
        ```

pdamme (approved):

pdamme:
Hi @lachezar-n,

1. Thanks for the fix!
2. Oh yes, you are right. I didn't notice that the condition is calculated again at the end of the loop's body. It is fine that way, and indeed a good option for supporting arbitrarily complex conditions.

So in terms of features, this PR is ready, from my point of view. **Thanks again for the excellent work!** *This contribution is far beyond what is normally expected for an AMLS project.*

I will take care of the remaining minor things/clean-ups and merge the code in (hopefully next week).

PR #578, opened 2023-07-30
Pandas Shared Memory Support

danielwetzel:
Added Pandas Shared Memory Support as well as a Test Script to check the Performance.

Test Script: "scripts/examples/daphnelib/data-exchange-pandas-performancetest.py"

Adjusted Files:
- src/api/daphnelib/DaphneLibResult.h
- src/api/python/context/daphne_context.py
- src/api/python/operator/nodes/frame.py
- src/api/python/operator/operation_node.py
- src/api/python/script_building/script.py
- src/api/python/utils/daphnelib.py
- src/runtime/local/kernels/SaveDaphneLibResult.h
- src/runtime/local/kernels/kernels.json


danielwetzel:
Functionality for TensorFlow also included, but not tested yet, since we cannot install TensorFlow within our Daphne Container.

corepointer:
> Functionality for TensorFlow also included, but not tested yet, since we cannot install TensorFlow within our Daphne Container.

What's the problem with installing TF in your DAPHNE container?

PR #586, opened 2023-08-08
Only deploy docs when build from main branch

m-birke:
docs are still built (but not deployed) for any PR to main branch


m-birke:
why closed?

m-birke:
ah okay thanks :) would have been okay, next time again from feature branch, its more convenient

PR #592, opened 2023-08-21
[dlib] relative python imports

m-birke:


m-birke:
@pdamme please review, this exploits that you already have a package structure, pwd of python cmd is now arbitrary

corepointer:
I'm not really a python pro but from what I gathered over the years it's rather error prone and to be avoided if possible to use relative imports. If this is all about setting the PYTHONPATH environment variable, then this would be the lesser evil imho. 

I had a quick look at the daphnelib code and there's a lot of potential to improve things (atm ``--libdir lib`` is hard coded and no other config options reach the DAPHNE system.

These are just my two cents as the python parts of DAPHNE aren't so much my turf :see_no_evil:

m-birke:
Please notify me whether I should solve the conflicts or withdraw the PR

corepointer:
> Please notify me whether I should solve the conflicts or withdraw the PR

Since I'm not coding the python parts I'd like to have an opinion from @pdamme. I can handle the conflicts while merging.

m-birke:
PR is obsolete

PR #596, opened 2023-08-30
Cc cuda ops

corepointer:
This pull request contains CUDA ops needed for the connected components algorithm (only dense versions for now).


corepointer:
Merging this now with a promise to deliver test cases :innocent:

PR #604, opened 2023-09-07
Misc improvements

corepointer:
Silencing compiler warnings and introducing two new loggers :)

### DIFF
--- a/src/compiler/lowering/SpecializeGenericFunctionsPass.cpp
+++ b/src/compiler/lowering/SpecializeGenericFunctionsPass.cpp
@@ -322,7 +322,7 @@ namespace {
                 bool hasConstantInput = llvm::any_of(
                         callOp.getOperands(),
                         [&](Value v) {
-                            return CompilerUtils::constantOfAnyType != nullptr;
+                            return CompilerUtils::constantOfAnyType(v) != nullptr;
                         }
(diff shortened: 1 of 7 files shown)


corepointer:
The github action seems to take forever here for unknown reasons. Local test runs (on x86-64) say these changes are alright, so I'll merge it off the table.

src/compiler/lowering/SpecializeGenericFunctionsPass.cpp
MarcusParadies, line 1: Not sure why this had to be done (now and on the "fast lane"), but thanks for the merge conflict in the other pull request. ;-)

corepointer:
Sorry to have caused you trouble there. Yes, these conflicts are a bit annoying but they come up every now and then for one reason or another.

I did not see that conflict coming right away. I put that log message there when testing your changes. I had to pull that out into a separate commit because I based other stuff on it indirectly (so much for the "now". Could have gone another way if I had foreseen the trouble). The conflict itself was easily resolved (accept left and right change in no particular order - I had to deal with the change again when merging to main).

If by "fast lane" you mean merging minor changes without review: Yes, we're notoriously understaffed when it comes to PR processing so we try to keep the noise level low by directly merging minor things.

PR #605, opened 2023-09-07
[DAPHNE-#595] Fix crash when calling print(f()) with empty function

### LINKED ISSUE #595: Crash when calling print on the return of an empty UDF
Daphne (v0.2 && main branch) crashes when invoking print() on an empty UDF (not return types, no statements in body). 

DSL script:

```
def f() {}
print(f());
```

Stacktrace:

```
(gdb) bt  
#0  0x000055ef013749bd in mlir::Type::isSignedInteger(unsigned int) const ()
#1  0x000055ef007abcb1 in mlir::daphne::__mlir_ods_local_type_constraint_DaphneOps37 (op=0x55ef0626b190, type=..., valueKind=..., valueIndex=0)
    at /home/marcus/code/daphne_fork/build/src/ir/daphneir/DaphneOps.cpp.inc:605
#2  0x000055ef0083e5cb in mlir::daphne::PrintOp::verifyInvariantsImpl (this=0x7f3c3c0cb550) at /home/marcus/code/daphne_fork/build/src/ir/daphneir/DaphneOps.cpp.inc:22591
#3  0x000055ef00a5704b in mlir::OpTrait::OpInvariants<mlir::daphne::PrintOp>::verifyTrait (op=0x55ef0626b190)
    at /home/marcus/code/daphne_fork/thirdparty/installed/include/mlir/IR/OpDefinition.h:371
#4  0x000055ef00a2b52e in mlir::op_definition_impl::verifyTrait<mlir::OpTrait::OpInvariants<mlir::daphne::PrintOp> > (op=0x55ef0626b190)
    at /home/marcus/code/daphne_fork/thirdparty/installed/include/mlir/IR/OpDefinition.h:1565
#5  0x000055ef009f18c0 in mlir::op_definition_impl::verifyTraits<mlir::OpTrait::ZeroRegions<mlir::daphne::PrintOp>, mlir::OpTrait::ZeroResults<mlir::daphne::PrintOp>, mlir::OpTrait::ZeroSuccessors<mlir::daphne::PrintOp>, mlir::OpTrait::NOperands<3u>::Impl<mlir::daphne::PrintOp>, mlir::OpTrait::OpInvariants<mlir::daphne::PrintOp> > (op=0x55ef0626b190)
    at /home/marcus/code/daphne_fork/thirdparty/installed/include/mlir/IR/OpDefinition.h:1576
#6  0x000055ef009cd957 in mlir::Op<mlir::daphne::PrintOp, mlir::OpTrait::ZeroRegions, mlir::OpTrait::ZeroResults, mlir::OpTrait::ZeroSuccessors, mlir::OpTrait::NOperands<3u>::Impl, mlir::OpTrait::OpInvariants>::verifyInvariants (op=0x55ef0626b190) at /home/marcus/code/daphne_fork/thirdparty/installed/include/mlir/IR/OpDefinition.h:1861
#7  0x000055ef00a1ff29 in llvm::detail::UniqueFunctionBase<mlir::LogicalResult, mlir::Operation*>::CallImpl<mlir::LogicalResult (* const)(mlir::Operation*)> (CallableAddr=0x7f3c3c0cb660,
    Params#0=0x55ef0626b190) at /home/marcus/code/daphne_fork/thirdparty/installed/include/llvm/ADT/FunctionExtras.h:221
#8  0x000055ef009b3293 in llvm::unique_function<mlir::LogicalResult (mlir::Operation*) const>::operator()(mlir::Operation*) const (this=0x7f3c3c0cb660, Params#0=0x55ef0626b190)
    at /home/marcus/code/daphne_fork/thirdparty/installed/include/llvm/ADT/FunctionExtras.h:409
#9  0x000055ef0098e4b5 in mlir::RegisteredOperationName::Model<mlir::daphne::PrintOp>::verifyInvariants (this=0x55ef06221650, op=0x55ef0626b190)
    at /home/marcus/code/daphne_fork/thirdparty/installed/include/mlir/IR/OperationSupport.h:411
#10 0x000055ef013770d9 in (anonymous namespace)::OperationVerifier::verifyOperation(mlir::Operation&) ()
#11 0x000055ef013773cb in (anonymous namespace)::OperationVerifier::verifyOperation(mlir::Operation&) ()
#12 0x000055ef01378ef6 in std::_Function_handler<void (), mlir::failableParallelForEach<mlir::Operation**, (anonymous namespace)::OperationVerifier::verifyOperation(mlir::Operation&)::$_0>(mlir::MLIRContext*, mlir::Operation**, mlir::Operation**, (anonymous namespace)::OperationVerifier::verifyOperation(mlir::Operation&)::$_0&&)::{lambda()#1}>::_M_invoke(std::_Any_data const&)
    ()
#13 0x000055ef01056bd2 in std::_Function_handler<void (), llvm::ThreadPool::createTaskAndFuture(std::function<void ()>)::{lambda()#1}>::_M_invoke(std::_Any_data const&) ()
#14 0x000055ef035f2274 in llvm::ThreadPool::processTasks(llvm::ThreadPoolTaskGroup*) ()
#15 0x000055ef035f2ef3 in void* llvm::thread::ThreadProxy<std::tuple<llvm::ThreadPool::grow(int)::$_0> >(void*) ()
#16 0x00007f3c3d494b43 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:442
#17 0x00007f3c3d526a00 in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:81
```
I'm assuming that since no return type is known/can be deduced there is a segfault when checking for the type. Imho the daphne compiler should not accept such a script since no return type can be deduced from the function. Options to handle this could be either to enforce the user to have at least one return statement within a UDF or fail the compilation of no return type can be deduced from the UDF.

corepointer:
This commit contains two changes more or less related to this issue:

1. There is a check preventing empty function definitions. While I thought the empty function had something to do with the crash at first, this turned out to be not the case. Nevertheless, I kept the change as reminding the user that there's unnecessary code would improve our compiler a tiny little bit (imho).

2. The core issue seems to be that functions without return value internally returned an invalid mlir::Value. Upon doing anything with this value, the program would crash (e.g. accessing any fields or calling any methods). Now we return nullptr and check for that, which gives a somewhat more deterministic behavior.

Closes #595


corepointer:
We actually do have a test case that has an empty function. I can also remove the empty function error if we want that functionality (if we do, an explanation why would be ideal)

MarcusParadies:
To me the fundamental problem is the combination of allowing empty functions (to be more precise: no return statement in a UDF) and at the same time making the return type declaration optional. 

E.g., in C++ you can do

```
auto a() { 
    return 0; 
} 
int main(void) {
    std::cout << a() << std::endl;
}
```

or 

```
int a() { } 
int main(void) {
    std::cout << a() << std::endl;
}
```

but you cannot do

```
auto a() { } 
int main(void) {
    std::cout << a() << std::endl;
}
```

Imho one of the two should be enforced by the compiler. Either specify a return type (and simply return the default value if no return statement is present) or force the user to specify at least one return statement. Personally, I find enforcing the user to specify the return type explicitly the cleanest.

Otherwise you cannot deduce the return type of the function at runtime. E.g. the following is also not supported in C++:

```
auto a() {
    if (true) {
        return 0;
    } else {
        return "0";
    }
}
int main(void) {
    std::cout << a() << std::endl;
}
```

corepointer:
I'm undecided on my opinion in this matter. On the one hand yes, stating return types explicitly would clean but also more code to type. Determining the return type in the compiler would also be nice as a modern compiler should be able to handle that. But then again this is something that I'd like to be cheaply implemented (or "for free" by some MLIR magic).
As for your last example - what about a check that all return statements at least return the same type?

corepointer:
I removed the empty function check as empty functions are used somewhere in the tests. The situation in the bug report does not crash DAPHNE anymore.  So for the empty function
``def empty() {}`` a print ``print(emtpy_func())`` yields ``[error]: While parsing: DSLBuiltins failed with the following message [ built-in function `print` expects between 1 and 3 argument(s), but got 0 ]
``
and with ``print(empty(),0);`` you'll get ``0`` as output.
If you (@pdamme) don't have any further comments/concerns about this I'd say we merge this and continue the discussion of what language features to support that started above some other time. Maybe in a dedicated gh issue.

corepointer:
PS: rebased on main and tested in the container environment :heavy_check_mark:

pdamme (approved):
Thanks for this fix, @corepointer!

Yes, the problem occurred when a call to a UDF with zero return values was used in a context where exactly one value is expected. So far, we couldn't distinguish the case of zero return values from one return value. Thus, it's good that we now explicitly return a `nullptr` in `visitCallExpr()`. (There was actually a comment in the code indicating that there could be a problem; I removed this comment now.)

However, the check for `isNotNull()` on each function call argument had some downsides: 
1. We cannot simply ignore erroneous arguments. For instance, when function `foo()` has no return values, then `print(foo(), 0);` should not print `0`, but  fail.
2. Call arguments are just one out of many places where a UDF with zero results could be used (erroneously) in DaphneDSL. All these places should be handled consistently. In fact, we have `valueOrError()` for that purpose and it already covers this case. So I undid this check and we rely on `valueOrError()` now.

The error message is not informative at the moment, but that will be improved in a follow-up commit.

I also added several script-level test cases, which check that DAPHNE does not accept UDFs with zero or more than one return value in places where exactly one value is expected. All of these tests on a UDF with zero return values crash DAPHNE with a segfault on main (so problem in #595 is not specific to call expressions). Actually, these are just a few examples, ideally we should have test cases for all uses of `expr` in the DaphneDSL grammar.

pdamme:
**JFYI:** As a follow-up I've just pushed commit 04094fff59be0b66ccb6bd81d77369097ba52777, which ensures informative error messages in such cases.

*Before we had:*
```
[error]: While parsing: something was expected to be an mlir::Value, but it was none
```

*Now we have:*
```
[error]: While parsing: DSLVisitor failed with the following message [ the expression was expected to return a single value, but it returns no value ]
    | Source file -> "/daphne/test/api/cli/functions/invalid_parser_13.daphne":7:6
    |
  7 | print(f(123));
    |       ^~~
```

Issue #560, opened 2023-06-29
Parsing minus-operator in DaphneDSL


pdamme:
The following trivial DaphneDSL script:
```
print(1-1);
```
yields the following error message:
```
Parser error: my-example.daphne:1:8 extraneous input '-1' expecting {',', ')'}
```

A quick fix is to insert spaces around the minus operator as follows:
```
print(1 - 1); # this way
print(1- 1);  # or this way
```

The workaround requires a space between the minus-operator and the right-hand-side literal. This indicates that the problem is related to parsing DaphneDSL expressions (literal expression vs. binary operation expression). Note that a literal can start with a minus, which could confuse the parser here.

**The spaces should not be necessary**, i.e., `print(1-1);` is valid.

Thanks to @lachezar-n for reporting this problem in a comment on #522.

pdamme:
This issue is related to (or a duplicate of) #186.

Issue #582, opened 2023-08-04
passing str variable to readFrame/readMatrix does not work


bl1zzardx:
I am using daphne-v0.2. Running this script below works for me, if I run the commented line for reading in the df. 
If I run it by using the passed `path` argument, it results in an Parser error.

```
def read_csv(path: str) {
    print(path);
    df = readMatrix(path);
    # df = readMatrix("/home/laberand/ifat_semiconductors/data/x.csv");
    print("Read csv at ", 0); print(path, 0);
}

read_csv("/home/laberand/ifat_semiconductors/data/x.csv");
```

Error:

> [error]: Parser error: basic_string::_M_construct null not valid

bl1zzardx:
I forgot to mention that you can actually `print(path);` normally. Even casting it explicitly to a string via `print(as.str(path));` does not fix the error.

corepointer:
Thx for reporting this. 
Quick pointer if anybody wants to solve this: We try to read file meta data while still parsing. So it looks like the parameter to the user defined function read_csv() is still a symbol and not really usable as a string. So this needs a not so trivial redesign of parsing readMatrix/readFrame.

m-birke:
+1

m-birke:
@bl1zzardx could you rename this issue? It is generally about the readX() functions and passing a str variable to them as argument. Also appears when using it directly within a script (so not a UDF)

m-birke:
Lemme mention that this is a really blocking issue, to scale out with a DSL application one somehow at least needs a possibility to iterate over a dataset of files which necessarily are represented as variables

bl1zzardx:
> @bl1zzardx could you rename this issue? It is generally about the readX() functions and passing a str variable to them as argument. Also appears when using it directly within a script (so not a UDF)

done.

PR #607, opened 2023-09-07
[DAPHNE-#560] Parsing unary minus operator

corepointer:
This change decouples the '-' from INT_LITERAL and FLOAT_LITERAL and handles a unary minus separately.

Testcases... I know... :roll_eyes:


corepointer:
I could fix the script argument test by some more DSL grammar changes. But the matrix literals remain an issue. Similar to #582, this functionality relies on constants to generate data, which is something we can not provide in the parsing stage (current fix would need a round of constant folding to do the multiply with -1).
Just mentioning this to imply that completing this PR might take a bit longer.

pdamme (approved):
Thanks for this fix, @corepointer! It basically looked good, but required some additional work to make it more robust.

For instance, some fixes in the grammar, support for unary minus before matrices (not only scalars), implementation of the unary minus via `EwMinusOp` instead of multiplication by -1 to avoid some type problems, restriction of script arguments to literals optionally preceded by a minus (as opposed to arbitrary expressions, as those could have unwanted side effects), more test cases, updates to the DaphneDSL docs. For more details, see the description of the commit I added.

As that was quite some work compared to the initial state of the PR, I take the liberty to add myself as a co-author of the resulting commit on main. This PR is ready to be merged now.

pdamme:
The failed CI run is most likely due to #729, so I will merge it anyway.

PR #623, opened 2023-10-03
Distributed worker: Insert IR arguments to body.

### LINKED ISSUE #557: MatMul operation fails on the distributed runtime
Matmul crashes on the distributed runtime. There seems to be an issue with the inputs, when parsing the IR code fragment on the worker.

Example input script:
```
a = rand(5, 5, 1.0, 10.0, 1.0, 1);
b = rand(5, 5, 1.0, 10.0, 1.0, 1);
print(a@b);
```
Distributed worker output:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  MatMulOp.getTransb() is expected to be a constant
Aborted (core dumped)
```

aristotelis96:
Distributed workers receive an IR code that needs to be computed. It is received as an IR function, that the worker currently execute by passing additional arguments (the arguments of a pipeline). 

However this can cause issues since some optimization passes require some of these arguments to be constant. Since these arguments are constant and stored in the worker's memory we simply need to modify the IR. With this commit, the workers modify the IR code by removing all arguments and inserting constant operations into the IR code fragment before executing it.

- Modified how the workers parse the IR code fragment before execution.
- Added CSRMatrix specialization for `MatrixConstant.h` kernel.
- Added a missing catch-error for the compute step using grpc.

Closes #557


pdamme (changes requested):
Thanks for this contribution, @aristotelis96. I haven't tried it, but reading the code, I think it would solve the problem described in #557. Besides that, turning the arguments of the `dist`-function into constants inside the function could even allow the DAPHNE compiler to perform some additional optimizations of the function body by exploiting the concrete values, in some cases.

Nevertheless, there are two things I need to point out:

**Required before merging**
- Please add a script-level test case for the distributed execution of a matrix multiplication, such that we can automatically check if #557 is/stays solved.

**Strongly recommended before merging**
- While it looks good in terms of functionality, I think the code has a performance bug. To solve #557, it would suffice to turn *scalar-type* function arguments to `ConstantOp`s inside the function. In this PR, you also turn *matrix-type* function arguments to `MatrixConstantOp`s inside the function. The problem with the latter is that the `MatrixConstant`-kernel (including the specialization you added for `CSRMatrix`) always copies the matrix object it is backed by **(*)**.  For the use in a distributed setup, this means that the worker will always perform an additional copy of each matrix it received, which may impact execution time and memory consumption. To avoid these problems, I recommend either (a) turning only scalar-type inputs of the `dist`-function to constants and keep passing matrix-type inputs as arguments, or (b) changing the way matrix constants work such that they do not require copies anymore. I think (a) is much simpler and sufficient to fix #557, while (b) would require some effort and testing.

-----

**(*)** The reason why the `MatrixConstant`-kernel copies the matrix it is backed by is to prevent the original matrix from being freed by DAPHNE's garbage collection, especially in presence of loops and repeated function calls. There may be other ways to prevent that, e.g., by a smart use of DAPHNE's reference counters. So far, we implicitly assumed that matrix constants stem from matrix literals in DaphneDSL, which would normally be used for small matrices (not gigabytes), but that assumption may not hold anymore if we use matrix constant for the inputs of the `dist`-function.

aristotelis96:
Thank you @pdamme for the review! I've been working on this and decided to follow a different approach. Instead of removing arguments and adding constants inside the IR on the worker level, we could also do that while generating the `DistributedPipeline` operation on the coordinator. This basically creates the same IR that the worker would create on it's own with two added benefits:
1. We don't use the Matrix constants. Matrices still remain arguments and therefor we do not call the `MatrixConstant`-kernel, and
2. We don't need to explicitly use messages (gRPC, MPI, etc.) to send these arguments to the worker, since all the information needed is already contained in the IR.

I've also added a script-level test case, rebased the branch and reverted the last two commits. 

Let me know what you think. It seems to work fine, but any tips related to `DistributedPipelinePass.cpp` would be appreciated.

pdamme (approved):
Thanks for the updates, @aristotelis96. The code looks good to me now. Moving the constants into the distributed pipelines in order to avoid the need to send them is a nice additional benefit.

**Feel free to *squash-and-merge* this PR.**

PR #626, opened 2023-10-15
[MINOR] Enable shape inference for frame type

DamianDinoiu:
- DAPHNE support shape inference for frame type but the feature was disabled.
   - The logic should be further revised as it is redundant to throw an error.

### DIFF
--- a/src/compiler/inference/InferencePass.cpp
+++ b/src/compiler/inference/InferencePass.cpp
@@ -188,7 +188,8 @@ class InferencePass : public PassWrapper<InferencePass, OperationPass<func::Func
                         );
                     // Set the infered shapes on all results of this operation.
                     for(size_t i = 0 ; i < numRes ; i++) {
-                        if(op->getResultTypes()[i].isa<mlir::daphne::MatrixType>()) {
+                        if(op->getResultTypes()[i].isa<mlir::daphne::MatrixType>() ||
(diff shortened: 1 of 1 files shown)


pdamme (approved):
Hi @DamianDinoiu, thanks for identifying this bug and providing a fix! This pull request reveals that shapes inferred for frames were not actually set at all, such that the shapes stayed unknown in the IR. I can confirm that your proposed change solves this problem.

On top of your changes, I merely formatted the code a bit for better readability. I also added the same fix for sparsity inference; even though we do not infer sparsity for frames yet, the same problem could pop up there in the future (and the code also checked for `FrameType`).

Part of the reason why we didn't discover this bug before may be that we hardly actually used the inferred shapes of frames so far. Therefore, it's currently not straightforward to add a test case (but that will become easier in the future). For now, one can check the behavior by executing this simple DaphneDSL script with `--explain property_inference`:
```
print(createFrame([1]));
```
Before your fix, the IR after property inference still indicated an unknown shape for the frame:
```
%5 = "daphne.createFrame"(%4, %2) : (!daphne.Matrix<1x1xsi64>, !daphne.String) -> !daphne.Frame<?x[?: si64], ["col_0"]>
```
With your changes, the shape is known in the IR:
```
%5 = "daphne.createFrame"(%4, %2) : (!daphne.Matrix<1x1xsi64>, !daphne.String) -> !daphne.Frame<1x[1: si64], ["col_0"]>
```

PR #633, opened 2023-10-24
MLIR-based code generation pipeline 

philipportner:
This PR aims to bring MLIR-based code generation for certain operations to DAPHNE. 

As an alternative way to implement DAPHNE operators this PR provides a code generation
pipeline which progressively lowers the DaphneIR available after parsing the
DaphneDSL script to operations in either the same dialect or operations from
other dialects. With that, DAPHNE can optionally replace certain kernels by
generating code directly, and also perform a hybrid compilation approach where
we mix kernel calls with code generation in order to exploit advantages of
both, precompiled kernel libraries and code generation. 

This includes the following features:
- Documentation ([Codegen.md](`doc/Codegen.md`))
- Code generation passes `src/compiler/lowering/`
- Kernels to facilitate interoperability between the `memref`-dialect and the `DenseMatrix` runtime object
- CLI arguments to enable code generation (`--codegen` and `--codegen --hybrid`) and explain-passes (`--explain code_gen`)
- End to end tests that run `daphne` with a DaphneDSL script as input in `test/compiler/e2e/`
- `llvm-lit` based testing of the IR produced from passes / pass pipelines in `test/compiler/lit/` using intermediate IR as input (`.mlir` files).
- The `daphne-opt` tool, DAPHNEs own version of the `mlir-opt` tool with its source files in `daphne-opt/` and a new build target `./build.sh --target daphne-opt`. 

`daphne-opt` is included in this PR since the `lit` tests require the `daphne-opt` target.
With `daphne-opt` one can test passes in isolation by providing the
input IR for a pass to `daphne-opt` and the correct flag to run the pass (or
a pass pipeline) on the IR. The `daphne-opt` tool inherits all the functionality of the
`mlir-opt` tool.

`daphne-opt --lower-ew --debug-only=dialect-conversion ew.mlir` performs the
`LowerEwOpPass` on the input file `ew.mlir` while providing dialect conversion
debug information.

The list of affected operations (only when `daphne` is executed with the `--codegen` flag):

- MatMulOp
- MapOp
- Ew(Sqrt | Abs | Add | Sub | Mul | Div | Pow | Mod)Op
- AllAggSumOp
- ConvertDenseMatrixToMemRef
- ConvertMemRefToDenseMatrix

`daphne --codegen --hybrid` can be used to disable code generation for certain operators in order to run a hybrid compilation pipeline (currently only the `MatMulLoweringPass` is affected by this.)


corepointer:
Hi @philipportner,
I've been following this PR a bit, as the topic is included in my line of work ;-) But I haven't tried it out yet :angel: 
I saw the note about the container update - will do.
We need to fix the name clash as I named my codegen also codegen :see_no_evil: What about "MLIR codegen" and "CUDA codegen" ? Altough the code I produce could be used as normal C++ kernels as well with some modification :-P But as a working title it'll be sufficient. 
Best regards, Mark

philipportner:
Hi @corepointer ,
thanks for looking into the container update!

Yes, we should fix that. I'll change mine to `--mlir-codegen`.

BR,
Philipp

philipportner:
@corepointer thanks for adding `lit` support!

I see you removed the symlinks I added to the `main.yml` file, I suppose this was the wrong way to add these and there is another way this should be done? 

```
- name: "Create symlinks"
      run: |
        ln -s bin/daphne /usr/bin/daphne
        ln -s bin/daphne-opt /usr/bin/daphne-opt
        ln -s thirdpart/build/llvm-project/bin/FileCheck /usr/bin/FileCheck
```

Preferably, the `*.mlir` files used by `lit` can contain the `RUN:` directive without hardcoding the path to the binaries, e.g., `// RUN: daphne-opt --lower-ew %s | FileCheck %s`. Making them absolute path that work on the CI system probably means one has to change something to make them work locally. Relative paths add the burden of making sure that they are always correct during local development and work on other systems. Simply being able to write the executable name in the `RUN:` directive would be the best solution for developers.

corepointer:
Hi!
It all worked just fine locally by setting the PATH variable correctly - no symlinking needed. I tried to port that change from test.sh to main.yml because we're invoking ```run_tests``` manually. I have to look at it again why it failed in the github actions.

philipportner:
> Hi! It all worked just fine locally by setting the PATH variable correctly - no symlinking needed. I tried to port that change from test.sh to main.yml because we're invoking `run_tests` manually. I have to look at it again why it failed in the github actions.

Hey @corepointer , 
let me have a look. I think this is actually something that should be done in the lit-config. I didn't realize that you can do that there. We have all the executables required in `bin/` and `thirdparty/build/` so that should work without any changes to the `main.yml`.

corepointer:
The executables from LLVM should already provided by the llvm-10* Ubuntu packages - that's how I tested it with the docker image locally and that worked. I added some debug output to see what runLIT() is actually doing.

philipportner:
> The executables from LLVM should already provided by the llvm-10* Ubuntu packages - that's how I tested it with the docker image locally and that worked. I added some debug output to see what runLIT() is actually doing.

Okay, that may work for `FileCheck` but the `lit`-based tests also need to run the new `daphne-opt` compiler driver. I added the paths to the `lit.cfg` and that should now work even if the `llvm-*` tools are not part of `$PATH`, though, seems like the `lit` python package is still missing. I don't think `lit` is part of the `llvm-10*` installation (see https://discourse.llvm.org/t/llvm-lit-not-built/3546/5?u=philipportner). As I anyway run this `run-lit.py` script through `python3` in the test I think the `lit` [PyPi](https://pypi.org/project/lit/) package needs to be installed.

corepointer:
The llvm-10-tools package was missing from the github-actions docker image. I thought that was included in the llvm meta package but it was not. The lit tests work now. The paths in lit.cfg are not needed (as it worked before without that). The lit stuff is pointed to by PYTHONPATH and the daphne-opt binary is available via $PWD/bin in PATH. 

You might want to squash these 151 commits into a few topic commits (or rather sub-topics). Otherwise it will most likely end up in one huge "MLIR codegen" commit during the merge to master - not strictly forbidden but maybe it helps in error tracking some day.

pdamme (changes requested):
**Thanks for this contribution**, @philipportner. Adding the generation of low-level MLIR-code for (certain) DaphneIR operations as an alternative to the current way of lowering to pre-compiled kernels is a great new feature that will surely enable lots of interesting work in the future and has the potential to improve the system in various ways.

I looked over your proposed changes, mainly focusing on the integration into the overall code base (and only briefly looking over the codegen details). **Overall, it looks very good to me, very good work!**. Nevertheless, there are a few points that need to be addressed before we can merge it in.

**Required changes before the merge.**

1. `GettingStarted.md` should mention `lit` as a dependency.
1. `doc/CodeGen.md` mentions the directory `test/compiler/unittest/`, which does not exist. Please adapt to the current directory structure.
1. Rewrite of `a % b` to `a & (b-1)` (`DaphneOpPass.cpp`): `optimization_viable()` checks if `rhs` is a multiple of two, but actually it must be a power of two.
1. `ManageObjRefsPass.cpp`: `processMemRefInterop()`: line 63: `v.getDefiningOp()->getOperand(0).getDefiningOp()->getResult(0));`: I think the last part (`.getDefiningOp()->getResult(0)`) should be removed, since (a) it seems to be redundant, (b) what if the operand of the `ConvertDenseMatrixToMemRef`-op is the second result of its defining op, and (c) what if the operand is a block argument and has no defining op.

**Optional changes.** These are not required before the merge. In order not to increase the size of this PR (whose code was developed over quite some time) even more, it is totally fine to address these at a later point in time. But feel free to address some of them before the merge if you like.

1. I recommend avoiding the word "tensor" in the source code (identifiers etc.), since DAPHNE does not support tensors so far. Please use matrix instead.
1. I recommend adding a short comment ideally short comment above each new compiler pass which briefly describes what the pass does. This is not always obvious from the pass's name, e.g., for `DaphneOptPass.cpp`.
1. I recommend a consistent naming scheme for the new passes, e.g., `LowerXOpPass` vs `XOpLoweringPass`.
1. So far, all canonicalizations for DaphneIR ops are defined in `DaphneDialect.cpp`. You placed the implementations for some new ops in `MemRefOps.cpp`. For consistency, I recommend moving them to `DaphneDialect.cpp`. I agree that `DaphneDialect.cpp` is too large and it would be great to split it, but I recommend doing that consistently for all operations in a separate commit.
1. So far, each kernel header file contains specializations of one operation. The new file `MemRefInterop.h` contains two different ops. I recommend splitting it into two files. These files should have the ops'/kernels' names as their filenames, such that we can avoid the non-self-explanatory word "Interop".
1. I like the test cases in `test/compiler/e2e/`. However, to my mind they should reside somewhere in `test/api/cli/`, with all the other script-level test cases.
1. In the new test cases, you could consider using our utility `compareDaphneToStr()` for more consice code.
1. `REQUIRE` vs `CHECK` in catch2 test cases: I generally recommend using `CHECK` unless the test case cannot safely continue if the condition is violated. With `REQUIRE`, the remaining checks are not executed and don't produce any output. This can make debugging harder. But I admit that this is a case-by-case decision.
1. `EwBinaryScalarTest.cpp`: Pedantically, `abs()` is not an elementwise binary function.
1. Script-level test cases: I recommend starting each DaphneDSL script with a comment line on what is being tested here (what detail to pay attention to). For instance, in `test/compiler/e2e/*.daphne` you frequently extract scalar elements from a matrix that is given by a matrix literal. I assume you do that to make sure that the scalars are unknown at compile-time to avoid constant folding. However, in the future the DAPHNE compiler can be smart enough to know these scalars at compile-time, thereby circumventing the code you want to test here. A comment could help us detect what needs to be kept in mind.

**Feel free to share your thoughts.**

pdamme:
By the way: I also applied a few smaller changes myself. Feel free to have a look.

pdamme:
So I think we are very close to merging this PR in. Again, those optional points can be addressed later after the merge; we should just not forget about them ;) .

philipportner:
Thank you for the review @pdamme! I'll implement the requested changes tomorrow. I think the optional changes also make a lot of sense so I'd rather fix them now too :)

pdamme (approved):
Thanks for the code revisions, @philipportner. All required and optional changes have been addressed. **From my point of view, this PR is ready to be merged.**

@corepointer: Since you were also looking into this PR and helped to fix a few things: Are there any additional comments from your side to take into account? If not, I would squash and merge this PR. (I agree that splitting it into a few smaller commits on individual contributions would be nice-to-have for the history, but I think it's not really worth the effort now.)

containers/daphne.Dockerfile
```diff
@@ -64,7 +64,7 @@ LABEL "org.opencontainers.image.version"="$TIMESTAMP"
 LABEL "org.opencontainers.image.created"="${CREATION_DATE}"
 LABEL "org.opencontainers.image.revision"="${GIT_HASH}"
 RUN apt-get -qq -y update && apt-get -y upgrade && apt-get -y --no-install-recommends install  \
-    libtinfo6 libssl1.1 zlib1g python3-numpy python3-pandas \
+    libtinfo6 libssl1.1 zlib1g python3-numpy python3-pandas lit\
```
corepointer, line 67: lit is not a Ubuntu deb package - please remove this. Intended functionality is already provided by the updated Docker images (on the main branch) to fix the tests of this PR

doc/GettingStarted.md
```diff
@@ -59,6 +59,7 @@ launching DAPHNE via Docker (see below) should work the same way as in a native
 | java (e.g. openjdk)                  | 11 (1.7 should be fine)   |                                                                                                                                         |
 | gfortran                             | 9.3.0                     |                                                                                                                                         |
 | uuid-dev                             |                           |                                                                                                                                         |
+| lit                                  | 17.0                      | Install with pip, `llvm-lit` is installed in thirdparty/build/llvm-project/bin                                                         |
```
corepointer, line 62: Name of the package is llvm-10-tools on Ubuntu 20 (our primarily supported Ubuntu version)

install-ubuntu-packages.sh
```diff
@@ -16,4 +16,4 @@
 
 # This is a convenience script to install the required packages on Ubuntu 20+ systems to compile DAPHNE
 sudo apt install build-essential clang cmake git libssl-dev libpfm4-dev lld ninja-build pkg-config python3-numpy \
- python3-pandas default-jdk-headless gfortran uuid-dev wget unzip jq
+ python3-pandas default-jdk-headless gfortran uuid-dev wget unzip jq lit
```
corepointer, line 19: llvm-10-tools

src/api/cli/DaphneUserConfig.h
```diff
@@ -42,13 +42,15 @@ struct DaphneUserConfig {
...
     bool use_phy_op_selection = true;
+    bool use_mlir_codegen = false;
     bool cuda_fuse_any = false;
     bool vectorized_single_queue = false;
     bool prePartitionRows = false;
     bool pinWorkers = false;
     bool hyperthreadingEnabled = false;
     bool debugMultiThreading = false;
     bool use_fpgaopencl = false;
+    bool hybrid = false;
```
corepointer, line 53: The name hybrid alone is not very describing. What about mlir_hybrid_codegen?

corepointer (approved):
I made a quick pass over the changes, mainly concentrating on changes over additions as a deep dive into the introduced passes etc is a bit much to squeeze into my schedule at the moment :see_no_evil: 
I guess there will be minor trouble compiling this with CUDA=ON (I only ran tests on the base image) but this should be sorted out quickly. 
Please note the few comments I added. Otherwise it LGTM.

containers/daphne.Dockerfile
philipportner: This is indeed a mistake, I'll remove this.

doc/GettingStarted.md
philipportner: Thanks, I'll correct this.

src/api/cli/DaphneUserConfig.h
philipportner: You'r right, I changed it as suggested.

pdamme:
Thanks for the feedback, @corepointer, and for incorporating it, @philipportner. I'll merge this PR now.

PR #638, opened 2023-11-14
613-Support missing aggregation functions in aggregation kernels

inikokali:
**_Issue_**
`var` is not supported for any of `AggAll`, `AggRow`, `AggCol` and `stddev` is not supported for full aggregation `(AggAll)` and row-wise aggregation `(AggRow)`.

**_Implementation_**
This pull request proposes implementation for the above aggregation functions following the implementation of `MEAN` aggregation function in each instance.  

**_Changes_**

-  Changes in files `AggAll.h`, `AggRow.h`, `AggCol.h` related to the implementation of `stddev` and `var`.
-  Updated file `AggOpCode.h` to add `var` opCode.
-  Updated file `kernels.json` to add `stddev` and `var` opCodes where needed. 
-  Include testing scenarios to ensure the validity of my implementation. There are tests for both `stddev` and `var` functions for all three kernels (`AggAll`, `AggRow`, `AggCol`). The tests can be found in the related test cases  ( `AggAllTest.cpp`, `AggRowTest.cpp`, `AggCoTest.cpp` ).  
-  Updated test cases `matrix_agg.daphne` and `matrix_agg.py`.


inikokali:
In `test.sh` file I forgot to remove the flags for building in container.
`line: 67: ./build.sh $BUILD_CUDA $BUILD_FPGAOPENCL $BUILD_DEBUG --no-deps --installPrefix /usr/local --target run_tests`

pdamme (approved):
Thanks for this contribution, @inikokali. With that, DAPHNE finally supports the aggregation functions SUM, MIN, MAX, MEAN, VAR, and STDDEV consistently for full, row-wise, and column-wise aggregation. Great!

Your code looks good to me and the tests pass. I only did some minor formatting and, more importantly, adapted the unit test cases to check for approximate equality of floating-point numbers. So far, the tiny example data in these test cases was chosen such that all results are "nice" numbers, without a lot of digits after the point. That's, of course, more difficult to construct for VAR and STDDEV. To avoid future test failures from insignificant floating-point deviations, we do approximate comparisons now.

PR #639, opened 2023-11-18
Integration of MPI Testing into test.sh

inikokali:
**_Issue_**

This pull request integrates **mpi testing** into `test.sh`, consolidating the testing operations into a centralized location. Previously, MPI testing occurred in a separate file from the main test.sh script responsible for all testing operations. 

**_Changes_**

- Add `--mpi` flag in `test.sh`.
- Add a testcase about MPI in file `DistributedTest.cpp`
- Minor changes in `Utils.h` and specifically in **runProgram()** so as it can be used to run mpi testing.


aristotelis96:
Thank you @inikokali  for this contribution.  This looks very good to me. We've been using MPI for some time now, but we were lacking proper test cases. With your additions, we can now test MPI using the same Daphne scripts employed for gRPC testing.

I've made a few minor changes:
- test names,
- renamed the old `mpi_testing` bash script to `mpi_example` and
- removed a simple `.daph` script that we used during the early development of MPI.

Issue #572, opened 2023-07-10
Bounds checks for left and right indexing


pdamme:
Left indexing (e.g., `X[100:200, ] = ...`) and right indexing (e.g., `... = X[100:200, ]`) in DaphneDSL, respectively the ops/kernels backing them (`InsertOp`, `ExtractOp`, `SliceOp`, `InsertRowOp`, `ExtractRowOp`, `SliceRowOp`, `InsertColOp`, `ExtractColOp`, `SliceColOp`) do not check if the accessed positions are in bounds. This can lead to undefined behavior and crashes. This problem should be fixed by adding bounds checks in the kernels, and maybe already in the compiler (when possible).

pdamme:
Sure, go ahead!

AlexRTer:
I've experimented with adding a template parameter `VTSel` so that these OPs can receive negative indexes as well
in order to return more accurate error messages.
However, I believe daphne forwards these to the kernel as `size_t` when used in a .daph script anyway?
So I'm wondering whether it is better to instead remove this template parameter again and simply check whether the value is
beyond a certain threshold (ie. `std::numeric_limits<size_t>::max() - 1000000`) and add a "possible overflow" to the error message.

Furthermore, considering a mention in the existing code of possibly making these boundary validations and other checks optional: Would wrapping these in a `#if(n)(def) ... #endif` and setting a compiler flag such as `-DNO_ERR_CHECK` be a viable solution? If that seems reasonable I would open a new issue to discuss this.

pdamme:
**Regarding types:** Indeed, the kernels for `Slice(Row/Col)Op` and `Insert(Row/Col)Op` hardcode the type of the position parameters to `size_t`. Introducing a template parameter for the positions (just like `Extract(Row/Col)Op` already do) makes sense. An instantiation with `int64_t` would allow us to pass negative numbers to the kernels.

[...]

PR #648, opened 2024-01-13
[DAPHNE-#572] Bounds checks for left and right indexing

AlexRTer:
This PR adds boundary checks for the kernels Extract-/Slice-/Insert- Row/Col like mentioned in #572 as well as to DenseMatrix.cpp and Frame.h. To that end, lowerIncl and upperExcl in .daph scripts are now cast to int64_t instead of size_t before forwarding them to these kernels. Previous asserts in these kernels are now always guaranteed.

Furthermore, `compiler-debug-cuda.txt` and `compiler-trace-cuda.txt` have both been added to `.gitignore`.

**Features:**
- Boundary checks for `SliceRow`, `SliceCol`, `InsertRow`, `InsertCol`, `ExtractRow`, `ExtractCol`
  - Check if arguments have matching dimensions and are within bounds then throws `std::runtime_error` or `std::out_of_range` accordingly (e.g. `invalid argument '7' passed to ExtractRow: out of bounds for frame with row boundaries '[0, 5)'`)
- Boundary checks for `DenseMatrix.cpp` and `Frame.h` for constructions with `rowLowerIncl, rowUpperExcl, colLowerIncl, colUpperExcl` (like when called through their `slice` methods)
  - Check if `src` is `null` and whether arguments are within bounds and throws according errors (e.g. `invalid arguments '-1, 1' passed to dense matrix constructor: it must hold 0 <= rowLowerIncl <= rowUpperExcl <= #rows and rowLowerIncl < #rows (unless both are zero) where #rows of src is '5'`)

**Changes:**
- Added template parameter for `lowerIncl` and `upperExcl` so they must not be `size_t`
- `DenseMatrix` and `Frame` now use `int64_t` for `rowLowerIncl, rowUpperExcl, colLowerIncl, colUpperExcl`
- Slice/Insert/Extract arguments in `.daph` scripts are now passed as `SI64` instead of `Size`
  - Added instantiations with `int64_t` to `kernels.json` for Slice-/Insert- Row/Col
  - Slight change to `test/codegen/ew.mlir` to reflect above changes
- Add tests for `std::out_of_range` errors for all above kernels and added `InsertRowTest.cpp`, `InsertColTest.cpp`
- `Frame.h` had an assert (line 266) that was conditional before, it is now always checked as the bound checks allow the previous exception
- Added `compiler-debug-cuda.txt` and `compiler-trace-cuda.txt` to `.gitignore`


pdamme (approved):
Thanks for this contribution, @AlexRTer! Bounds checks for left and right indexing are very useful and will help us to discover bugs in DaphneDSL scripts much quicker.

**Overall, the code looks good to me.** I still polished it a bit (see the commit I added), but those were only syntactical changes.

-----

In the future, we should also add script-level test cases for out-of-bounds left/right indexing in DaphneDSL in addition to the unit test cases of the `Insert`/`Slice`/`Extract`-kernels you provided with this PR. Those could go to `test/api/cli/indexing/`.

PR #651, opened 2024-01-22
#650 fix get topology

### LINKED ISSUE #650: [BUG] `get_topology` function does not destroy `hwloc_topology_t` instance, leading to `Too many open files` error and segfault
The `get_topology` function is responsible to set up several `vector`s with the correct size based on the underlying topology.

 https://github.com/daphne-eu/daphne/blob/ccba9387a9b33e5c6b46df0aa3477d3bbab0fbfd/src/runtime/local/vectorized/MTWrapper.h#L87-L96

However, the `topology` object keeps a file open while alive, and this object is not destroyed.
This issue leads to a sneaky `Too many open files`error, and makes `hwloc` return nonsense values which cause segfault later on in the program.

`get_topology` should call at the end of the scope by calling
```c++
hwloc_topology_destroy(topology);
```
doc: https://www.open-mpi.org/projects/hwloc/doc/v2.10.0/a00139.php#ga9f34a640b6fd28d23699d4d084667b15

I will prepare a pull request

GuilloteauQ:
The `get_topology` function is returning weird values since the usage of `hwloc` #563.

https://github.com/daphne-eu/daphne/blob/ccba9387a9b33e5c6b46df0aa3477d3bbab0fbfd/src/runtime/local/vectorized/MTWrapper.h#L87-L96

Namely, the function was just constructing vectors of the correct sizes, but not actually setting the information.

This PR uses `hwloc` to return the same information as before #563 and fix #650

### DIFF
--- a/src/runtime/local/vectorized/MTWrapper.h
+++ b/src/runtime/local/vectorized/MTWrapper.h
@@ -85,14 +85,16 @@ protected:
     }
 
     void get_topology(std::vector<int> &physicalIds, std::vector<int> &uniqueThreads, std::vector<int> &responsibleThreads) {
-	hwloc_topology_t topology;
+        hwloc_topology_t topology;
 
-	hwloc_topology_init(&topology);
-	hwloc_topology_load(topology);
+        hwloc_topology_init(&topology);
+        hwloc_topology_load(topology);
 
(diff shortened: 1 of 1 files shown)


philipportner:
Hi @GuilloteauQ , great catch on the missing call to `destroy`, your fix for #650 and the topology walk look good to me. Thank you for the contribution.

Issue #627, opened 2023-10-20
Optimizing Code Generation For Matrix Multiplication


pdamme:
Thanks @philipportner for the issue description!

## Motivation

General matrix multiplication (GEMM) is not only a core operation used in many machine learning algorithms, but also one of the heavy-hitters in regards to the runtime performance of these algorithms. DAPHNE provides a highly optimized implementation of the GEMM operator through a pre-compiled kernel from the openBLAS library. Recently, a different implementation, which generates low-level MLIR instructions for the `MatMulOp`, has been introduced to DAPHNE. In contrast to the pre-compiled kernel, this code generation approach support various value types and enables further optimizations.

However, the currently generated code for the GEMM operator is a naive iterative implementation of the `MatMul`-operator and thus is not competitive in terms of runtime performance with the openBLAS implementation.

## Task

This project aims to bring a more optimized, but still codegen-based, implementation of the `MatMulOp` to DAPHNE. The currently produced IR (see below) is a non-optimized version of the naive iterative matrix multiply algorithm. This produces `affine.for`-loops and `memref` operations for the loop structure and data access. The `llvm.intr.fma` operation performs a fused-multiply-add with the loaded values which is finally stored in the output memref.

```
affine.for %arg0 = 0 to 10 {
  affine.for %arg1 = 0 to 10 {
    affine.for %arg2 = 0 to 10 {
      %12 = memref.load %dm0[%arg0, %arg1] : memref<10x10xf64>
      %13 = memref.load %dm1[%arg1, %arg2] : memref<10x10xf64>
      %14 = memref.load %output[%arg0, %arg2] : memref<10x10xf64>
      %15 = llvm.intr.fma(%12, %13, %14)  : (f64, f64, f64) -> f64
      memref.store %15, %output[%arg0, %arg2] : memref<10x10xf64>
    }
  }
}
```

To improve this implementation, one could start by getting familiar with GEMM optimization techniques (see [1, 2, 3, 4]). Play around with DAPHNE and read through the code of the current MatMulOp lowering pass (src/compiler/lowering/LowerMatMulPass.cpp). Once familiar, start implementing optimizations such as reducing the accesses to our output memref by adding an accumulator (see `iterativeMatmulRegisterSum` in [2]), this will get you used to writing MLIR code for DAPHNE and understanding the produced IR. Additionally, optimizations such as tiling, vectorization, parallelization, and a combination of these could be implemented.

[...]

PR #653, opened 2024-01-25
Code Generation for Matrix Multiplication

resting-dove:
Work for the project in "Large Scale Data Engineering" at TU Berlin WS 23/ 24. The project is supervised by @philipportner. The issue is https://github.com/daphne-eu/daphne/issues/627.

The work mainly concerns an extension to the MatMulLoweringPass of Daphne. 

Several command line options are introduced:
`--mlir-codegen` lowers to the naive structure with three loops.
`--matmul-tile` attempts to find tile sizes such that Register, L1, L2 and L3 caches are reused more efficiently.
`--matmul-fixed-tile-sizes=4,4` can be used to specify up to 5 tile sizes to be used in a tiling scheme adapted from https://github.com/bondhugula/llvm-project/blob/hop/mlir/docs/HighPerfCodeGen.md.
`matmul-unroll-factor=2` additionally unrolls the inner most loop in the tiling scheme by up to the specified factor.
`matmul-unroll-jam-factor=4` additionally unroll jams the original inner most two loops of the tiled nest, if the tile sizes divide the loop size.
`--matmul-vec-size-bits=64` attempts to use vector instructions with the specified bitwidth, but at least one element. Is only executed, if the resulting vector size divides the matrix size.
`--matmul-num-vec-registers=16`gives the number of vector registers available for automatic tile size calculation.

Automated test are added under `test/api/cli/codegen`. 

Evaluations of the options are reported in https://github.com/resting-dove/lde_project.


src/compiler/execution/DaphneIrExecutor.cpp
```diff
@@ -300,23 +302,34 @@ void DaphneIrExecutor::buildCodegenPipeline(mlir::PassManager &pm) {
...
 
-    pm.addPass(mlir::daphne::createEwOpLoweringPass());
+    // pm.addPass(mlir::daphne::createEwOpLoweringPass());
     pm.addPass(mlir::createConvertMathToLLVMPass());
-    pm.addPass(mlir::daphne::createModOpLoweringPass());
+    // pm.addPass(mlir::daphne::createModOpLoweringPass());
     pm.addPass(mlir::createCanonicalizerPass());
     pm.addPass(mlir::createCSEPass());
-    pm.addNestedPass<mlir::func::FuncOp>(mlir::createLoopFusionPass());
+    //pm.addNestedPass<mlir::func::FuncOp>(mlir::createLoopFusionPass());
```
philipportner, line 325: Please make sure the normal pipeline still runs as expected.

src/compiler/execution/DaphneIrExecutor.cpp
```diff
@@ -300,23 +302,34 @@ void DaphneIrExecutor::buildCodegenPipeline(mlir::PassManager &pm) {
...
     pm.addNestedPass<mlir::func::FuncOp>(
         mlir::createAffineScalarReplacementPass());
     pm.addPass(mlir::createLowerAffinePass());
-
+    mlir::LowerVectorToLLVMOptions lowerVectorToLLVMOptions;
+    lowerVectorToLLVMOptions.enableX86Vector(true);
```
philipportner, line 330: What happens if this run on an ARM system? You may want to check and use `lowerVectorToLLVMOptions.enableArmSVE();` instead.

src/compiler/lowering/MatMulOpLowering.cpp
```diff
@@ -14,198 +14,564 @@
...
-    auto rowLoop = rewriter.create<AffineForOp>(loc, 0, lhsShape[ROW], 1);
-    for (Operation &nested : *rowLoop.getBody()) {
-        rewriter.eraseOp(&nested);
+struct LowerMatMulOpOptions {
+  LowerMatMulOpOptions() {}
+  int vec_size_bits{0};
+  bool vectorize{false};
+  bool tile{false};
+  bool useFixedTileSizes{false};
+  int register_size{4 * 4 * 64};
```
philipportner, line 83: Which register size is this referring to?

src/compiler/lowering/MatMulOpLowering.cpp
```diff
@@ -14,198 +14,564 @@
...
+  auto a = rewriter.create<AffineLoadOp>(
+      loc, lhs,
+      ValueRange{rowLoop.getInductionVar(), fmaLoop.getInductionVar()});
+  auto b = rewriter.create<AffineLoadOp>(
+      loc, rhs,
+      ValueRange{fmaLoop.getInductionVar(), colLoop.getInductionVar()});
+  auto c = rewriter.create<AffineLoadOp>(
+      loc, output,
+      ValueRange{rowLoop.getInductionVar(), colLoop.getInductionVar()});
+  // split out into add and multiply
```
philipportner, line 172: Either remove the comment or split the `FMAOp`, both are fine.

src/compiler/lowering/MatMulOpLowering.cpp
```diff
@@ -14,198 +14,564 @@
...
+llvm::SmallVector<AffineForOp, 3>
+vectorizedAffineMatMul(mlir::Value &lhs, mlir::Value &rhs, mlir::Value &output,
+                       ConversionPatternRewriter &rewriter, mlir::Location loc,
+                       ArrayRef<int64_t> lhsShape, ArrayRef<int64_t> rhsShape,
+                       mlir::MLIRContext *ctx,
+                       llvm::SmallVector<AffineForOp, 3> &loops,
+                       Type elementType, int64_t vec_size) {
+  auto vec_Type = mlir::VectorType::get({vec_size}, elementType);
+
+  // TODO: We need an option to enable smaller vector sizes for the ends of each
```
philipportner, line 199: TODO, in case this is out of scope for this PR, maybe it is wort it to create an issue with more explanation and  reasoning for the suggested change and reference the issue here as `TODO(#xxx)`

src/compiler/lowering/MatMulOpLowering.cpp
```diff
@@ -14,198 +14,564 @@
...
+      return false;
+    }
+    if (!matrixElementType.isa<FloatType>()) {
+      return false;
+    }
+    return true;
+  }
+
+  // TODO: Test why not
+  // Cannot tile non square matmul.
```
philipportner, line 264: This would be nice to have, especially for your benchmarks. In case you run into too much troubles fixing this, handle it in the same way as the `TODO` above.

src/compiler/lowering/MatMulOpLowering.cpp
```diff
@@ -14,198 +14,564 @@
...
+
+    auto lhsRows = lhsMatrixType.getNumRows();
+    auto lhsCols = lhsMatrixType.getNumCols();
+
+    auto rhsRows = rhsMatrixType.getNumRows();
+    auto rhsCols = rhsMatrixType.getNumCols();
+
+    auto matrixElementType = lhsMatrixType.getElementType();
+    // Cannot lower integer matmul, because we cannot create Memrefs with
+    // integer type at the moment.
```
philipportner, line 289: Integer memrefs should generally work, see `src/compiler/lowering/EwOpsLowering.cpp`. Handle this as already discussed in our meeting, so similarly to the `TODO` above.

src/compiler/lowering/MatMulOpLowering.cpp
```diff
@@ -14,198 +14,564 @@
...
+           "tTN: 2 should have step size 1."); // k loops
+    assert(twiceTiledNest[7].getStep() == 1 &&
+           "tTN: 7 should have step size 1.");
+
+    // permute loops to final order (i / MC, j / NC, k / KC, i / MR, i mod MR, j
+    // / NR, j mod NR, k mod KC) ->
+    //                              (j / NC, k / KC, i / MC, j / NR, i / MR, k
+    //                              mod KC, j mod NR, i mod MR)
+    // TODO: This assert only fails in debug mode?!
+    // assert(isValidLoopInterchangePermutation(twiceTiledNest, {2, 0, 1, 4, 7,
```
philipportner, line 460: Same as the `TODO` above.

src/compiler/lowering/MatMulOpLowering.cpp
```diff
@@ -14,198 +14,564 @@
...
+      }
+    return sizes;
+  }
 };
-}  // end anonymous namespace
+} // end anonymous namespace
 
 void MatMulLoweringPass::runOnOperation() {
+  auto module = getOperation();
+  {
```
philipportner, line 574: I think this additional scope can be removed.

src/ir/daphneir/Passes.td
```diff
@@ -71,6 +67,180 @@ def MapOpLoweringPass: Pass<"lower-map", "::mlir::func::FuncOp"> {
     let constructor = "mlir::daphne::createMapOpLoweringPass()";
 }
 
+def MatMulOpLoweringPass: Pass<"lower-mm", "::mlir::ModuleOp"> {
+  let summary = "Lower the MatMulOp acting on Daphne matrices to an affine loop structure.";
+  let dependentDialects = ["vector::VectorDialect", "mlir::LLVM::LLVMDialect", "mlir::AffineDialect",
+                            "mlir::memref::MemRefDialect"];
+  let constructor = "mlir::daphne::createMatMulOpLoweringPass()";
+  let description = [{
```
philipportner, line 75: Very good and thorough description!

philipportner (changes requested):
Very good work @resting-dove ! A few small things to iron out but nothing major. You did a really good and thorough job improving our matrix multiplication codegen.

src/compiler/execution/DaphneIrExecutor.cpp
resting-dove: I moved some of those passes before the matmul lowering, so that the loop fusion pass doesn't interfere with the loops we create that are potentially intentionally tiled.

src/compiler/execution/DaphneIrExecutor.cpp
resting-dove: As far as I could tell, it didn't actually produce any x86 specific instructions so I removed it.

src/compiler/lowering/MatMulOpLowering.cpp
resting-dove: It took me a while to really understand what Uday Bondhugula was referring to with this in the blog post. It seems to mean the SIMD vector size x how many vectors there are. I changed the variable name to something that should make this more clear.

src/compiler/lowering/MatMulOpLowering.cpp
resting-dove: I will create a PR for this.

src/compiler/lowering/MatMulOpLowering.cpp
resting-dove: Seems to work currently also for non square matrices. I will do some more testing.

src/compiler/lowering/MatMulOpLowering.cpp
resting-dove: The original problem was three fold. 1. Filling the output memref with correctly signed 0 integer values, when we so far only used arith::ConstantOps which are MLIR unsigned. 2. Similarly arith::AddIOp only operates on signless types. 3. SumAllOp could not handle integer types for the same reasons.
In EwOps this is handled by casting from signfull to signless where necessary. I took this approach and applied it to MatMulLowering and SumAllOpLowering.

src/compiler/lowering/MatMulOpLowering.cpp
resting-dove: I think this check simply cannot deal with for loops that have more complex step lengths etc. Thus, I removed this check. The other checks should ensure we know where our loops are and therefore that this is a valid interchange.

src/ir/daphneir/Passes.td
resting-dove: Thanks!

philipportner:
Thanks again @resting-dove , took a while but the changes are now merged in. Great work improving and testing optimized code generation for matrix multiplication!

Issue #622, opened 2023-10-02
Error creating matrix from function arguments?


divjakm:
I don't know whether I'm doing something wrong or is there a Daphne DSL error in this example:

```
def test (a:f64, b:f64, c:f64)
{
    //A = a + b + c;          // works
    //A = [1.0, 2.0, 3.0];    // works
    A = [a, b, c];        // error
    print(A);
}
test (1.0, 2.0, 3.0);
```
I get the following error:
`Parser error: test7.daph:40:9 mismatched input 'a' expecting {']', 'true', 'false', INT_LITERAL, FLOAT_LITERAL, STRING_LITERAL}`

Any help is very appreciated!

pdamme:
Hi @divjakm, so far, matrix literals in DaphneDSL only support parse-time *literals* for their elements, while *expressions* are not supported yet (not even simple ones like just a variable).

However, we definitely want to support expressions for the elements, and there is an issue for that #523. We would highly welcome a pull request for that issue, from anyone who would like to help.

This missing feature has been reported before in #584, where you can also find a workaround. Put shortly, you could currently do something like `A = rbind(rbind(as.matrix(a), as.matrix(b)), as.matrix(c));` or `A = fill(0.0, 3, 1); A[0, ] = as.matrix(a); A[1, ] = as.matrix(b); A[2, ] = as.matrix(c);`. But ultimately, `A = [a, b, c];` will work.

Issue #523, opened 2023-04-25
Flexible matrix/frame literals in DaphneDSL


pdamme:
**Motivation:** DaphneDSL’s matrix literals are a convenient way to specify small matrices. For instance one can write `X = [1, 2, 3];` to get a (3x1) matrix containing the values 1, 2, and 3. This feature is very handy for writing small examples and test cases, and is also useful in complex scripts. Nevertheless, there are still a few points which could be made more convenient. Furthermore, besides matrices, DaphneDSL also supports frames, but there are no frame literals yet.

**Task:** This project is about adding more syntactic sugar around matrix and frame literals in DaphneDSLs. Implementation in C++. In particular, the following features should be supported:
- *Mixed value types:* Currently, all values in a matrix literal must be of the same value type. Ensure that matrix literals like `[1, 2.2]` or `[false, 3]` are allowed. The value type of the resulting matrix should be the most general of the values types of the elements. You can use `src/compiler/inference/TypeInferenceUtils.h` to determine the most general value type.
- *Expressions for the matrix elements:* Currently, all values in a matrix literal must themselves be literals, i.e., they must be known at parse-time. Ensure that the elements can be the results of complex expressions such as `[sqrt(2), 3.1, sum([1, 4, 2] + fill(x, 3, 1))]`. In such a case, the matrix cannot be constructed at parse-time. Instead, the expressions should be evaluated at runtime and operations should be generated to compose the overall matrix out of them.
- *Shape:* Currently, matrix literals are a linear list of *n* elements and result in a *(n x 1)* matrix, i.e., *n* rows and *1* column. Any other shape must be achieved by explicitly reshaping the matrix. For instance `reshape([1, 2, 3, 4], 2, 2)` creates a (2x2) matrix. This is a bit verbose; a notation like `[1, 2, 3, 4](2x2)` would be great.
- *Frame literals:* So far, DaphneDSL does not support frame literals. Instead, frames can be constructed from column matrices using the `createFrame()` built-in function, e.g., `createFrame([1, 2, 3], [1.1, -2.2, 3.3], "a", "b")`. As a first step, a simplification should be supported, such as `{"a": [1, 2, 3], "b": [1.1, -2.2, 3.3]}` (where the order of the columns is as specified).
[...]

PR #675, opened 2024-03-07
[DAPHNE-#523] DaphneDSL flexible matrix/frame literals

### LINKED ISSUE #584: Hand crafted Frame does not accept variables?
I can run this without a problem:

```
y = createFrame(
    [  0,   5],
    [750, 450],
    "d", "c");
print(y);
```

Yet, if i try it this way around, it gives me an error:
```
tn = 0;
fp = 12;
fn = 10;
tp = 5;

cf = createFrame(
     [tn, fp], 
     [fn, tp], 
     "neg", "pos");
print(cf);

```

> Parser error: <linenumber> mismatched input 'tn' expecting {']', 'true', 'false', INT_LITERAL, FLOAT_LITERAL, STRING_LITERAL}

AlexRTer:
This PR implements #523 and allows for matrix literal elements to be expressions of different value types instead of only literals with the same value type. Furthermore, matrix literal dimensions can now be set directly using parentheses (`[...](dim_x, dim_y)`) instead of having to call `reshape`. Lastly, it adds new syntax to create column-major as well as row-major frames in DaphneDSL.
It also resolves #584 and #622.

*Features:*
- Flexible value types: Matrix and frame literals are allowed to have elements of different value types. These are then casted to the most general value type among them (or within a column for frames). Currently supported types are `double`, `int64_t`, and `bool` (in order of generality).
  - `[true, 2, 3.0]` becomes a matrix of `double` type
- Complex expressions for elements: Matrix and frame literal elements can now be expressions. This includes (but is not limited to) built-in as well as user-defined functions, and (prior defined) variables. Their output only needs to be of a supported type (see above). Currently, there is a limitation for row-major frames as explained below.
  - `[1, sqrt(9), sum([1, 4, 2] + fill(3, 3, 1))] -> [1, 9, 16]`
- Reshaping: Matrix literals can now be reshaped by simply adding parentheses behind the matrix literal with comma-separated expressions for its dimensions. Both dimensions can be complex expressions. Matrices default to column-matrices if no dimension is given or infer missing dimensions if only 1 has been given.
  - `[1, 2, 3, 4] -> (3 x 1)`, `[1, 2, 3, 4](1,) -> (1 x 3)`, `[1, 2, 3, 4](2,2) -> (2 x 2)`
- Frame literals: Frames can now be created using curly brackets. The syntax differs slightly for column-major and row-major format:
  - column-major: columns are given as comma-separated pairs of `label : matrix`, where each matrix contains all values of a column. 
    E.g. `{"col 1" : [1, 2], "col 2" : [1.1, 2.2]}` (1st column of type `int64_t` and 2nd column of type `double`)
  - row-major: labels are given within brackets first, then comma-separated matrices define each row.
    E.g. `{["col 1", "col 2"], [1, 1.1], [2, 2.2]}` (equivalent to example above)

*Changes:*
- Matrix syntax has been extended and new syntax has been added for frames and row-major frames in [DaphneDSLGrammar.g4](https://github.com/daphne-eu/daphne/blob/main/src/parser/daphnedsl/DaphneDSLGrammar.g4)
- New helper function `buildColMatrixFromValues` for handling matrix construction has been added to [DaphneDSLVisitor](https://github.com/daphne-eu/daphne/blob/main/src/parser/daphnedsl/DaphneDSLVisitor.cpp). Furthermore, the visitor for matrices has been modified and new visitor functions for frames have been added
- `castF64If` has been added to [ParserUtils](https://github.com/daphne-eu/daphne/blob/main/src/parser/ParserUtils.h)
- 1 new instantiation for the `insertRow` and `Fill` kernels have been added to handle special cases when creating matrices of `bool` type
- The [LanguageRef](https://github.com/daphne-eu/daphne/blob/main/doc/DaphneDSL/LanguageRef.md) doc has been edited to reflect the above changes

*Limitations:*
- Row-major frame literals currently do not support labels and matrices to be passed as variables (i.e. `row=[1, 2]; rowFrame={["col 1"], row};`). This is because of issues in inferring the most general value type of a column which is lost if the visitor for matrix literals is called on them beforehands and casts all values to the most general value type of each row before they can be rearranged to make up the columns instead.

In the future, this could also be extended to support matrix literals where an element itself is a (flat) matrix by adding the matrix's element type to `valueTypes` and adding `slice` and `rbind` operations at the end of the `buildColMatrixFromValues` helper function.


AlexRTer:
It appears there is an issue when running `./test.sh`.
Running `./test.sh ~"EwBinary*"` passes all tests. However, so does `./test.sh "EwBinary*"`.
Only when running all tests it seems to get stuck on line 35 of [EwBinaryScalarTest.cpp](https://github.com/daphne-eu/daphne/blob/main/test/api/cli/codegen/EwBinaryScalarTest.cpp) which calls `runDaphne`, but I can't make out why.

AlexRTer:
> It appears there is an issue when running `./test.sh`. Running `./test.sh ~"EwBinary*"` passes all tests. However, so does `./test.sh "EwBinary*"`. Only when running all tests it seems to get stuck on line 35 of [EwBinaryScalarTest.cpp](https://github.com/daphne-eu/daphne/blob/main/test/api/cli/codegen/EwBinaryScalarTest.cpp) which calls `runDaphne`, but I can't make out why.

Specifically, the `runDaphne` call gets stuck after calling `runProgram` in the same [Utils.h](https://github.com/daphne-eu/daphne/blob/main/test/api/cli/Utils.h#L95) file and reaching line 95 where it attempts to read from an stdout buffer. Nothing within the loop seems to be executed.

I still do not know why, but omitting the final `ReshapeOp` in the matrix literal visitor and returning `colMatrix` instead of its `result` does not cause it to get stuck and instead runs all tests successfully (except the new one specifically testing the reshaping). So the problem seems to stem from here:
```cpp
mlir::Value result = static_cast<mlir::Value>(builder.create<mlir::daphne::ReshapeOp>(loc, utils.matrixOf(valueType), colMatrix, rows, cols));
```

pdamme:
Thanks for this PR @AlexRTer. I can confirm the problems with the test cases, but I don't see the reason at first glance. I will look deeper into it.

AlexRTer:
Thank you very much for your help

AlexRTer:
I added some negative tests for matrix and frame literals.

Currently Row-major Frames allow for different column lengths, as long as the total amount of elements is a multiple of the amount of columns (the elements are simply split up evenly to all columns). For Column-major Frames, shape inference catches this case.
Seeing as Column/Row -major notation is equivalent, perhaps it would be better to convert Row-major notation into Column-major notation before visiting the Frame expressions. Ideally, this would simplify Frame Literals to a single Visitor for both notations. Does Daphne currently have a way to do this?

> It appears there is an issue when running `./test.sh`. Running `./test.sh ~"EwBinary*"` passes all tests. However, so does `./test.sh "EwBinary*"`. Only when running all tests it seems to get stuck on line 35 of [EwBinaryScalarTest.cpp](https://github.com/daphne-eu/daphne/blob/main/test/api/cli/codegen/EwBinaryScalarTest.cpp) which calls `runDaphne`, but I can't make out why.

Thanks to c2900a3, the above issue has been resolved.

However, there seems to be an error with `codegen/CodegenTest.cpp` (https://github.com/daphne-eu/daphne/actions/runs/9034047874/job/24825710151#step:6:28) after the last commit that I cannot reproduce. The last commit 0e7090a only added a few matrix/frame tests and minor formatting changes. On the other hand, the tests beforehand (9bf8281) (https://github.com/daphne-eu/daphne/actions/runs/9033304615/job/24823355246) as well as my tests with the latest build both ran succesfully.

pdamme (changes requested):
Hi @AlexRTer, thanks for already addressing (some of the) points we discussed offline. Overall, this contribution already **looks good** to me.

I've polished the updates to the DaphneDSL language reference a bit and applied some small changes and simplifications to the code (see the commit I added for details and let me know in case you disagree with any of those).

**Required changes before the merge:**
1. Please **make sure that all rows in a frame literal of rows must have the same length**. This could otherwise be a frequent source of mistakes. I don't think that we need to convert from by-row to by-col syntax of a frame literal to that end. My suggestion is: In the DaphneDSL grammar, introduce a new parser rule for a single row of a frame literal of rows, and use that in the definition of `rowMajorFrameLiteralExpr`. Then you should be able to loop over all rows and get the number of elements for each row, such that you can compare them. Please also add a test case on that.
1. Please **add test cases for all 4 combinations of #rows and/or #cols given in matrix literals** (see `matrix_literal_success_1.daphne`).

**Optional changes:** *(can be done either now or after the merge)*
1. Please **support all value types in matrix/frame literals**. The current restriction to `f64`, `si64`, and `bool` seems unnecessary and is not intuitive from a user's point of view. Then, `matrix_literal_failure_2.daphne` should be removed: `si8` should not be an invalid value type of a matrix literal.
1. **Allow expressions for the column labels in frame literals of rows**. Is there a reason why this would be more difficult than for frame literals of columns?

**Suggestions for follow-up issues:** *(should not be addressed in this PR anymore)*
1. **Support empty matrix literals**, since they would be quite convenient and it's not intuitive that they are not supported. They could use a default value type, e.g., the least general one, such that they don't dominate type inference (just an idea).
1. In general, any intermediate result (or, `mlir::Value`) could be of **unknown type** *at parse-time*. The same holds for the string-values labels. Their types may become known only during type inference. Here's a simple example that should work, but doesn't do so at the moment (because the type of the extracted value is unknown at parse-time):
   ```r
   f = createFrame([1], "a");
   print( [ as.scalar(f[0, "a"]) ] );
   ```

**Feel free to share your opinion on these comments.**

-----

By the way: In terms of wording, I recommend avoiding the terms *row-major/column-major frame literals*, because those terms are normally used for storage layouts. However, the way a frame literal is specified is just a matter of convenience and does not imply any physical layout. I suggest *frame literal of rows/columns*, instead.

AlexRTer:
Thanks for your feedback and editing @pdamme, I like the changes you made.

I implemented another parser rule like you suggested which fixes the issues with verifying frame literal row/column sizes, and allows expressions to be labels for frame literals of rows. Tests for matrix literals with one or no dimensions specified have also been added.

Before you merge it, I would like to refactor the code a bit and look into supporting the missing value types. Also, I still need to edit the documentation regarding support for expressions as labels.

pdamme (approved):
Thanks for addressing all required and optional changes, @AlexRTer. **This PR is ready to be merged**.

Issue #357, opened 2022-05-06
DaphneDSL: Function to determine the type of an object


RetoKrummenacher:
Similar to R's typeof(), it would be nice to have a function to check the object type in DaphneDSL. The output can be a pure string, meaning it is a feature to get information and not to be used in casting operations.

Some possible examples could be:
```
f = 5.0;
typeof(f); # --> f64 
d = reshape(seq(1.0,5.0,1.0),5,1);
typeof(d1); # --> matrix double
i = reshape(seq(1,5,1),1,5);
typeof(i); # --> matrix int64
```

This is useful when working with type casting. As printing alone does not provide enough information.

pdamme:
I agree, this would be a useful feature, especially while debugging DaphneDSL scripts. We will implement it.

pdamme:
This is actually an interesting "good first issue", which provides an opportunity to take a tour through various layers of the system. Here are a few suggestions on addressing it.

Ultimately, we want a DaphneDSL built-in function `typeof`, which returns the data/value type of its single argument as a string. I propose it should return the physical data type (e.g., `DenseMatrix`/`CSRMatrix`) instead of the logical data type (e.g., `matrix`), since that will be more useful for debugging. A few examples for possible outputs: `f64`, `si32`, `str`, `DenseMatrix<f64>`, `CSRMatrix<ui8>`, `Frame<si32, f64, ui8>`.

**Some hints on required changes:**
- *DaphneIR*
  - A new DaphneIR operation (see `src/ir/daphneir/DaphneOps.td`), say `TypeOfOp`, which gets a value of any matrix, frame, or scalar type as input, and yields a string as output.
- *DaphneDSL*
  - A new built-in function, say `typeof` in DaphneDSL (see `src/parser/daphnedsl/DaphneDSLBuiltins.cpp`), which produces a `TypeOfOp` in the IR.
- *Runtime*
  - A new kernel for `TypeOfOp` in `src/runtime/local/kernels/`. See the developer documentation on implementing a new kernel in `doc/development/ImplementBuiltinKernel.md`. We need a kernel-struct, a convenience function, and specializations for all possible data/value types.
  - To concisely handle the combinations of data and value types, the specializations for `DenseMatrix`, `CSRMatrix`, and `Frame` could call those for the scalar types to get the string representation of their value/column type(s).
[...]

PR #677, opened 2024-03-14
[DAPHNE-#659] Scientific notation for float literals and digit separators

### LINKED ISSUE #659: Float literals in scientific notation
Currently, DaphneDSL supports float literals only of the form `123.4567`, plus some special cases like `inf` and `nan`. However, floating-point numbers are also commonly expressed in the scientific notation, e.g. `1.234567e2`. This notation is especially convenient for writing very small or very large numbers, e.g., `10000` becomes `1e5` and `0.000001` becomes `1e-6` (there could be far more zeros).

This task is to support the scientific notation in DaphneDSL, in addition to the currently supported floating-point notations.

**Hints:**
- Find out what syntax and potential corner cases are supported in other programming languages, such as Python, R, C++, etc., and implement that in DAPHNE.
- Search for `#659` in the code base (especially in `scripts/algorithms/`), several points where the scientific notation should be used are marked that way.

AlexRTer:
This PR closes #659. It adds scientific notation for float literals as well as digit separators for both float and integer literals to improve readability.

*Features:*
- Scientific (E) notation that is case insensitive (accepts `e` and `E`)
  -  e.g. `1.2e3 = 1200`, `1E-3 = 0.001`, both of type `f64`
- Digit separators `'` and `_` for float and integer literals
  - `1e6 = 1'000'000 = 1_000_000`

*Changes:*
- Added scientific notation and digit separators to [DaphneDSLGrammar.g4](https://github.com/daphne-eu/daphne/blob/main/src/parser/daphnedsl/DaphneDSLGrammar.g4)
- Updated marked daphne scripts with scientific notation
- Added a few test cases for literals
- Update to the [LanguageRef](https://github.com/daphne-eu/daphne/blob/main/doc/DaphneDSL/LanguageRef.md) doc


pdamme (changes requested):
Thanks for this contribution, @AlexRTer. It already **looks quite good overall**, but I spotted some debatable points in the DaphneDSL lexer rule `FLOAT_LITERAL` (e.g., no `+` allowed after `e`, although an output can look like this, no `f` allowed with scientific notation, exponent could be sth like `000`). I tried to address those and also added a few more test cases.

Since the literal parsing is quite fundamental and mistakes in this part could cause quite some headache when debugging a DaphneDSL script, I think we should have a more comprehensive test coverage here. **It would be great if you could add systematic test cases** for all possible routes through the `FLOAT_LITERAL` rule (ideally a few different examples for each syntactical case), plus a few negative tests for wrong float literals that should cause syntax errors. Feel free to completely replace the existing `float_*.daphne/txt` files with new ones.

AlexRTer:
Thank you for your feedback, I will go ahead and implement some better testing as well.

AlexRTer:
I changed the grammar slightly to disallow digit separators right behind the decimal point.

The tests now include some negative tests, i.e.
- `infff`, `1.`, `01.0`, `10.0_`, `1.'0`

and cover (corresponding file index in brackets)
- [1] special values (e.g. nan, inf)
- [2] zeros
- [3,4] simple floats with varying length (e.g. `12.34`)
- [5,6] floats with digit separators (e.g. `1_000.0`)
- [7,8] scientific notation without decimal points (e.g. `1e6`)
- [9,10] scientific notation with decimal points (e.g. `0.000000e-6`)
- [11] scientific notation with digit separators (e.g. `1.0_0e6`)

For all tested ordinary values there is a test without signs, with a negative sign, as float type and mixed (negative and float type). The tests also cover small and capital e for scientific notation as well as `+`, `-` signs.

I believe a test to verify that e.g. `1.0f` is of float type is still missing in Daphne. Maybe a built-in `typeof` function would be a good addition to DaphneDSL.

pdamme (approved):
Thanks for adding more test cases and for further improving the DaphneDSL grammar. This PR is **ready to be merged** now.

pdamme:
By the way, there is an open issue on introducing a `typeof()` built-in function in DaphneDSL: #357.

PR #678, opened 2024-03-16
[DAPHNE-#669] properly handle views in `WriteCsv` kernel

### LINKED ISSUE #669: DaphneDSL `write()` handles views incorrectly
The following DaphneDSL script should write the second column of a small matrix to a file:
```R
# Create (3x2) matrix [[1, 2], [3, 4], [5, 6]].
X = reshape(seq(1, 6, 1), 3, 2);
print(X);
# Take just the second column...
Y = X[, 1];
print(Y);
# ...and write it to a file.
write(Y, "Y.csv");
```
Afterwards, `Y.csv` should have the following contents:
```text
2
4
6
```
But it contains:
```text
2
3
4
```
This is a bug. Most likely, the `write`-kernel does not take the `rowSkip` of its input `DenseMatrix` into account when iterating over all values to write.

AlexRTer:
This PR fixes #669 for the case of `.csv` files.
Now `WriteCsv` takes the `rowSkip` argument of the matrix it's iterating over into account.

### DIFF
--- a/src/runtime/local/io/WriteCsv.h
+++ b/src/runtime/local/io/WriteCsv.h
@@ -64,15 +64,17 @@ struct WriteCsv<DenseMatrix<VT>> {
     static void apply(const DenseMatrix<VT> *arg, File* file) {
         assert(file != nullptr && "File required");
         const VT * valuesArg = arg->getValues();
-        size_t cell = 0;
+        const size_t rowSkip = arg->getRowSkip();
+        const size_t argNumCols = arg->getNumCols();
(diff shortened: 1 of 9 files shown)


pdamme (approved):
Thanks for this bug fix, @AlexRTer. I can confirm that it solves the problem.

I added some small script-level test cases that check if the problem is (and stays) solved. In general, for each bug fix, I recommend adding a test case that fails in the old version of the code and passes after the bug fix.

PR #684, opened 2024-04-03
[DAPHNE-###] Build environment for unsupported platforms

corepointer:
This commit adds scripts to build the compiler and necessary libraries to build DAPHNE in unsupported environments (RHEL UBI, CentOS, ...) Besides the build scripts, there is a docker file to create a UBI image to build for Redhat 8.

Usage:
* Create Docker image with ```build-ubi8.sh```
* Run the Docker image with ```run-ubi8.sh```
* Run ```build-all.sh```
* Run ```source env.sh``` inside or outside the Docker image to set PATH and linker variables to use the created environment (you need to cd into the directory containing env.sh as this uses $PWD)

Sorry for the long delay @m-birke, I ran into some issues while making this a "push button solution" (which it isn't anyway, but it boils down to running ```build-all.sh``` now) and then there was the easter break...  :egg: :rabbit: :fried_egg: :see_no_evil: :stuck_out_tongue_winking_eye: 

Please test and tell me if that works for you.

### DIFF
--- a/buildenv/README.md
+++ b/buildenv/README.md
@@ -1,0 +1,22 @@
+<!--
+Copyright 2024 The DAPHNE Consortium
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


corepointer:
Beware that this procedure **needs ~50GB of free disk space**. Also, the provided CUDA SDK **expects a recent driver  (550+)** version. That will most likely be an issue on large installations - exchange the relevant version and file names for another CUDA version in env.sh to build another version.

m-birke:
Hi @corepointer thank you for your work, you will hear from me

m-birke (approved):

Issue #116, opened 2021-09-14
Declaration of `operator==` for matrices is too general


daphne-eu:
In GitLab by @pdamme on Sep 14, 2021, 21:00

Along with the `checkEq`-kernel, in `src/runtime/local/kernels/CheckEq.h` we also declare `operator==` and define it to use `checkEq` for the given data/value types. The intended behavior is that this definition of `operator==` should be used only for our Daphne data structures. However, the way it is currently declared is actually more general and could accidentally be used for comparisons of completely unrelated types. For instance, merely including this header in `src/runtime/local/vectorized/MultiThreadedKernelTest.cpp` yields a C++ compiler error, since the C++ compiler tries to instantiate the template for `mlir::LocationAttr`.

A possible solution could be to make `operator==` a member of each Daphne data structure class.

PR #694, opened 2024-04-18
Tensor Support - 1: Data structure

CPestka:
# Tensor Support

## Part 1: The Data structures - i.e. This Pr

Introduces support for tensors via the data structures:
* ContiguousTensor
* ChunkedTensor

This includes some data layout transformation functionality like dicing and re-chunking, as well as the capability to convert between the relevant types (e.g. ContiguousTensor, DenseMatrix, ChunkedTensor).

## Finished Functionality not included in this pr

* io_uring support
* Zarr (a chunked "file format" for tensors) support - including asyc and partial reads
* A tensor implementation for most of the currently existing kernels for the other data structures (Aggregate, Casts, Binary/Unary-Ops, Slicing/Dicing, etc.)

## Work in progress

The DSL/Compiler integration is currently incomplete and what is there needs some clean up. What we got to work is that one is able to issue a partial of a tensor from a zarr file, create and index a tensor, but most of the other kernels are not wired up to the DSL, at the Compiler level some of the tensor/kernel properties need to be made variadic etc..


pdamme:
Hi @CPestka, thanks for this contribution and sorry for the delay. Support for tensors, Zarr files etc. would be very valuable and highly welcome to DAPHNE. I will look into it soon and provide feedback and/or merge it by Wed next week.

pdamme (changes requested):
Hi @CPestka, overall, **the code looks very good and clean to me**. I have mainly focused on the integration of the new tensor data types into the code base, and just skimmed the implementation of `ChunkedTensor`. Thanks for also addressing #116.

I merged `upstream/main` again and polished the code a bit. Please see the commits I added for details and let me know if you disagree with anything there.

**Required changes before merging:**
1. Please **make the new test cases pass**. DAPHNE can be built successfully, but the new test cases fail, seemingly due to some kind of memory corruption. When I execute `./test.sh ~"Tensor*"` (i.e., all test cases except for the new ones), all tests pass. When I execute `./test.sh "Tensor*"` (only the new test cases) or just `./test.sh` (all test cases), the test execution prints `malloc(): invalid size (unsorted)` and seems to be stuck forever (on my system). Likewise, the CI tests seem to be stuck (I canceled them manually). This problem already occurred before I merged `upstream/main` and adding some polishing changes.
1. **Clarification regarding switch to C++20**: The code builds even when we stay at C++17 (the tests still fail). For what reason do you need C++20? Is it required for some feature beyond this PR (one of those mentioned in the PR description above)? I would be open to switching to C++20, but we should at least double-check if this change could cause any problems with some existing DAPHNE deployments on older systems.

**Optional changes:** *(recommended, but can be done after the merge)*
1. Please **add some documentation** to the code (e.g., doxygen-style), i.e., a short description of the most important classes, member variables, member functions etc. I think especially the idea behind `ChunkedTensor` is not self-evident and its layout not clear at first glance.
1. The output format of `ChunkedTensor::print()` seems to deviate from that of the other data structures, please make it consistent.
1. `ContiguousTensor::ContiguousTensor(const DenseMatrix<ValueType> *other)`: I recommend double-checking the line `strides = {1, other->getNumCols()};`. I think it should be `other->getRowSkip()`, since `other` could be a view into a larger `DenseMatrix`. Maybe add a test case on that.

CPestka:
@pdamme Thanks, for having a look.

Your changes look good to me. Silent errors should not be a thing ^^

As requested, I adjusted the formatting of print() for the chunked tensor and added a bit of documentation to the classes.
Good catch within the non-copying "conversion"-constructor, I forgot about that.

Regarding C++20:

Sry, I should have addressed that initially. As far as I remember, nothing in this Pr explicitly depends on any C++20 features and thus it could have been left out of this Pr. However, all of the follow up Prs will "need" C++20 due to different reasons and having it in this initial Pr would make it slightly more convenient to make the other Prs in parallel. If you feel like removing this change from this Pr would be better and moving it to the followup prs that actually need it, I would also be ok with that.

Why C++20:
1. std::endian and friends. Needed for the zarr pr. (Currently also ignored in the daphne serializer if i remember correctly) Could also be found out programmatically, but AFAIK most/(all?) of the methods are technically UB in C++ and thus might also just stop working randomly especially on higher optimization lvls.
2. wait() and notify() of atomic variables (not condition variables themselves) (used in the io_uring "thread pool") Could be replaced by using the OS APIs directly, this would of course mean at least a couple of OS specific implementations.

Also I think I used some C++20 features of the filesystem API, which is mostly C++17 otherwise, if I remember correctly.

C++20 breaking things:
Yes, it indeed broke some things. Mostly on the build system side though, by, i think, changing the include order somewhere. This is the reason the fix #116 was included.

Regarding the failing tests:
On my machine all tests pass, including the newly added one.
Screenshot from 2024-05-23 14-22-22
Screenshot from 2024-05-23 14-25-13
Screenshot from 2024-05-23 14-28-31
Screenshot from 2024-05-23 14-28-43
On marcus machine they pass as well.
The screenshots are from the most recent version of the branch, but they also all passed before, otherwise I would not have opened the Pr.
Im a bit surprised, that they fail on the CI machine and apparently on yours as well. My guess would be that this issue is then probably due to a difference in the setups? Im using gcc-12, since clang does not work, and have the dependencies installed locally and I think marcus does the same.
Not quite sure how we proceed from here, since I can not reproduce the failure on my machine and the error message in the CI log is rather useless. Can I or somebody else ssh into the CI machine to debug this issue there locally? Not quite sure how the CI is setup here on github / for daphne specifically, I have only managed a CI server in gitlab before.

Cheers, Constantin

pdamme:
**Thanks for addressing all three optional changes**, @CPestka. (I quickly fixed two typos that caught my eye.)

**The justification for C++20 sounds reasonable.** @corepointer: Do you see any potential problem of switching to C++20 related to any existing deployments of DAPHNE you are aware of (e.g., on RedHat)? Maybe we could postpone the swtich to C++20 to after the upcoming v0.3 release, to avoid any unexpected side effects. Switching later should not hurt this PR.

**Interesting to see that the tests pass in your setup.** Yes, a difference in setups is a likely reason. The GitHub action for the CI uses a container image for the environment. You can easily run it on your local machine by executing the following commands from the DAPHNE root directory. This will make your host system's DAPHNE root directory available in `/daphne` inside the container such that you can build DAPHNE and run the tests in the same OS/software environment as the CI (there may still be differences in hardware, though). You may need to use `sudo` for docker.

*On the host:*
```bash
docker pull daphneeu/github-action
docker run -it --rm -w /daphne -v $(pwd):/daphne daphneeu/github-action:latest bash
```

*In the container:*
```bash
# to avoid cmake complaints
./build.sh --clean -y
# build the test cases (effectively all targets) using prebuilt dependencies
./build.sh --no-deps --target run_tests
# run the tests and print which test is running
./test.sh -nb -d yes
```

You could then do anything required to debug the problem inside the CI container. If you need any additional tools inside the container, you may first need to install them there, since it's a minimal environment.

philipportner:
> So this seems to be, imo, a bug in gcc-9.

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=99006
This is indeed a bug in `gcc` and was fixed with `gcc` version `9.5`, `10.4`, and `11.2`. I think the container uses `9.4`.

> Generally I would like us to be clearer on what compiler and compiler versions are supported. I personally expect, if it is not stated otherwise, that any c++ compiler that supports the used c++ version (besides actual compiler bugs) to work, but at the very least both clang and gcc. However, for me e.g. clang did not work at all (I think i tried 10, 12 and 17), which is e.g. not mentioned in the getting started section (there is only a clang version listed, which imo implies that clang actually **is** supported).

I agree that `clang` should also work, I played around with that a bit some time ago, could get it to compile but ran into runtime errors which I didn't debug at the time. I think I'll take another look after the release.
I think `clang` is part of the documentation because it is actually used as the compiler for LLVM (see [here](https://github.com/daphne-eu/daphne/blob/69d5eddedfe591fe9498fca963f956495b9415b5/build.sh#L965-L973)).

CPestka:
And by the way I can not reproduce the other error that is happening in the ci machine now. Now for me all tests pass, also in the container.
Screenshot from 2024-05-24 14-38-42

philipportner:
> And by the way I can not reproduce the other error that is happening in the ci machine now. Now for me all tests pass, also in the container.

Yes, we do have a problem there which I'm currently looking into. See #729

pdamme (approved):
Thanks for replacing `make_shared` to make the tests pass even when compiled with gcc-9 inside the CI container. This PR is **ready to be merged** now. I will squash all commits of this PR into one. Please let me know if @MarcusParadies shall be a co-author of the commit, otherwise it will be just @CPestka.

Regarding the other memory corruptions that you encountered, it would be great if you could report them as issues.

Even though we stay at C++17 for this PR, feel free to use C++20 for your follow-up PRs. We can do the upgrade soon after the upcoming v0.3 release (which could be overlapped with potential discussions on those PRs). In that context, we should also upgrade some dependencies as well as the default OS and compiler version, in my opinion. Support for multiple OSs/versions and compilers/versions woud be awesome; we would need to make checks for that part of the CI.

CPestka:
> I assume we can attribute the failure of the most recent CI run to https://github.com/daphne-eu/daphne/issues/729. The previous run succeeded and on my machine it also works (even in the CI container). So I would merge it nevertheless.

At least it is the same error msg, so i think that is most likely the case.

> Regarding the other memory corruptions that you encountered, it would be great if you could report them as issues.

What i meant there was, that the compiler bug caused a lot of different, non-deterministic memory bugs (maybe due to the testing being multithreaded or alsr being enabled or smth else) and not a single straightforward segfault. These went away with the compiler bug, so no further issues to report.

> Please let me know if @MarcusParadies shall be a co-author of the commit, otherwise it will be just @CPestka.

Yeah, pls add marcus.

pdamme:
Thanks again for your first contribution to the DAPHNE code base, @CPestka; plus thanks to @MarcusParadies. Looking forward to the follow-up PRs mentioned above!

PR #695, opened 2024-04-19
[DAPHNE-#687] OneHot bound check

### LINKED ISSUE #687: OneHot kernel does not check bounds
The OneHot kernel at [src/runtime/local/kernels/OneHot.h](https://github.com/daphne-eu/daphne/blob/main/src/runtime/local/kernels/OneHot.h) currently does not check for bounds when encoding a row (l. 96).  
Given a big enough value in `arg` this leads to overwriting memory outside of the DenseMatrix values array. An example of this is
```
infoMat = [2];
argMat = [10];

print(infoMat);
print(argMat);
print(oneHot(argMat, infoMat));
```
The index this writes to is `10` whereas the result matrix has shape `1x2`. This often crashes during execution too.

Should an added bounds check simply ignore the value and log an error or throw an error instead?

AlexRTer:
This PR fixes #687 by adding a simple bound check for the `cArg` value used for the OneHot encoding. Previous asserts have been replaced with runtime errors and are tested in `OneHotTest.cpp`. The loop setting all values within an encoded segment to zero is now also a single `memset` instruction.

Onehot now also accepts `0` as an argument for `info`, which signals to omit the corresponding column instead of encoding or copying it to `res`. Currently, if all values in `info` are `0` a runtime error is thrown. This could be changed to return an empty `res` instead.

### DIFF
--- a/doc/DaphneDSL/Builtins.md
+++ b/doc/DaphneDSL/Builtins.md
@@ -550,9 +550,10 @@ These must be provided in a separate [`.meta`-file](/doc/FileMetaDataFormat.md).
     The *(1 x m)* row-matrix `info` specifies the details (in the following, *d[j]* is short for `info[0, j]`):
 
     - If *d[j]* == -1, then the *j*-th column of `arg` will remain as it is.
-    - If *d[j]* >= 0, then the *j*-th column of `arg` will be encoded.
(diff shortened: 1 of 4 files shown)


corepointer:
LGTM - thx for fixing this.

I'd merge this later today if no further discussion comes up.

PR #701, opened 2024-04-25
Address Wrong results using Integer Matmul Kernels after Matrix resizing

### LINKED ISSUE #700: Wrong results using Integer Matmul Kernels after Matrix resizing
The results of Matrix multiplications for integer value types are wrong in case that the Matrix has been resized before.
```
N = 3;
EWs = seq(1, N + 1, 1);
A = diagMatrix(EWs);
A = A[:N, :N];
B = A;
v = fill(1, N, 1);
print(A);

print(A@B);

print(A@t(B));

print(A@v);
```

```
DenseMatrix(3x3, int64_t)
1 0 0
0 2 0
0 0 3
DenseMatrix(3x3, int64_t)
1 0 0
0 0 0
0 0 0
DenseMatrix(3x3, int64_t)
1 0 0
0 4 0
0 0 0
DenseMatrix(3x1, int64_t)
1
2
0
```

## Likely Explanation
Integer MatMul kernels are generated using the Eigen library. To use the library the Daphne Matrices are transformed into Eigen Matrices, assuming that the row offset in memory is the row length. When Daphne Matrices are resized they are not necessarily moved in memory, so the row offset does not equal the row length.

resting-dove:
Fixes #700 
#700 mentions an error in the integer MatMul kernels. This PR gives the following correct result for the mentioned example:
```
DenseMatrix(3x3, int64_t)
1 0 0
0 2 0
0 0 3
DenseMatrix(3x3, int64_t)
1 0 0
0 4 0
0 0 9
DenseMatrix(3x3, int64_t)
1 0 0
0 4 0
0 0 9
DenseMatrix(3x1, int64_t)
1
2
3
```

### DIFF
--- a/src/runtime/local/kernels/MatMul.cpp
+++ b/src/runtime/local/kernels/MatMul.cpp
@@ -18,6 +18,8 @@
 
 #include <cblas.h>
 #include <Eigen/Dense>
+#include <Eigen/src/Core/util/Constants.h>
+#include <cstdint>
 #include <spdlog/spdlog.h>
 
 // ****************************************************************************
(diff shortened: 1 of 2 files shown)


resting-dove:
In [c932d25](https://github.com/daphne-eu/daphne/pull/701/commits/c932d2573d5dfa77990afd5c7f8f0b050d48d08d) I added a regression test for the wrong result in the MatMul operation. This revealed that the error also occurs under some circumstances for floating point value types. Specifically when vectors are sliced out of a matrix and then vector vector multiplied.

The later commit [6bc8d14](https://github.com/daphne-eu/daphne/pull/701/commits/6bc8d14a0ecb3655a750300a25e92df8df97f54c) applies clangd formatting to the areas that I changed in MatMul.cpp. It seems that this is not the formatter that was used on this file previously. Please feel free to let me know what formatter is used, or simply go without that commit.

corepointer:
Thx for detecting, reporting and fixing it right away. Good catch :+1: 

I'm not a fan of reviewing PRs where the formatter's gone haywire as it makes spotting actual relevant changes hard. But as we don't have any strict requirements there (yet?) and it occasionally happens to #me too, I won't complain.

I'll merge it later if the tests run without issue.

resting-dove:
Thanks for taking a look at this!
I'm sorry it didn't clear all the tests but I'm glad it caught that there's still some issues after my PR. 

The rowskip is being ignored if the vectors are not views, which is wrong as far as I understand it --> Change to same behaviour whether vectors are views or not.
Also the the rhs of gemv is a vector, so it gets the skiprow logic from the dot product.

However, now some of the algorithms are breaking but I cannot determine how I have caused that or how to fix it. 
I might have some more time to investigate next week but for now I apologise for having to leave this PR in such an unfinished state for now.

corepointer:
Considering whether it is a view matrix or not is important because in a vectorized pipeline, input might seem like a vector but has the memory layout of a matrix. Interestingly, the tests seem to run fine now.

resting-dove:
Thanks for the clarification! I think also when vectors are just views of larger matrices the skip row should be applied in case the vector is transposed as is done after this PR.

The tests are all passing on my system as well now. I had run bin/run_tests directly before and the daphnelib tests failed since I hadn't set the global variables.

corepointer:
LGTM - thx for fixing this @resting-dove. I have squashed the commits and removed the one with the clangd formatting changes as these changes were distracting and not so ideal imho (2 space indent instead of 4, etc). 
All tests ran fine locally and with GitHub Actions.

corepointer:
Btw you might want to fix your git config as this and the previous commit (911c72276dc9ca3aed67e4489518693c904af09b) use different name/email settings.

PR #702, opened 2024-04-26
[DAPHNE-#702] Config parameter for sparsity threshold

corepointer:
This commit adds a configuration item to DaphneUserConfig to let the user specify a sparsity threshold. **This is arbitrarily set to 25% by default (from the non-configurable previous default of 10%).** Furthermore, all the required changes to read the threshold (from json) to passing it down to the SelectMatrixRepresentationsPass are implemented. Two more misc comments:
* Added the config param explain_select_matrix_repr to the default UserConfig.json
* Removed createSelectMatrixRepresentationPass from Passes.td as passing arguments is not supported by TableGen (as I found out here [1])

[1](https://github.com/llvm/llvm-project/issues/61301)

Closes #702


pdamme (approved):
Thanks for this contribution, @corepointer. **It looks good to me.** Allowing users to set the sparsity threshold is a good idea and surely useful for now. (At some point in time it would be even better to select the physical matrix representation in a way that minimizes the physical size or estimated operation runtime.)

*I removed two things that I think are unrelated to this contribution:* (1) additional logging config, and (2) unused logger in `SelectMatrixRepresentationsPass`. Please double-check and merge this PR if you agree with these changes (no need to make me a co-author just for the clean-up).

**Suggestions for the commit message:**
1. You refer to issue 702, but that is the PR number. To my mind, we don't need a separate issue number for this contribution.
2. Please remove the `daphne-eu` in the issue number (or the entire issue number).

corepointer:
Thx for the review @pdamme!

* I squashed your cleanups into the main commit
* For commits that go into main via a PR and do not have a separate GH issue number I use the PR number to reference the discussion. This should be fine because issue and PR number share the same sequence. 
* I did not add the "daphne-eu". In the plain text commit message it's just "DAPHNE- # 702" (spaces added so it is not rendered to fancy format here). Maybe it's just a GH display issue?

Issue #617, opened 2023-09-19
Connected Components fails when reading an input matrix.


aristotelis96:
**Setup**

`scripts/algorithms/components.daph` generates a random matrix and runs the connected components algorithm. 

Instead of generating a matrix, we can also use an existing matrix which can be read using the `readMatrix` operation. We only need to comment-out the random-generated input and use the `readMatrix` op: 
```
# read sparse graph 
G = readMatrix($G);
# n = $n;
# e = $e;
# UT = upperTri(rand(n, n, 1.0, 1.0, 2.0*e/n^2.0, -1), false, false);
# G = UT + t(UT);
```
**Example**

```
./bin/daphne scripts/algorithms/components.daph G=\"amazon.mtx\" 
``` 
- Any input dataset will do to reproduce the error.

**Issue**

However this causes daphne to crash with seg fault:
```
daphne: /home/avontzalidis/prototype/thirdparty/installed/include/mlir/Pass/Pass.h:169: mlir::detail::PassExecutionState& mlir::Pass::getPassState(): Assertion `passState && "pass state was never initialized"' failed.
Segmentation fault (core dumped)
```

By re-running multiple times I was able to get a more helpful error (once every ~50 tries):
```
daphne: /home/avontzalidis/prototype/thirdparty/installed/include/mlir/Pass/Pass.h:169: mlir::detail::PassExecutionState& mlir::Pass::getPassState(): Assertion `passState && "pass state was never initialized"' failed.
[error]: Exception in /home/avontzalidis/prototype/src/compiler/inference/InferencePass.cpp:165: 
type inference returned an unknown result type for some op, but partial inference is not allowed at this point: scf.while
[error]: Got an abort signal from the execution engine. Most likely an exception in a shared library. Check logs!
Segmentation fault (core dumped)
```

**Notes**

1. We should have an additional test case for the `readMatrix` version of connected components, when this is fixed.
2. Before commit 2fc8a9ee572be441ffc72528bb1ec20f7ee7f478 this used to work.
3. A quick fix that makes the script work again is to replace the while loop with a for loop. Of course this means you are not able to terminate the execution when `diff==0` and instead you need to iterate over a specified amount of iterations:
```
# while( as.si64(diff > 0.0) && (maxi==0 || iter<=maxi) ) {
for (iter in 1:maxi) {
    ...
    # iter = iter + 1; # Remove this
}
```

MarcusParadies:
[...]

PR #703, opened 2024-04-26
[DAPHNE-#703] Fix whileOp type inference

corepointer:
This patch resolves whileOp's region arguments that might have been set to unknown and are not updated anymore during type inference.

Closes #703

### DIFF
--- a/src/compiler/inference/InferencePass.cpp
+++ b/src/compiler/inference/InferencePass.cpp
@@ -205,7 +205,7 @@ class InferencePass : public PassWrapper<InferencePass, OperationPass<func::Func
                             " shapes, but the op has " +
                             std::to_string(numRes) + " results");
                 }
-                // Set the infered shapes on all results of this operation.
+                // Set the inferred shapes on all results of this operation.
                 for(size_t i = 0 ; i < numRes ; i++) {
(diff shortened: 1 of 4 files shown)


pdamme (changes requested):
Dear @corepointer, thanks for working on this. Before I review the code, let me kindly ask you to (1) explain what's the problem (ideally by adding a script-level test case that fails before this contribution, or otherwise exhibits some unwanted behavior), and (2) to tidy up the code a bit (e.g., fix indentation mistakes to not clutter the diff, tidy up commented out code). That would make it easier :) .

corepointer:
Yes you're right @pdamme, this one needs cleanup :speak_no_evil: The problem that this PR solves is documented in #617 Similar to the conv2d+read issue that I fixed recently by adding the type inference for conv2d, the daphne.unknown type is propagated from early stages while still parsing and is not properly handled in later compiler passes. Adding a testcase using components_read.daph should test the issue. I'llFixit©

corepointer:
This problem seems to have been fixed since 9a3aaf064a992a423b5b2bf3cfc257387b2cc9a5 so I'll retract this PR. It seems that DAPHNE can now better handle unknown data types. I'll keep the branch around if we need to revisit the whileOp data type passing.

Issue #555, opened 2023-06-28
Exception handling on ARM


corepointer:
Currently DAPHNE source code throws exceptions across "binary boundaries" which is said to be bad practice and apparently not supported on ARM (there, abort() is called due to the uncaught exceptions). This means that, for example, exceptions in a shared library like libAllKernels.so need to be caught in that shared library as they do not get passed on to the main executables like daphne and run_tests. Seems to also be a problem with static linking as the problem already showed up in the MLIR compilation stages (some SQL test case that is supposed to throw an exception and return StatusCode::PASS_ERROR revealed the problem). 

I fixed the problem and will push after cleanup and further testing. The solution concerning shared libs involves the use of signals which might lead us to Linux or at least POSIX dependence. But that is fine for now I guess. If some day we need a port to a non-POSIX system (e.g., Windows?), a solution involving their API can be implemented.

pdamme:
Thanks for bringing up this point, @corepointer. I wasn't aware of this problem (it doesn't seem to occur in the environments where I have used DAPHNE so far), but would be interested in your solution, since I can imagine it could have quite some impact on the way we handle exceptions/errors in DAPHNE.

PR #706, opened 2024-04-29
Improve error handling during compilation + runtime

philipportner:
This PR aims to improve error handling during compilation and during runtime.

The general structure of the error message displayed to the user is the following:
```
 [error]: HEADER [ MSG ]
        | Source file -> FILE:LINE:COL
        |
   LINE | SOURCE
        | ^~~
        |
        |

```
The **HEADER** should contain information about the origin of the exception, e.g., the responsible pass or kernel symbol.
The **MSG** to the user should explain what failed, and, when possible, why the error occurred.
The **FILE**, **LINE**, **COL**, **SOURCE** will display related information from the DaphneDSL source file.
In addition to the line referenced by the location information, the two subsequent lines are also printed.

This mapping from error to source location is now also possible during execution of our precompiled-kernels.
An example:
```
$ daphne source.daphne 
[error]: Execution error: The kernel-function _filterRow__DenseMatrix_int64_t__DenseMatrix_int64_t__DenseMatrix_int64_t failed during runtime with the following message [ sel must have exactly one entry (row) for each row in arg ]
    | Source file -> "/home/user/daphne/source.daphne":3:9
    |
  3 | print(X[[[1, 1, 1], ]]); # 3 given, 4 expected
    |          ^~~
    | Y = fill(123, 4, 3);
    | print(Y[[[1, 1, 1, 1], ]]);

```

Mapping to source location during compilation is done via the `mlir::Location` information available.
To provide this mapping during runtime, we now store an identifier for each dispatch to a kernel. These `KDMInfo` objects contain the symbol name of the precompiled kernel, and file, line, col information. `KDMInfo` instances are stored in the `KernelDispatchMapping` which is now part of the `DaphneContext`. 
Kernel dispatches now also pass an `int` identifier to each kernel, allowing to retrieve the correct `KDMInfo` for a certain kernel. With that, we have a mapping for each physical kernel to the source position in the DaphneDSL script. This is currently omitted for vectorized and distributed ops.


pdamme:
Awesome, thanks for this contribution @philipportner! Improved error handling/messages are a highly required feature when writing more complex DaphneDSL scripts. I will look over your code in the next few days.

philipportner:
closes #707

pdamme (changes requested):
Thanks, @philipportner for this **highly valuable contribution on improving error handling and error messages** in DAPHNE! This **will make writing complex DaphneDSL scripts much more productive**.

It's a great step toward a unified appearance of error messages (from anywhere in the system: parser, compiler, runtime, ...), with very informative error output including the DaphneDSL file/line/col location and source snippet. It's also a great step toward unified error handling (replaces assertions by exceptions in many places, replaces MLIR-based error passing by exceptions). I like the design of the `KernelDispatchMapping` and how it's connected to the kernels. Thanks also for the high amount of helpful source code comments and updates to the developer docs.

**I read the code and tried it out** with various failure script-level test cases from `test/api/cli/`. In that context, **I still detected some issues**. Sorry for the very long feedback, but error handling and error messages are quite important and affect almost all corners of the code base and are directly visible to users. *However, let's don't blow up this PR too much, let's focus on the critical points and anything else can be added afterward.*

**Issues that must be addressed/discussed before merging**
1. There were several places in the code where exceptions were caught, printed through the logger, but not re-thrown, since they would otherwise cross library/binary boundaries. @corepointer ensured that a while ago, since throwing exceptions across these boundaries can cause problems, especially in the ARM setup he was working with, if I remember correctly. In this PR, you replaced many/all of these prints with re-throws. Did yout take the problem of throwing exceptions beyond library/binary boundaries into account? Occurs at least in: `DaphneInferTypesOpInterface.cpp`, `RewriteToCallKernelOpPass.cpp`, `InferencePass.cpp`, `SQLParser.cpp`.
1. There are a few places that simply do `throw;` without any exception object. Please clarify how that fits into the overall error handling process. Occurs at least in: `DaphneIrExecutor.cpp`, `SpecializedGenericFunctionsPass.cpp`.
1. Please undo the replacement of `static_assert`s with exceptions. Unlike exceptions, `static_assert`s are evaluated at C++ compile-time. Thus, certain errors (e.g., based on the type combination a kernel is instantiated with) can already be reported at C++ compile-time. We should not wait until run-time if we know of the problem earlier. Occurs at least in: `RandMatrix.h` (2x), `Sample.h` (1x).

**Optional issues can be addressed in this PR or later**
1. Please fix printing the previous/next lines in the DaphneDSL script.
   - Try `bin/daphne test/api/cli/import/import_failure_1.daphne` and compare the output to the DaphneDSL script.
   - The previous source line is not shown, the last line before EOF gets repeated.
   - ```text
     [error]: While parsing: DSLVisitor failed with the following message [ You cannot import the same file twice: /daphne/test/api/cli/import/sandbox/b.daphne ]
        | Source file -> "/daphne/test/api/cli/import/import_failure_1.daphne":3:0
        | 
        3 | import "sandbox/b.daphne" as "something";
        | ^~~
        | import "sandbox/b.daphne" as "something";
        | import "sandbox/b.daphne" as "something";
     ```
1. It seems to me like the first kernel call that is registered in the `KernelDispatchMapping` gets the id `0`. However, `0` is used in `ErrorHandler::runtimeError()` to check for a valid kernel id. The first kernel call is probably something like `createDaphneContext`, which is unlikely to fail. Nevertheless, the collision should be avoided. I recommend using `-1` to indicate "unknown", just like we do in other places (especially in property inference).
1. In `genKernelInst.py`: Please fix the indentation of the C++ code generated for error handling (have a look at the generated `kernels.cpp`).
1. There are still *numerous* assertions in the code base. I'm not sure if you intended to replace them all by exceptions in this PR. Anyway, we should replace them all at some point in time.
1. You changed many error messages thrown in kernels to contain the operation mnemonic. That surprised me since one of the contributions of this PR is actually to add such information automatically. For instance, the kernel function name is added automatically, the same could be done for the operation mnemonic (which is currently included in the kernel names anyway). I recommend removing the operation mnemonics from the very error messages. Occurs at least in `ColBind.h`, `EigenCal.h`, `Gemv.h`, `MatMul.cpp`, `MatMul.h`, `Now.h`, `OneHot.h`, `Quantize.h`, `RandMatrix.h`, `Replace.h`.
1. In some cases, error messages still deviate from the template.
   - *Lexer/parser error from ANTLR*
     - Example: `bin/daphne test/api/cli/syntax/general_4.daphne`
     - ```text
       [error]: While parsing: test/api/cli/syntax/general_4.daphne:1:11 extraneous input ';' expecting {',', ')'}
       ```
   - *Some compiler errors*
     - Example 1: `bin/daphne test/api/cli/sql/basic_pass_failure_2.daphne`
       - ```text
         [error]: Lowering pipeline error.
         RewriteSqlOpPass
         at SQLParser66
         at SQLVisitor (visitSelect)
         at Error during From statement. Couldn't create Frame.
          SQLVisitor (visitTableIdentifierExpr)
         at Error during From statement. Couldn't find Frame.
         PassManager failed module lowering, responsible IR written to module_fail.log.
         ```
       - Looks quite different.
     - Example 2: Use the following DaphneDSL script
       ```r
       def bar(y:si64) {print(y);}
       def foo(x) {bar(x);}
       foo(1.1);
       ```
       - ```text
         loc("my-example.daphne":6:4): error: 'llvm.call' op operand type mismatch for operand 0: 'f64' != 'i64'
         "builtin.module"() ({
         …
         }) : () -> ()
         loc("my-example.daphne":0:0): error: module pass error
         ```
     - Example 3: `bin/daphne test/api/cli/secondorder/map_failure_5.daphne` (I've just added this test case on main)
       - ```text
         [error]: Lowering pipeline error.
         SpecializeGenericFunctionsPass failed with ...
         ```
       - The line break after "Lowering pipeline error." is unexpected.
1. The two commits in this PR do make sense individually, so we could *"rebase & merge"* then (not *"squash & merge"*). If that is what you want, please: (a) replace the "feat: " from the commit titles, since we don't use this specific prefix, and (b) fix the typo: "Adaption" to "Adoption" in the title of the second commit.
1. The formatting (indentation, line breaks, white spaces in general) was changed in *numerous* places, even in lines that have no other changes. This unnecessarily blows up the diff. I know we don't have a clear code style yet, but I would always argue to keep formatting as it is in lines where nothing is really changed. Undoing all this would be a waste of time now, but maybe keep it in mind for next time :) .

**Recommended follow-up issues (don't include in this PR anymore)**
1. Make colored error messages optional. They are really great for users on the console, but
   - they are problematic when redirecting to text file (cluttered with control sequences)
   - they are not good for catch2 test case output (catch2 uses it's own formatting (e.g., bold yellow), which gets turned off by ours in the middle; this makes reading the messages harder)
1. Make the very error messages consistent, e.g
   - start with lower-case letter (unless first word is an upper-case identifier)
   - no dot or other punctuation at the end
   - consistently include line number (`__LINE__`) etc. or not
   - remove information that is anyway represented in the error output in another way (e.g., operation/kernel name)
1. I like the concept of the "action" in the error output (roughly the system component that failed). However, currently the level of detail varies quite a bit. Sometimes it's just a high-level component name, sometimes it's the concrete function name. Maybe we could unify that a bit. When we include the function name, instead of hardcoding it (which is error-prone in the context of refactoring), we could use some pre-compiler macro that resolves to the function name.
1. If I'm not mistaken, all errors thrown in the SQL parser will show only the location of the `SqlOp`, not the location inside the SQL query.
1. Support `VectorizedPipelineOp` and `DistributedPipelineOp` for error handling (another kind of location information) and remove the work-around in `genKernelInst.py`.

**Feel free to share your thoughts on these comments.**

philipportner:
> **Issues that must be addressed/discussed before merging** 
> 1. There were several places in the code where exceptions were caught, printed through the logger, but not re-thrown, since they would otherwise cross library/binary boundaries. @corepointer ensured that a while ago, since throwing exceptions across these boundaries can cause problems, especially in the ARM setup he was working with, if I remember correctly. In this PR, you replaced many/all of these prints with re-throws. Did yout take the problem of throwing exceptions beyond library/binary boundaries into account? Occurs at least in: `DaphneInferTypesOpInterface.cpp`, `RewriteToCallKernelOpPass.cpp`, `InferencePass.cpp`, `SQLParser.cpp`.

As all our internal libraries are compiled by the same compiler, we can be sure that all of them follow the same ABI. When no handler can be found during stack unwinding `std::terminate` will be called, which calls `abort`. For that, we have a signal handler in `daphne_internal.cpp`.

> 2. There are a few places that simply do `throw;` without any exception object. Please clarify how that fits into the overall error handling process. Occurs at least in: `DaphneIrExecutor.cpp`, `SpecializedGenericFunctionsPass.cpp`.

The standalone `throw;` would rethrow the current exception as is, without creating a new one. That's fine in `DaphneIrExecutor.cpp`, arguably also fine in `SpecializedGenericFunctionsPass.cpp` or anywhere else, if one does not want to add additional information in the specific `catch` block. I'll add a 'rethrowError' here so we'll have more info about the call stack.

> 3. Please undo the replacement of `static_assert`s with exceptions. Unlike exceptions, `static_assert`s are evaluated at C++ compile-time. Thus, certain errors (e.g., based on the type combination a kernel is instantiated with) can already be reported at C++ compile-time. We should not wait until run-time if we know of the problem earlier. Occurs at least in: `RandMatrix.h` (2x), `Sample.h` (1x).

Good catch, I was under the impression that the same problem mentioned in #707 for `assert` also applies to `static_assert`.
 
> **Optional issues can be addressed in this PR or later**
> **Recommended follow-up issues (don't include in this PR anymore)**

I'll respond to these tomorrow where necessary, thought I read through them and generally agree. 

Thanks a lot for the detailed review @pdamme !

philipportner:
Just tried it again on ARM, it does catch the `SIGABRT` as expected, and even prints the message we throw because it displays the output of `what()` of the exception.

pdamme:
Thanks for the response, @philipportner.

**Regarding the "issues that must be addressed/discussed before merging"**:
1. Thanks for the clarification. In our meeting today, you also mentioned that you've successfully tried it on ARM and the exception crossed library boundaries and it's message was displayed. To me this sounds plausible, but I'm not an expert in this matter. Unless @corepointer (who dealt with the exception/signal handling on ARM in detail, see #555) has any further comments on it, I'm fine with changing it.
2. Okay, that was my bad, wasn't aware that `throw;` rethrows the exception currently being handled...
3. Thanks. I've just double-checked it, is `static_assert` is effective even in release builds.

philipportner:
>        def bar(y:si64) {print(y);}
>        def foo(x) {bar(x);}
>        foo(1.1);

>          loc("my-example.daphne":6:4): error: 'llvm.call' op operand type mismatch for operand 0: 'f64' != 'i64'
>          "builtin.module"() ({
>          …
>          }) : () -> ()
>          loc("my-example.daphne":0:0): error: module pass error

That's because there is no exception generated but rather an MLIR internal pass signals pass failure with `emitOpError()` at `mlir/lib/Dialect/LLVMIR/IR/LLVMDialect.cpp:1243`. As our handling is not connected to the diagnostics engine of MLIR there's no way to catch this at the moment. Let's see if we can integrate the diagnostics engine in the future.

> 8. The formatting (indentation, line breaks, white spaces in general) was changed in _numerous_ places, even in lines that have no other changes. This unnecessarily blows up the diff.

FYI: this can be ignored by checking the `Hide whitespace` checkbox when viewing a PR changelist. Let's work towards consistent formatting, I'll create a PR for that.

I addressed most of your points, for the rest I'll create separate issues. Thanks again for the review.

corepointer:
Too bad that I'm a bit late to the discussion on this PR (my bad as I've been mentioned several times). Two items of (constructive) criticism:
* The new error handler bypasses the logger and therefore output can not be configured/controlled anymore.
* Please relate this discussion to the commits by adding something like [DAPHNE-#706] to the commit message.
Other than that it's great to see that we have more meaningful error messages now :+1:

corepointer:
> The new error handler bypasses the logger and therefore output can not be configured/controlled anymore.

The constructive version of that: Please use the logging facility to make output controllable via the configuration :)

philipportner:
> > The new error handler bypasses the logger and therefore output can not be configured/controlled anymore.
> 
> The constructive version of that: Please use the logging facility to make output controllable via the configuration :)

I'm not really sure what you mean by that because I didn't change that for almost all the locations in this PR, there's only a few places that had an `spdlog` before that I changed, everything else was also done through exceptions already.
The message that's finally output to the user still goes through one of the `spdlog` in `daphne_internal.cpp`. So the output to the user should be controllable through the `spdlog` configuration.

corepointer:
Maybe I missed something there :see_no_evil: But isn't the file output done manually? That can be spdlogged as well. And a pro tip: Spdlog pulled in the fmt library as a dependency. String formatting can be conveniently done with that :muscle:

PR #725, opened 2024-05-10
[DAPHNE-#531] Simplified GettingStarted.md.

### LINKED ISSUE #531: Simplify GettingStarted.md
Our instructions for getting started with DAPHNE (`doc/GettingStarted.md`) are meant to make it easy for new users and developers to set up a development environment. However, I think this document has grown over time and has by now become too complicated. We should simplify it.

These are some concrete ideas:
- Initially, we wanted to make it easy to set up DAPHNE natively on a system, but that has proven to be quite difficult. Therefore, we started using containers. However, the GettingStarted still reads like one should install DAPHNE natively. The container approach must be presented much more prominently (and first of all, the containers must work out of the box, see #530)
- The document contains a lot of details on various aspects, many of which should be moved to individual documents (e.g., some of the details on containers, overview of the code base, how to pull, how to execute a DaphneDSL script, ...)

pdamme:
**Commit message:**
- The guidelines on getting started (for users and developers) have grown over time and were not simple anymore.
- Initially, we had a focus on native builds of DAPHNE; hence, the document starts by listing the dependencies etc.
- However, meanwhile, the Docker containers have become the de-facto and simplest way to get started.
- This commit simplifies getting started by prepending quickstart instructions for both users and developers to the beginning of the document, such that these are immediately found.
- These are slightly simplified instructions that should work in most cases.
- For all more complex cases, the original contents of the file is still present (with minor fixes), but under the headline of "Additional Details for Custom Setups".
- The important information from Quickstart.md was merged into GettingStarted.md; thus, Quickstart.md could be removed (for a less fragmented documentation).
- Furthermore, the example `hello-world.daph` was simplified to really only print "Hello World!"; everything else in this script was rather useless from today's point of view.
- Closes #531.

**I invite everyone to share their comments on these changes.** If there are no comments, I will merge it on Fri May 17.


pdamme:
I will merge this PR now. If any comments arise later, we can always change the documentation in follow-up commits.

Issue #729, opened 2024-05-17
Codegen test fails non-deterministically in GitHub action


pdamme:
The test suite contains a test case on codegen (`test/codegen/CodegenTest.cpp`). Currently, this test case sometimes succeeds and sometimes fails when run during the GitHub action. For instance, in PR #728, all test cases passed in the GitHub action. After putting the single commit in the PR on the main branch (rebase & merge resulting in 4eb07576b58e854ef096877722f0ebbc7339b5c4, no additional changes at all), the test case fails in the CI. The problem is also affecting PR #675 at the moment, but I guess it could randomly happen in any PR or commit on main at the moment.

On my local machine, I cannot reproduce this issue. It may be related to multi-threading or any other source of non-determinism that is different in the CI environment.

There have been problems with the codegen test case before, when it got stuck non-deterministically (reported in the context PR #675) and fixed in c2900a39189a8e7e407082278a33f6ef3981ff76.

It is not clear if the problem is caused by the codegen test or if it only becomes effective there.

philipportner:
Looks like we finally got logs of an error.

https://github.com/daphne-eu/daphne/actions/runs/9701687915/job/26775861276

On what CPU is the CI executed? Is it the same HW every time?
Could be that for certain CPUs the tiling logic  simply does not achieve the tiling required from the description when using `FileCheck` for the `matmul_tile.mlir` test case.

philipportner:
Still happening, https://github.com/daphne-eu/daphne/actions/runs/11124869078

PR #733, opened 2024-05-24
Non-deterministic test fail during Github CI action

philipportner:
TODO: for now just to trigger CI action and produce debug output (#729)


philipportner:
This is merged for now to get actual debug output once the CI run fails again. Debug output will be removed once we figure out the problem and close #729

PR #734, opened 2024-05-28
Dnn ops

corepointer:
This is a dirty and buggy snapshot of the progress on integrating all the needed ops for DNN.

Also contains fixes for some of the bugs :)

* CUDA version of DNN ops mostly implemented (some backward passes missing)
* CUDA convolution keeps crashing :-/
* Script level alternatives for the operations are there to replace the CUDA ops
* Pooling script level operation is crashing :-/
* Contains some implementations for shape/type inference but this is not in a solid state
* In some locations of the scripts there needed to be a type indicator to make it work. So all these ops rely on <f64> atm, which might not be ideal.
* Contains an example with a LeNet implementation for MNIST character classification ([ported from SystemDS](https://github.com/apache/systemds/blob/main/scripts/nn/examples/mnist_lenet-train.dml))


corepointer:
This PR is now based on #758

PR #741, opened 2024-06-03
Zarr Support

CPestka:
## Zarr Support

Adds basic Read Support for the [v 2 Zarr file format](https://zarr-specs.readthedocs.io/en/latest/v2/v2.0.html). Some features have been omitted for now (groups, filters, compression and some of the "non cpp native" data types, most of which thus also don't have a respective daphne supported type.

This Pr simply uses the stl for i/o, but once #736 has been merged, another Pr will follow that adds an async zarr reader.

This Pr also fixes a couple bugs in the rechunk and dice variants and cleans a couple of other things up in the tensor implementations.

As mentioned in #694 and like #736 this Pr also again switches to c++20, here mostly for std::endian, but also uses it a lot elsewhere for convenience e.g. using enum, make_shared<T[]> :) etc..

### One possible discussion point

This pr includes 2 zarr "files" for testing, which total ~8MB and ~1k files. I don't particularly like dumping a bunch of binary files into a repo. There are already a couple of "io test files" there, but these would be ~ as large as the rest combined and it is just a lot of files. Should we avoid this or is this fine with everybody? The alternative would be to add a writer implementation or a dependency to generate some tmp zarr files on the fly.


CPestka:
Seems like the compiler on the CI machine again does not like the c++20 features....

> Jun 03 15:33:30 /__w/daphne/daphne/src/parser/metadata/ZarrFileMetaDataParser.cpp:85:15: error: expected nested-name-specifier before 'enum'
Jun 03 15:33:30    85 |         using enum ZarrDatatype;

pdamme (changes requested):
Thanks for this PR, @CPestka. Support for reading Zarr files into `ContiguousTensor`/`ChunkedTensor` is highly welcome. I've looked over your code and it **looks good to me**. As I'm not familiar with the details of the Zarr format, I mostly concentrated on the integration into the overall codebase and trust your expertise for the Zarr-specific aspects. Thanks also for the helpful comments in `ReadZarr.h`.

I've only fixed minor issues like a few typos in the comments and identifiers and some missing license headers.

**Required before the merge:**
1. Regarding the two Zarr "files" for the test cases: We should at least make a conscious decision here. It would be great if we could avoid putting so many little binary files that sum up to a few MBs into the repository. As far as I understand, the reason why there are so many chunk files is that you use a tensor with 10x10x10 = 1000 chunks (with 10x10x10 = 1000 fp64 elements = 8 KB each). Couldn't we simply decrease the data size by having less and/or smaller chunks, e.g., by 2x2x2 = 8 chunks of 10x10x10 elements each? Shouldn't that suffice to test the Zarr reader?
2. When I build your code with g++-13.1.0 on Ubuntu 20.04.6, there are still several warnings, most of them from the MLIR code and related to `operator==`. I assume these warnings are not related to your changes, but to the switch to C++20, so we will probably handle them in the commit that switches to C++20. But before the merge, we need to double-check that they are gone from this PR.

**Optional changes** (not required before the merge):
1. `ZarrFileMetaData.cpp`, `operator<<` for `ZarrDatatype` prints `"double"` for `ZarrDatatype::FP32`. Is that intended or should it be `"float"` or `"single"`?
2. `ReadZarrTest.cpp`: In many of the test cases you `memcmp` the data in the reference tensor and the read tensor *and* compare the tensors through `==`. Isn't the `memcmp` redundant as the object comparison includes comparing the data?

**Apart from that, we will merge this PR as soon as we have switched to C++20...**

CPestka:
@pdamme Thx for having a look and fixing my numerous typos :)

> ZarrFileMetaData.cpp, operator<< for ZarrDatatype prints "double" for ZarrDatatype::FP32. Is that intended or should it be "float" or "single"?

y, thats a mistake.

> ReadZarrTest.cpp: In many of the test cases you memcmp the data in the reference tensor and the read tensor and compare the tensors through ==. Isn't the memcmp redundant as the object comparison includes comparing the data?

The == tests more than just the actual memory of the elements, but yes that memcmp is redundant in a couple of the tests, as well as some other checks also done by the == call. I removed the duplicates.

> When I build your code with g++-13.1.0 on Ubuntu 20.04.6, there are still several warnings, most of them from the MLIR code and related to operator==. I assume these warnings are not related to your changes, but to the switch to C++20, so we will probably handle them in the commit that switches to C++20. But before the merge, we need to double-check that they are gone from this PR.

I just tested it with g++-13.2.0 on 24.04 and y it indeed spits out some new warnings, but as far as i can tell these come from mlir and llvm code, which i think are caused by a "missing" const on a operator==(), as well as one warning from antlr and spdlog each. Not sure what you want to do about these e.g. update their version or silence them.

Btw on Ubuntu 24.04 building protobuf fails due to missing libnsl. I guess we should add libnsl-dev to the dependency list for ubuntu 24.04.

> Regarding the two Zarr "files" for the test cases: We should at least make a conscious decision here. It would be great if we could avoid putting so many little binary files that sum up to a few MBs into the repository. As far as I understand, the reason why there are so many chunk files is that you use a tensor with 10x10x10 = 1000 chunks (with 10x10x10 = 1000 fp64 elements = 8 KB each). Couldn't we simply decrease the data size by having less and/or smaller chunks, e.g., by 2x2x2 = 8 chunks of 10x10x10 elements each? Shouldn't that suffice to test the Zarr reader?

Y, that sounds reasonable.

corepointer:
my 2 cents:
* We should find an alternative to having 1000+ files added in test/
* Yes, most of the warnings come from LLVM/MLIR. Now that Ubuntu24 has landed in DAPHNE Docker Land we'll have to deal with it somehow :face_with_spiral_eyes: 
* Would be nice to have this hooked up to the generic read kernel (e.g., "if (zarr file) -> readZarr()")

MarcusParadies:
> my 2 cents:
> 
> * We should find an alternative to having 1000+ files added in test/
> * Yes, most of the warnings come from LLVM/MLIR. Now that Ubuntu24 has landed in DAPHNE Docker Land we'll have to deal with it somehow 😵‍💫
> * Would be nice to have this hooked up to the generic read kernel (e.g., "if (zarr file) -> readZarr()")

We could just use a little python helper to automatically generate the test files on-the-fly.
Regarding the overall integration. we already have a "running" prototype that does the wiring, but requires some more polishing before opening a PR.

PR #743, opened 2024-06-05
daphnelib: improve pyproject.toml and readme

m-birke:
extending docs
naming project to daphne-lib

### DIFF
--- a/src/api/python/README.md
+++ b/src/api/python/README.md
@@ -82,8 +82,8 @@ Use [twine](https://twine.readthedocs.io/en/stable/) for publishing to [PyPI](ht
 
 1. Set `version` in `pyproject.toml`
 1. Build according to __Build__ section
-1. `twine check dist/daphne-<version>-py3-none-any.whl`
+1. `twine check dist/daphne_lib-<version>-py3-none-any.whl`
     - checks the wheel file
-1. `twine upload -u __token__ dist/daphne-<version>-py3-none-any.whl`
+1. `twine upload -u __token__ dist/daphne_lib-<version>-py3-none-any.whl`
     - to publish to PyPI
     - twine prompts for your PyPI token
--- a/src/api/python/README.md
+++ b/src/api/python/README.md
@@ -16,11 +16,46 @@ limitations under the License.
 
 # DaphneLib
 
-Refer to the [online docs](https://daphne-eu.github.io/daphne/) for documentation.
+Refer to the online documentation:
+
+- [DaphneLib](https://github.com/m-birke/daphne/tree/main/scripts/examples/daphnelib)
+- [Container Usage](https://daphne-eu.github.io/daphne/GettingStarted/#2-download-the-daphne-development-container-image)
+- [Building DAPHNE](https://daphne-eu.github.io/daphne/development/BuildingDaphne/)
 
 ## Setup
 
 The environment variable `DAPHNELIB_DIR_PATH` must be set to the directory with `libdaphnelib.so` and `libAllKernels.so` in it.
+
+```sh
+export DAPHNELIB_DIR_PATH='pathto/daphne/lib'
+```
+
+## Usage
+
+More script examples on [github](https://github.com/m-birke/daphne/tree/main/scripts/examples/daphnelib)
+
+```python
+from daphne.context.daphne_context import DaphneContext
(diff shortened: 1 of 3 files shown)


pdamme (approved):
Thanks for this contribution in preparation of making DAPHNE/DaphneLib available on PyPI, @m-birke. That will definitely simplify things for users.

Overall, the **changes look good to me**, but **a few things need clarification to work out of the box**. These things should be improved, but as nobody currently depends on them, we don't need to put them on the critical path for this PR, if you want to merge it quickly.

**Recommended changes:** (can be addressed after merging)

1. I think **the setup instructions should clearly mention that the DAPHNE system must be obtained separately**. Currently that's a bit hidden as seemingly extra information above. Maybe we could write something like:
    1. "Follow the instructions at ... to obtain DAPHNE". Instead of the links into the detailed docs, I recommend pointing just to the [GettingStarted](https://daphne-eu.github.io/daphne/GettingStarted/) (or perhaps to the section on the [Quickstart for users](https://daphne-eu.github.io/daphne/GettingStarted/#quickstart-for-users)) and maybe to the overall [DAPHNE documentation](https://daphne-eu.github.io/daphne/). Othwise, we might over-emphasize certain sub-topics that could lead users astray.
    2. `pip install daphne-lib` (even though that may sound trivial)
    3. Configuration of environment variables etc.
2. I think **the setup instructions should mention the environment variable `LD_LIBRARY_PATH`**, too. When I follow the setup instructions in the container I use for building/running DAPHNE and try to execute the given script, everything is fine. However, when I try it on my host system, I get the following error message: `OSError: libantlr4-runtime.so.4.9.2: cannot open shared object file: No such file or directory`. That can be fixed by `export LD_LIBRARY_PATH=path-to-daphne/lib:path-to-daphne/thirdparty/installed/lib`. Then the script works.
3. It **would be good to mention that one should use the corresponding (=same) version of daphne-lib and DAPHNE**. Otherwise, there could be incompatibilities. For instance, when I try the setup instructions (plus `export LD_LIBRARY_PATH=$DAPHNELIB_DIR_PATH:$LD_LIBRARY_PATH`) with the v0.2 binary release, then the given script yields `ValueError: NULL pointer access`. I didn't investigate that in detail, but I assume that the v0.2 binary release and the current DaphneLib Python code are not compatible anymore.

**Optional changes:** (may not be needed at all)
1. As the package name `daphne` is already taken on PyPI, and we basically decided to use `daphne-lib`: Should we also rename our package to `daphne-lib`/`daphnelib`? I'm mainly thinking about imports in Python, e.g. `from daphne.context.daphne_context import DaphneContext` vs. `from daphnelib.context.daphne_context import DaphneContext`. Otherwise, what would happen if anyone ever wanted to use the unrelated `daphne` package and our DaphneLib in the same Python script?

pdamme (changes requested):
*PS: What I forgot to mention...*

**Required changes:** (need to be addressed before the merge)
1. **Make the DaphneLib test cases pass again.** Currently they fail as TensorFlow prints warnings to stderr that confuse our checks for empty stderr. Actually, these warnings should get silenced by the line `export TF_CPP_MIN_LOG_LEVEL=3` in `test.sh`... (?)

pdamme:
Okay, the DaphneLib test cases also fail in other PRs, so this is probably unrelated to your changes (would have been surprising anyway). I will look into that problem separately.

m-birke:
Thanks @pdamme for the review.

I tried to incorporate all recommended changes.

To the optional changes:

It should not be a problem out of several reasons:

1. The unrelated daphne lib is a webserver (supposed usage as CLI) and probably not used together with daphne-lib
2. If both are used together in an installation, the imports are getting resolved correctly, there only can occur a problem if one does sth. like

```python
import daphne

daphne.func()
v = daphne.MyClass()
```

but it anyway is the proposed (and only possible atm.) method to make full imports

```python
from daphne.... import MyClass
```
3. we can never assure that there is no name collision with another PyPI project, there could be a project named 'Foobarzumsing' which provides a Python package called 'daphnelib'

pdamme (approved):
Thanks for addressing the recommended improvements, @m-birke. Regarding the optional comment on the package name: Thanks for the clarification, fair enough. **This PR is ready to be merged now.**

PR #744, opened 2024-06-05
[DAPHNE-#680] bug fix for signed integer casts

### LINKED ISSUE #680: Incorrect conversion of `true` bool to signed integer
When casting a `true` bool to a signed integer of any length in a Daphne script it is cast to `-1` instead of `1`.

```
print(as.si8(true));     // -1
print(as.si32(true));    // -1
print(as.si64(true));    // -1

// print(as.ui64(true)); // 1
```
The IR immediately after parsing still shows:
```
%1 = "daphne.constant"() {value = true} : () -> i1
%2 = "daphne.cast"(%1) : (i1) -> si8
```
whereas after simplifications (`parsing_simplified`) these have been wrongfully folded to just
```
%2 = "daphne.constant"() {value = -1 : si8} : () -> si8
```

AlexRTer:
This PR aims to fix #680.

When converting to any signed integer type that was longer than the input type during `CastOp` folding, the operand was always sign extended. This resulted in an incorrect conversion for unsigned and signless operands with a set most significant bit. E.g.
```
print(as.si8(true));            // -1
print(as.si32(as.ui8(255)));    // -1
```

Operands that are casted to a signed integer type are now zero extended instead if the input was not a signed integer type as well. There is a small test to guarantee the correct conversion.


philipportner:
Thanks for the fix @AlexRTer !

PR #748, opened 2024-06-06
[DAPHNE-#747] Use test.sh in CI again.

### LINKED ISSUE #747: CI tests fail due to TensorFlow warnings
The CI tests for DaphneLib currently fail because TensorFlow prints warnings to stderr, but the test cases expect that stderr is empty. Actually, the line `export TF_CPP_MIN_LOG_LEVEL=3` in `test.sh` should suppress all such warnings from TensorFlow (the warnings themselves are not the problem, but they should be silenced). 

This problem has started to appear in the last few days (and already shows up in multiple recently updated PRs) even though there were no related changes in DaphneLib. Thus, I suspect it has to do with recent updates to the DAPHNE container images. Maybe 6bc48f29924791099b0326b951cc6d1ccb2c5f8c, which added TensorFlow and PyTorch to the containers.

*Example output from the CI run:*

```
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
run_tests is a Catch v2.13.8 host application.
Run with -? for options

-------------------------------------------------------------------------------
data_transfer_numpy_1.py
-------------------------------------------------------------------------------
/__w/daphne/daphne/test/api/python/DaphneLibTest.cpp:64
...............................................................................

/__w/daphne/daphne/test/api/cli/Utils.h:447: FAILED:
  CHECK( errDaphneLib.str() == "" )
with expansion:
  "2024-06-05 22:36:16.387755: I tensorflow/tsl/cuda/cudart_stub.cc:28] Could
  not find cuda drivers on your machine, GPU will not be used.
  2024-06-05 22:36:16.419796: I tensorflow/core/platform/cpu_feature_guard.cc:
  182] This TensorFlow binary is optimized to use available CPU instructions in
  performance-critical operations.
  To enable the following instructions: AVX2 FMA, in other operations, rebuild
  TensorFlow with the appropriate compiler flags.
  2024-06-05 22:36:17.114345: W tensorflow/compiler/tf2tensorrt/utils/py_utils.
  cc:38] TF-TRT Warning: Could not find TensorRT
  "
  ==
  ""
```

pdamme:
- As a part of our CI workflows, we run the entire DAPHNE test suite.
  - The tests reside in the catch2-based executable run_tests.
  - Normally (e.g., in a local dev environment), we don't run run_tests directly, but invoke it indirectly through test.sh.
  - test.sh offers a few arguments and sets a few environment variables such that the tests can run as expected.
- In the CI workflow, we initially executed test.sh to run the test cases.
- However, in ebfcc3bfe9457fc8106ba87c62c89ba7b2b4f48f, this was changed to invoking run_tests directly (plus adding a few essential environment variables on the command line).
  - The motivation back then was that test.sh re-builds run_tests, but potentially with unwanted parameters (in the worst case, this can result in pulling the LLVM repo).
- In the meantime, test.sh has become more complex, e.g., by setting more environment variables.
  - These changes have not been reflected in the CI workflow (.github/workflows/main.yml).
  - So far, this inconsistency did not lead to any problems.
    - Except, maybe, the one described in 00eec3cb714df336c818f400beb72c990691e9f5.
  - However, recently TensorFlow was added to the CI container image. - That caused the DaphneLib test cases to import TensorFlow (so far, it was not imported, since it is an optional dependency that is only imported if installed). - TensorFlow has the habit of printing various warnings to stderr. - As these warnings confuse the checks in our test cases (for empty stderr), test.sh suppresses these warnings by setting the environment variable "export TF_CPP_MIN_LOG_LEVEL=3"). - As the CI workflow bypasses test.sh, it did not silence the TensorFlow warnings. - As a consequence, all DaphneLib test cases started failing in the CI workload after the recent container image update.
- To solve this problem, we could
  - either: add those additional environment variables before the invocation of run_tests in main.yml - Error-prone code duplication, difficult in case of conditional environment variables (which we do have meanwhile).
  - or: simply use test.sh in the CI workflow again
    - Consistent solution, since environment variables are contained in a single place.
    - The original motivation for swtiching to run_tests in ebfcc3bfe9457fc8106ba87c62c89ba7b2b4f48f is meanwhile gone, since test.sh has an argument --no-build.
- Thus, this commit makes the CI workflow use test.sh instead of run_tests again.
  - Following the comment in the commit message of ebfcc3bfe9457fc8106ba87c62c89ba7b2b4f48f, we don't pipe the output of test.sh to ts now.
- Fixes #747.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -45,7 +45,7 @@ jobs:
       shell: bash
 
     - name: Testing
-      run: LD_LIBRARY_PATH=$PWD/lib:$LD_LIBRARY_PATH PATH=$PWD/bin:/usr/lib/llvm-10/bin:$PATH DAPHNELIB_DIR_PATH=$PWD/lib PYTHONPATH="$PYTHONPATH:$PWD/src/api/python/:/usr/lib/llvm-10/build/utils/lit/" bin/run_tests
+      run: ./test.sh --no-build
 
     - name: "List generated files"
       run: |


pdamme:
I opened this PR mainly to notify everyone interested about this little change to the CI workflow. I will merge it myself as soon as the CI tests have finished, to make the CI tests pass again for the open PRs.

PR #756, opened 2024-06-13
[DAPHNE-#666] optional inc for seq function

### LINKED ISSUE #666: `seq()` DaphneDSL built-in: Make `inc` argument optional
The `seq()` built-in function in DaphneDSL creates an increasing sequence of numbers from a given `start` to a given `end` in increments of a given `inc`. In fact, the increment is very often `1`, so for convenience, it would be great to make the increment an optional argument.

This can and should be done purely at the DSL level (no need for a deeper integration into DaphneDSL or the kernels). The change can be implemented in `src/parser/daphnedsl/DaphneDSLBuiltins.cpp`. It should be accompanied by some script-level test cases.

saminbassiri:
This PR closes #666 and make the `inc` argument in `seq()` built-in function an optional argument.

**Features:**

* Use `seq(from, to)` to generate a column matrix containing an arithmetic sequence of values starting at `from`, going through `to`  in increments of one.

**Changes:**

* make argument `inc` optional with default value `1` in `seq()` function at the DSL level
* Add a script-level test, ensuring that script execution works as expected
* Update `doc/DaphneDSL/Builtins.md` for `seq()` function with optional `inc` arguments


pdamme (approved):
Thanks @saminbassiri for this contribution. Making the `inc` argument of the `seq()` built-in function optional (and default to `1`) simplifies that common case for DaphneDSL users.

**Your code looks very good to me.** Thanks also for the additional script-level test cases and for updating the DaphneDSL documentation.

The only small thing I changed is to use type `si8` instead of `si64` for the default value of `inc` to ensure that it does not have an impact on the value type of the result after type inference/promotion. (Please have a look at the commit I added for details.) You had actually suggested that yourself in our discussion, maybe you just forgot it here.

**Anyway, thanks again for this improvement and congrats on your first completed PR in DAPHNE!**

Issue #658, opened 2024-02-15
stop() built-in/op/kernel


pdamme:
DaphneDSL should offer a `stop()` built-in function which stops the execution of the script (similar to the one in Apache SystemDS). This would be very helpful, e.g., since it would allow UDFs to react to invalid arguments. In such cases, the program should not simply continue (and printing a warning could easily be overlooked), the program should stop.

**Example:**
```R
def foo(x:si64) {
    if(x < 0)
        stop("x must not be less than zero");
    return 1 + sqrt(x);
}
```

**We need:**
- A `stop()` built-in function in DaphneDSL, which should take an optional message string that is displayed along with an error message. If no argument is provided at DSL level, an empty string could be used under the covers.
- A `StopOp` DaphneIR operation, with mandatory message operand.
- A `stop` kernel; the most straightforward way to ensure that an error message is printed and the program execution terminates in a controlled way would be to simply throw a `std::runtime_error` in the kernel. This exception would be caught by DAPHNE's (WIP) error handling infrastructure as any other error.
- Ideally support for `stop` in DaphneLib (Python API), maybe as a function on the `DaphneContext`.
- Unit test cases checking for an exception (`CHECK_THROWS` in catch2) and script-level test cases checking for unsuccessful program termination.

**Hints:**
- Please search for `#658` in the code base (especially in the `scripts/` directory), since multiple places where `stop()` should be used instead of a `print()` are marked that way.
- As a possible extension (and that could be done later), the DAPHNE compiler could detect occurrences of `stop()` that happen unconditionally (e.g., because the check surrounding it has been resolved at compile-time through constant propagation). In such cases, we do not even need to start the script execution, but could immediately abort with the given error message at compile-time.

Garic152:
I'd like to work on this issue.

pdamme:
Great, thanks. Go ahead!

Garic152:
I've just pushed the first version of the stop() function to my fork: [Fork: dev-issue-658](https://github.com/Garic152/daphne/tree/dev-issue-658).

The functionality is as follows:

### **Custom Input**

**File:**
```daphne
stop("Custom error");
```

**Output:**
```
[error]: Execution error: The kernel-function _stop__char failed during runtime with the following message [ System stopped: Custom Error ]
    | Source file -> "/daphne/test.daph":1:0
    |
  1 | stop("Custom Error");
    | ^~~
```

### **No Input**

**File:**
```daphne
stop();
```

```
[...]

PR #763, opened 2024-06-21
[DAPHNE-#658] typeOf() function to determine the type of an object

Garic152:
This PR implements https://github.com/daphne-eu/daphne/issues/357 and allows the user to determine the type of a Scalar or Object in DAPHNE DSL-Scrips using the typeOf() built-in function.

Features:
- `typeOf(scalar)`: Returns the C++ representation of the scalar value type using ValueTypeUtils.h.
- `typeOf(object)`: Returns the dimension and ValueType of an object as `(dim, scalarType)`.

Changes:
- Added the typeOf() DaphneDSL built-in function, the typeOf DaphneIR operation, as well as the typeOf() kernel
- Added corresponding script-level tests as a new test file TypeOfTest.cpp in `.../cli/operations`, ensuring that the correct types get printed for scalar types and both sparse and dense matrices


AlexRTer (changes requested):
Thank you for this contribution @Garic152!
A `typeOf` function is a very useful tool, especially for debugging DaphneDSL scripts or when exploring and verifying type conversions. The implementation **looks very good** to me already. I only made some minor changes such as changing the `TypeOfObj` kernel slightly to avoid having to handle the memory `scaType` points to.

In fact, this PR already highlighted a bug in the way the `DaphneDSLVisitor` handles parsing for special values of float type (e.g. `typeOf(inff)` is a double, not a float)!

There are **two small required changes** that still need to be made before merging.
1.  Please **add the `typeOf` function to the Builtins section of the documentation** (doc/DaphneDSL/Builtins.md) as well as a small description
2. **Add support for the output of one value type per column for Frames** (see the code comment for details)

Please let me know if you have any questions or would like to share your opinion on any changes I made.

Garic152:
Thank you for reviewing the code, I totally overlooked the fact that the frames were able to contain multiple value types.
I changed the `typeOf(frame)` kernel to now iterate over the columns of the frame and return the result formatted like the `print()` function.
I also adjusted the existing tests and added another line testing the newly added functionality with all supported types in one frame.

Let me know if there are still some things missing and I'll add them as soon as possible!

AlexRTer (approved):

AlexRTer:
Thank you for implementing all requested changes and in such a short time frame as well.
From my point of view, **this PR is ready to be merged.**

pdamme (approved):
Thanks for contributing this feature, @Garic152. Thanks for your helpful review and improvements, @AlexRTer. This PR **looks very good** to me, too. It's **ready to be merged**.

I slightly adjusted the positioning of the new `typeOf`-builtin/op by putting it into the "matrix dimensions" category of operations, which I renamed to "matrix meta data". Furthermore, I added a few more instantiations of the `typeOfSca`/`typeOfObj`-kernels, such that we can use them with any scalar or data object.

Issue #759, opened 2024-06-17
MAC OSX support


flashboss:
I'm on Mac OSX 23.5.0 Darwin Kernel Version 23.5.0 arm64

I would compile the last Daphne version but When I do ,/build.sh I get this error:

checking for architecture... arm64
checking for OS... darwin
checking for OS version... 23.5.0
checking for perf_event workaround level... autodetect
checking for if MIC should be used... no
checking for if NEC should be used... no
checking for xlc... no
checking for icc... no
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking for xlf... no
checking for ifort... no
checking for gfortran... gfortran
checking whether we are using the GNU Fortran 77 compiler... yes
checking whether gfortran accepts -g... yes
checking for mpicc... no
checking for gawk... no
checking for mawk... no
checking for nawk... no
checking for awk... awk
checking how to run the C preprocessor... gcc -E
checking whether ln -s works... yes
checking whether make sets $(MAKE)... yes
checking for ranlib... ranlib
checking for grep that handles long lines and -e... /usr/bin/grep
checking for egrep... /usr/bin/grep -E
checking for ANSI C header files... rm: conftest.dSYM: is a directory
rm: conftest.dSYM: is a directory
yes
checking for sys/types.h... yes
checking for sys/stat.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for memory.h... yes
checking for strings.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for unistd.h... yes
checking minix/config.h usability... no
checking minix/config.h presence... no
checking for minix/config.h... no
checking whether it is safe to define __EXTENSIONS__... yes
checking for ANSI C header files... (cached) yes
checking for inline... inline
checking whether time.h and sys/time.h may both be included... yes
checking sys/time.h usability... yes
checking sys/time.h presence... yes
checking for sys/time.h... yes
checking c_asm.h usability... no
checking c_asm.h presence... no
checking for c_asm.h... no
checking intrinsics.h usability... no
checking intrinsics.h presence... no
checking for intrinsics.h... no
checking mach/mach_time.h usability... yes
checking mach/mach_time.h presence... yes
[...]

PR #764, opened 2024-06-21
mac support ongoing

auge:
WiP to support #759


pdamme (approved):
Thanks for this contribution, @auge. Using `mkdir -p` instead of `make --parents` makes the build script more portable and is a necessary step towards supporting DAPHNE on Mac OS.

PR #767, opened 2024-06-26
Hdfs support

aristotelis96:
This PR introduces support for Hadoop Distributed File System (HDFS).

- Added dependency libhdfs3 (Native C/C++ HDFS Client)
- Updated container.
- Kernels for reading and writing to HDFS in Daphne binary format and csv format.
- Distributed kernels to support parallel read/write for the distributed runtime (only synchronous gRPC for now).
- Added cli arguments (and updated config file).


CMakeLists.txt
```diff
@@ -178,6 +178,9 @@ set(CMAKE_VERBOSE_MAKEFILE ON)
 include_directories(${PROJECT_SOURCE_DIR}/src)
 include_directories(${PROJECT_BINARY_DIR}/src)
 
+include_directories(${PROJECT_SOURCE_DIR}/thirdparty/installed/include/hdfs)
```
corepointer, line 181: This line should fail in the docker environment where dependencies are installed in /usr/local

corepointer (changes requested):
This feature is looking good in general. A few things could be improved (see code comments). I did not test-run the code but tried compilation of the dependencies to check support in the Docker images. 
Please make the integration optional like CUDA or MPI.

aristotelis96:
Thank you for the valuable feedback @corepointer. I will address the comments and make HDFS build optional. 👍

aristotelis96:
Thank you @corepointer again for your review. I've added an optional `--hdfs` build argument similar to CUDA and MPI.

corepointer:
LGTM - thx for your contribution @aristotelis96 @psomas @KostasBitsakos 

I had to fix a few things to make everything compile and squashed into two commits for dependencies and feature.
The changes for the Docker scripts should have gone to main before merging the PR - this is why these changes now show up as part of the PR. The final result is the same, just the presented changes here might be confusing.

corepointer:
Test cases would be nice ;-)
I could imagine an automated setup of a mini single-node cluster that reads from localhost [1].
Might be a nice student project :D

[1] https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-common/SingleCluster.html

PR #770, opened 2024-06-28
Refactor getLine() to avoid potential memory leak

psomas:
Daphne's getLine() calls libc's getline() with a NULL lineptr on each invocation. This results in getline() allocating a new buffer on each call, which, I think, never gets freed resulting in a memory leak.

### DIFF
--- a/src/runtime/local/io/File.h
+++ b/src/runtime/local/io/File.h
@@ -23,7 +23,9 @@
 struct File {
   FILE *identifier;
   unsigned long pos;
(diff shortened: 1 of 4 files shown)


corepointer (approved):
LGTM - thx for spotting and fixing this @psomas 

I pimped the commit message a bit and fixed a compiler warning. Other than that I did not bother with the (already present) formatting issues as this whole component should be replaced by modern C++ anyway imho.

PR #771, opened 2024-07-01
657 DaphneLib implement indexing

### LINKED ISSUE #657: DaphneLib: Implement Indexing
Indexing is missing in DaphneLib which would be necessary for certain use cases

Left indexing, starting with 0, incrementing has higher prio imho

saminbassiri:
This PR closes #657  and makes _right_ and _left_ indexing available on matrices in DaphneLib.

**Features:**
* Use _right indexing_ to access specific slices or indexes of a matrix
* Assign a value to  specific slices or indexes of a matrix using _left indexing_

**Changes:**
*  Script-level test cases are added in `test/api/python/`, ensuring that execution works as expected.
*  Add `__getitem__` and `__setitem__` built-in functions for matrices in DaphneLib. 
*  Update the DaphneLib document in  `doc\DaphneLib\Overview.md`  to add indexing.
*  Update `OperationNode` in `src\api\python\daphne\operator\operation_node.py` to generate the special DaphneDSL syntax for left indexing


pdamme (changes requested):
Thanks for this contribution, @saminbassiri. Support for left and right indexing in DaphneLib would be highly desired. Overall, your implementation looks good to me. I only fixed a few minor things and improved the test cases a bit.

While reading the code, I noticed a potential problem with the way we do left indexing via `__setitem__()`. We didn't think of this case when we discussed the way to go. More precisely, as we change the current DAG node to a left indexing operation in `__setitem__()`, this change also affects all other users of the current DAG node. I've added a simple example in `test/api/python/matrix_indexing_3.py`, which triggers this bug. Please have a look.

This problem could be fixed by making all other DAG nodes that use the current DAG node as an input use the newly created copy as their input. To find those nodes, we could establish a bidirectional connection between DAG nodes. Currently, we only have a unidirectional connection, since each node knows its children (producers), but not its parents (consumers). One idea is to store a list of consumers/parents in each `OperationNode`. For a newly created node, this list is empty. Whenever we create a new `OperationNode`, we should add this node to the consumer lists of all the nodes it uses (i.e., its unnamed parameters). Then, in left indexing/`__setitem__()` we need to retrieve the consumers of the current DAG node, and replace all occurrences of the current DAG node in their unnamed parameters by the newly created copy node.

**Required changes before the merge:**
- **Fixing the bug mentioned above** (making `matrix_indexing_3.py` pass). Left indexing on a matrix shall not affect previous consumers of this matrix.

**Optional changes:**
- **Support right indexing by position matrices.** Just like in DaphneDSL it should be possible to use a column matrix of positions as the key in right indexing. E.g.,
  ```python
  pos = dc.from_numpy([1, 5, 27])
  Y = X[pos, :]
  ```
  To that end, we should allow not only `int` and `slice`, but also `Matrix` for the key in right indexing. Left indexing does not need to support it for now. *We definitely need this feature, but in case it turns out to be difficult, we can address it separately after merging this PR.*

saminbassiri:
Thank @pdamme for your assistance.

To fix the problem related to the left indexing, I have made some changes based on your approach as below:

- Add a new attribute `consumer_list` to `OperationNode` class to store consumers of the Node
- Add a new method `update_node_in_input_list` to `OperationNode` class to update the current node in `unnamed_input_nodes`
- In `__setitem__`, replace all occurrences of the current DAG node in the unnamed parameters of the consumers of the current DAG node, with the newly created copy node

pdamme (approved):
Thanks for incorporating the feedback, @saminbassiri. The code looks very good to me. I only simplified the DaphneLib documentation on indexing a bit. **This PR is ready to be merged now.**

PR #723, opened 2024-05-03
Introduce .clang-format file

philipportner:
This PR aims to introduce a `.clang-format` file in order to have automated consistent code formatting for supported languages. For DAPHNE this mainly aims to do so for C, C++ and TableGen files.
`clang-format` is the standard tool to provide consistent formatting across C and C++ code bases and is supported by almost every editor and IDE out there (e.g., automatically recognized by Clion/VScode). 

It even has [integration](https://clang.llvm.org/docs/ClangFormat.html#git-integration) with `git` itself.

The following steps would need to be completed:

- [x] Decide on format
- [x] Run `clang-format` on whole project
- [x] Push these changes with a systems account
- [x] optional - Handle blame annotation for project

I don't really care which format we use, main thing is consistency. I added the `LLVM` style for now since that makes the most sense for me with our dependency on LLVM.

Happy to discuss specifics and feel free to ask any questions and please let me know if there's more steps needed.

https://clang.llvm.org/docs/ClangFormat.html

Issue #518, opened 2023-04-25
Loop vectorization: Transform loops to coarse-grained operations


pdamme:
**Motivation:** DaphneDSL offers users coarse-grained operations from linear and relational algebra, which operate on entire matrices and frames. This is much more efficient than looping over data objects and accessing their elements one by one. Nevertheless, DaphneDSL also allows fine-grained access to individual data elements via left and right indexing. Unfortunately, some users try to re-implement typical operations (e.g., elementwise addition of two matrices) with the fine-grained method, which will use in suboptimal performance. For instance the following two scripts print the same result, but the former is much more efficient and should be preferred:
```
X = reshape([1, 2, 3,  4,  5,  6], 2, 3);
Y = reshape([7, 8, 9, 10, 11, 12], 2, 3);
res = X + Y;
print(res);
```
```
X = reshape([1, 2, 3,  4,  5,  6], 2, 3);
Y = reshape([7, 8, 9, 10, 11, 12], 2, 3);
res = fill(0, nrow(X), ncol(X));
for(i in 0:nrow(X))
    for(j in 0:ncol(X))
        res[i, j] = as.matrix(as.scalar(X[i, j]) + as.scalar(Y[i, j]));
print(res);
```

**Task:** This project is to write a DAPHNE compiler pass that automatically identifies situations when users employ loops and access to individual data elements to re-implement operations which are also available as coarse-grained operations in DAPHNE. If such a case is detected, the intermediate repsentation shall be rewritten to employ the respective coarse-grained operation. Implementation in C++.

**Hints:**
- Interesting operations include: elementwise unary operations, elementwise binary operations, aggregation (full, row-wise, column-wise), ... (amount of work depending on the team size).
- Support at least the case of one loop per coarse-grained operation. Teams can also tackle loops combining multiple operation. For instance
  ```
  for(i in 0:nrow(X))
      for(j in 0:ncol(X))
          res[i, j] = as.matrix(sqrt(as.scalar(X[i, j])) + as.scalar(Y[i, j]) * as.scalar(Z[i, j]));
  ```
  should become
  ```
  sqrt(X) + Y * Z
  ```

PR #777, opened 2024-07-03
Issue #518: Add Loop Vectorization

nuernbergk:
Implements #518: Loop Vectorization.
All ew-Binary and Unary Operations are implemented, as well as nested expressions of these.
Currently **not supported** are casts and aggregation.
The test case `loop-vectorization.mlir` corresponds to [this daphne code](https://gist.github.com/nuernbergk/50a2ff08d1e1944225a485641217bcc6).

Note that there are formatting changes in `DaphneIrExectur.cpp` and `Passes.h`, as I used the `.clang-format` provided in #723.
If necessary, I can of course revert these.

### DIFF
--- a/UserConfig.json
+++ b/UserConfig.json
@@ -15,6 +15,7 @@
     "debug_llvm": false,
     "explain_kernels": false,
     "explain_llvm": false,
+    "explain_loop_vectorization": false,
     "explain_parsing": false,
     "explain_parsing_simplified": false,
     "explain_property_inference": false,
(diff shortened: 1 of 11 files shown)


philipportner:
> The build failed on the include of the `SCF.h` file, this might because of a typo (capitalization error). Fixed. Can't test on my machine because it worked there even before fixing.

Hi Nils, this works, I applied the same patch before when compiling the code on my machine. Thanks for pushing the fix!

Issue #766, opened 2024-06-25
Sparsity estimation for most DaphneIR ops


pdamme:
In order to offer *physical data independence* to users, DAPHNE must be able to automatically decide the physical representation of a data object. In this context, the decision for a dense or a sparse representation of a matrix is especially crucial for linear algebra operations. This decision is based on *sparsity estimates* for the intermediate results. DAPHNE can already make this decision *for some DaphneIR operations*, but to support informed decisions throughout a larger program, we need sparsity estimation for the results of (virtually) all DaphneIR operations.

**Hints:**
- For some general background on sparsity estimation, see Matthias Boehm's AMLS lecture at TU Berlin: [03_Compilation.pdf](https://mboehm7.github.io/teaching/ss24_amls/03_Compilation.pdf) (slide 10ff).
- Get familiar with MLIR [traits](https://mlir.llvm.org/docs/Traits/) and [interfaces](https://mlir.llvm.org/docs/Interfaces/), since we use these concepts for type and property inference in the DAPHNE compiler.
- Get familiar with type and property inference in the DAPHNE compiler.
  - For a high-level overview: DAPHNE project deliverables [D3.3](http://daphne-eu.eu/wp-content/uploads/2023/05/DAPHNE_D3.3_ExtendedCompilerPrototype_v1.1.pdf) and [D3.4](http://daphne-eu.eu/wp-content/uploads/2023/12/D3.4-Compiler-Design-and-Overview-.pdf).
  - Traits and interfaces for the type, shape, sparsity, etc. can be found in `src/ir/daphneir/`: `DaphneInfer*OpInterface.td/h/cpp` and `DaphneInfer*Traits.td`.
  - These traits and interfaces are attached to the DaphneIR ops in `src/ir/daphneir/DaphneOps.td`.
  - The actual type and property inference happens in the inference pass (`src/compiler/inference/InferencePass.cpp`).
  - Given a DaphneDSL script, the infered types and properties can be displayed by `bin/daphne --explain property_inference myScript.daphne`.
[...]

PR #790, opened 2024-07-19
Implementation of new SparsityOp

Garic152:
This PR introduces a new `SparsityOp` that allows users to print the compile-time sparsity estimate of a matrix to the screen. The function can be called using `sparsity(Matrix)`, which will be replaced with the sparsity value through canonicalization if the sparsity is known. If the sparsity is unknown, it will return -1.

This PR does not yet include tests. The SparsityOp functionality will be tested as part of [#766](https://github.com/daphne-eu/daphne/issues/766).

### DIFF
--- a/doc/DaphneDSL/Builtins.md
+++ b/doc/DaphneDSL/Builtins.md
@@ -100,6 +100,11 @@ The following built-in functions allow to find out the shape/dimensions of matri
 
     Returns the number of cells in `arg`.
     This is the product of the number of rows and the number of columns.
+  
+- **`sparsity`**`(arg:matrix)`
+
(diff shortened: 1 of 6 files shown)


pdamme (approved):
Thanks for this contribution, @Garic152. The `sparsity` built-in/op will be useful for testing sparsity estimation for now (maybe we find a better solution in the future, but I think this approach is quite straightforward). **Your code looks very good to me.** I only applied some minor changes (see my commit). **This PR is ready to be merged.**

pdamme:
All tests pass successfully on my machine and did pass in the CI before my mostly syntactical changes. At the moment, the CI fails due to problems in setting up the container. Thus, I think it's safe to merge this PR.

Issue #774, opened 2024-07-02
Simplification rewrites for elementwise unary minus (additive inverse)


pdamme:
DaphneDSL supports the elementwise additive inverse through the typical unary minus operator:
```
X = [1.1, 2.2, 3.3];
print(-X);
```

Internally, this operation is represented as a `EwMinusOp` in DaphneIR.

**Task:** This task is to add typical simplifications of this operation to the DAPHNE compiler, such as:
- *Constant folding:* Expressions like `-123` could be evaluated at compile-time.
- *Static rewrites:* E.g., `--x` to `x`.

Overall, such simplications can make the IR simpler and more readable, unlock further simplifications of consuming operations, and reduce the computational effort at run-time.

**Hints:**
- Get familiar with [canonicalization and constant folding in MLIR](https://mlir.llvm.org/docs/Canonicalization/).
- Have a look at the existing implementations of such simplifications for other DaphneIR operations in `src/ir/daphneir/DaphneDialect.cpp`.

ldirry:
Hello, please assign me to this issue and #775 .

pdamme:
Thanks for your interest. Sure, go ahead!

PR #796, opened 2024-07-28
[DAPHNE-#774] Simplification rewrites for elementwise unary minus (ad…

ldirry:
**Task**: This task is to add typical simplifications of this operation to the DAPHNE compiler, such as:

- Constant folding: Expressions like -123 could be evaluated at compile-time.
- Static rewrites: E.g., --x to x.

Overall, such simplications can make the IR simpler and more readable, unlock further simplifications of consuming operations, and reduce the computational effort at run-time.

- Implemented canonicalization rule: --x simplifies to x
- Added constant folding for -x where applicable
- Added relevant test cases to verify new simplifications rewrites

Tests were performed as an end to end test where the Output of Daphne is compared with a Referencefile. To verify that the rewrite pattern and constant folding for additive inverse are working, Daphne is being executed with the explain=parser_simplified flag.

### DIFF
--- a/src/ir/daphneir/DaphneDialect.cpp
+++ b/src/ir/daphneir/DaphneDialect.cpp
@@ -611,6 +611,26 @@ mlir::Attribute constFoldBinaryOp(mlir::Location loc, mlir::Type resultType, llv
     }
     return {};
 }
+template<class AttrElementT,
+    class ElementValueT = typename AttrElementT::ValueType,
+    class CalculationT = std::function<ElementValueT(const ElementValueT &)>>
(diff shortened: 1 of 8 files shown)


corepointer (commented):
The code LGTM and could be merged if the tests run through imho. But the example "--x" raises the question: Do we not want to have prefix and suffix increment/decrement operators in our DSL?

philipportner:
I triggered the tests again as this seems to be related to #729

corepointer:
Merging as promised, now that the tests run again (problems caused by an unrelated issue)

Issue #773, opened 2024-07-02
Undetected invalid script args


pdamme:
When calling `daphne`, users can provide named *script arguments* that are available inside the DaphneDSL script. For a very simple example, see `test/api/cli/scriptargs/printSingleArg.daphne`:

```
print($foo);
```

According to the [DaphneDSL language reference](https://daphne-eu.github.io/daphne/DaphneDSL/LanguageRef/#script-arguments), only literals may be passed as script arguments. For example:

```bash
bin/daphne test/api/cli/scriptargs/printSingleArg.daphne foo=10
```
```
10
```

However, the following calls of `daphne` with invalid script arguments all pass "successfully" (status code 0) and some don't even indicate any error. Note that none of them behaves as one might expect when passing the respective invalid argument:

- ```bash
  # Unclear what the user may have wanted.
  bin/daphne test/api/cli/scriptargs/printSingleArg.daphne foo=10xyz23
  ```
  ```
  10
  ```
- ```bash
  # Seems like the user wanted DAPHNE to evaluate the expression.
  bin/daphne test/api/cli/scriptargs/printSingleArg.daphne foo="1.23*4.56"
  ```
  ```
  1.23
  ```
- ```bash
  # Seems like the user wanted DAPHNE to evaluate the expression.
  bin/daphne test/api/cli/scriptargs/printSingleArg.daphne foo="sin(1.23)"
  ```
  ```
  line 1:0 mismatched input 'sin' expecting {'true', 'false', INT_LITERAL, FLOAT_LITERAL, STRING_LITERAL}
  1.23
  ```

pdamme:
Search for `see #773` in the code base.

ldirry:
Hi, please assign me to this issue.

PR #800, opened 2024-08-04
[DAPHNE-#773] Undetected invalid script args

ldirry:
When calling daphne, users can provide named script arguments that are available inside the DaphneDSL script. 
According to the [DaphneDSL language reference](https://daphne-eu.github.io/daphne/DaphneDSL/LanguageRef/#script-arguments), only literals may be passed as script arguments

However, the following calls of daphne with invalid script arguments all passed "successfully" (status code 0) and some didn't even indicate any error. 

- Fixed invalid script arguments that try to evaulate an expression
- `bin/daphne test/api/cli/scriptargs/printSingleArg.daphne foo="sin(1.23)"`
- `bin/daphne test/api/cli/scriptargs/printSingleArg.daphne foo="1.23*4.56"`
- Fixed invalid script arguments that were not literals
- `bin/daphne test/api/cli/scriptargs/printSingleArg.daphne foo=10xyz23`
- Added testcases for single script arguments
- Added testcases where Daphne fails in case of an invalid script arguments

### DIFF
--- a/src/parser/daphnedsl/DaphneDSLVisitor.cpp
+++ b/src/parser/daphnedsl/DaphneDSLVisitor.cpp
@@ -794,27 +794,41 @@ antlrcpp::Any DaphneDSLVisitor::visitArgExpr(DaphneDSLGrammarParser::ArgExprCont
             "argument " + arg + " referenced, but not provided as a command line argument"
         );
 
-    bool hasMinus;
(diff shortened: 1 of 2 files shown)


corepointer:
Thx for your work @ldirry 

Code and tests LGTM

PR #609, opened 2023-09-11
Enable Update-in-Place Optimization in DAPHNE (Issue #498)

sweetpellegrino:
## Overview
This change introduces an approach to enable in-place operand updates (see [#498](https://github.com/daphne-eu/daphne/issues/498)). It aims to reduce memory allocation and potentially shorten execution time. The chosen approach propagates compile-time knowledge to the runtime to make accurate decisions about the reusability of an operand in the kernel.

## Implementation
**For that the following changes were made:**
1. New OpInterface (src/ir/daphneir/“DaphneUpdateInPlaceOpInterface”) for denoting which operands of an operation in Daphne should be considered for in-place update.
2. Modified the operation definitions in DaphneOps.td to accommodate the changes.
3. New compiler pass (src/compiler/lowering/FlagUpdateInPlacePass.cpp), that checks the future use of an operand.
4. Updated RewriteToCallKernelOpPass.cpp to include compile-time knowledge as boolean values for each operation.
5. Introduction of a utility class (src/runtime/local/kernels/InPlaceUtils.h) for providing methods for uniformly checking the possibility of being in-placeable. 
6. Changes to individual kernels, which includes changes to the function signature and algorithms.

## Tests, Documentation & Benchmarking
There is a series of new tests (test/inplace/\*) that ensure the stability of this feature. These tests include individual function tests (InPlaceKernelTest.cpp and InPlaceUtilsTest.cpp) as well as checks to ensure the decision-making of the compiler pass (InPlaceAPITest.cpp).

Furthermore, we have created new documentation (doc/development/EnablingInPlaceUpdates.md) that describes how to enable your own kernel for implementing update-in-place.

The benchmarking approach is described in bench/README.md. We utilized Catch2 features to implement a microbenchmark and included it in our test cases (test/inplace/InPlaceMicroBench.cpp). To conduct the end-to-end tests, we employed various DAPHNE scripts and measured their execution time and peak memory usage.

## Additional notes

**Currently supported Kernels:**

| Kernel Name    | Dense | CSR | Frame |
| -------------- | ----- | --- | ----- |
| EwBinaryMat    | ✅    | ❌  |       |
| EwBinaryObjMat | ✅    | ❌  | ✅    |
| EwUnaryMat     | ✅    | ❌  |       |
| Transpose      | ✅    | ❌  |       |
| Reverse  | ✅ | ❌   |  |

None of these kernels currently support in-place updates when using CUDA or FPGA.

Additional kernels/operations that could support in-place updates:
[...]

PR #806, opened 2024-08-07
[DAPHNE-#755] Initial support for lists in DaphneDSL.

### LINKED ISSUE #755: Lists in DaphneDSL
We need support for lists in DaphneDSL that can store matrices of homogenous type but heterogeneous shapes. These lists should at least support appending elements to the back and extracting/removing elements from the front.

Background: The decision tree implementation we ported from Apache SystemDS (see `scripts/algorithms/decisionTree_.daph`) uses lists *as queues* in the original implementation. As DaphneDSL does not support lists yet, we emulate them using concatenation of matrices (`rbind`/`cbind`), but this is very expensive due to repeated copying, if the stored matrices are very large. At the moment, the management of these queues takes up most of the time in our decision trees implementation.

pdamme:
This PR adds basic support for lists in DaphneDSL. These are urgently required to speed up the decision trees DaphneDSL script, whose performance is crucial to our use case partners.

As the introduction of lists is a larger change, I would like to make everyone interested aware and enable a discussion on this PR. From my point of view, this PR is fine (the limitations mentioned below are not crucial and can be addressed later).

-----

**From the commit message:**

- This PR adds basic support for lists of matrices of homogeneous physical data type and value type.
- Supported operations:
  - createList(), length(), append(), remove(), print()
  - append() and remove() do not modify the given list, but return the resulting list (consistent with the bahavior of matrix/frame operations)
- Concrete changes:
  - DaphneDSL - Four new built-in functions: createList(), length(), append(), remove(). - Updated the DaphneDSL language reference and the list of built-in functions.
  - DaphneIR/DAPHNE compiler
    - A new custom MLIR type: List.
    - Four new DaphneIR operations: CreateListOp, LengthOp, AppendOp, RemoveOp. - Type inference for CreateListOp, AppendOp, and RemoveOp. - Consideration of list types in the kernel extension catalog. - Lowering of the four new ops to kernel calls. - Lowering of the List type in LowerToLLVMPass in the same way as matrices/frames.
  - DAPHNE runtime - Four new kernels: createList, length, append, remove. - Registration of the new kernels in kernels.json.
  - Garbage collection of list items
    - ManageObjRefsPass treats lists like matrices/frames.
    - Besides that:
      - The reference counter of a data object is increased when it is inserted into a list to ensure that the object is not freed as long as the list exists.
      - When a list is destroyed, the reference counter of all its elements is decreased by 1. - When an element is removed from a list, its reference counter is *not* decreased, because we return the removed element (it lives on).
  - Added script-level test cases for the usage of lists.
- Current limitations:
  - No support for heterogeneous data/value type (e.g., one cannot store DenseMatrix and CSRMatrix or DenseMatrix<double> and DenseMatrix<float> in the same list).
  - Not possible to create an empty list in DaphneDSL, because that would complicate type inference.
  - Lists are not supported in DaphneLib yet.
  - No get/set on list elements yet, only append and remove.
  - The append/remove kernels copy the input list instead of modifying it in place. While this is consistent with the behavior of matrix/frame ops, it is inefficient. However, compared to workarounds to lists, this is still much faster; Furthermore, we can use the same update-in-place mechanisms as for matrices/frames in the future, when the input list is not used anymore afterwards (see PR #609).
  - Information about interesting data properties gets lost by inserting and removing a matrix into/from a list.
  - Subclassing Structure may not be optimal; it might be better to subclass a new superclass of Structure, but I wanted to keep the refactoring overhead low for now.
- Effect on decision trees DaphneDSL script
  - The decision trees script uses queues to keep certain matrices around.
  - So far, we emulated these queues through matrix concatenation (cbind/rbind), which turned out to be very inefficient.
  - The script uses lists for the queues now, which leads to significant performance improvements (~15x for a concrete use-case pipeline on my machine).
- Closes #755.

### DIFF
--- a/doc/DaphneDSL/Builtins.md
+++ b/doc/DaphneDSL/Builtins.md
@@ -49,6 +49,7 @@ DaphneDSL's built-in functions can be categorized as follows:
 - Input/output
 - Data preprocessing
 - Measurements
+- List operations
 
 ## Data generation
 
@@ -652,3 +653,24 @@ These must be provided in a separate [`.meta`-file](/doc/FileMetaDataFormat.md).
 - **`now`**`()`
 
     Returns the current time since the epoch in nano seconds.
+
+## List operations
+
+- **`createList`**`(elm:matrix, ...)`
+
+    Creates and returns a new list from the given elements `elm`.
+    At least one element must be specified.
+
+- **`length`**`(lst:list)`
+
+    Returns the number of elements in the given list `lst`.
+
+- **`append`**`(lst:list, elm:matrix)`
+
+    Appends the given matrix `elm` to the given list `lst`.
+    Returns the result as a new list (the argument list stays unchanged).
+
+- **`remove`**`(lst:list, idx:size)`
+
+    Removes the element at position `idx` (counting starts at zero) from the given list `lst`.
+    Returns (1) the result as a new list (the argument list stays unchanged), and (2) the removed element.
--- a/doc/DaphneDSL/LanguageRef.md
+++ b/doc/DaphneDSL/LanguageRef.md
@@ -45,7 +45,7 @@ Variables are used to refer to values.
 
 **Valid identifiers** start with a letter (`a-z`, `A-Z`) or an underscore (`_`) that can be followed by any number of letters (`a-z`, `A-Z`), underscores (`_`), and decimal digits (`0-9`).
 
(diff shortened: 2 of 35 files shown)


philipportner:
Great feature to add to DAPHNE, thanks @pdamme !
I rebased the branch and merged it in as people want to see this in main.

I've done a code review, overall the code looks very good to me, I have noted down a list of things I'd like to follow up on once you're back. As I don't think these should block the PR and can be fixed in the future I merged it into main.

Issue #717, opened 2024-05-02
Handle DAPHNE failure in DaphneLib


pdamme:
DaphneLib internally calls DAPHNE. More precisely, the Python function `execute()` in `src/api/python/daphne/script_building/script.py` calls the compiled C++ function `daphne()` in `src/api/daphnelib/daphnelib.cpp`, which in turn calls the compiled C++ function `mainInternal()` in `src/api/internal/daphne_internal.cpp`. `main_internal()` runs DAPHNE's entire parser/compiler/runtime stack. Currently, DaphneLib does not check if any error happened in DAPHNE nor does it react in any meaningful way. This should be improved for a more robust DaphneLib including meaningful error messages to DaphneLib users.

saminbassiri:
Hi, Please assign me to this issue.

philipportner:
Hey @saminbassiri , go ahead!

PR #808, opened 2024-08-12
[DAPHNE-#717]  handle daphne failure in daphne lib

saminbassiri:
[DaphneLib-#717]  handle daphne failure in daphne lib
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Summary
This PR closes issue #717 by handling execution errors in DaphneLib.

Changes:

* `error_message`: Introduced a new variable in DaphneLibResult to store execution error messages from DSL scripts.
* Error Handling in C++: Updated `src/api/internal/daphne_internal.cpp`  to capture and store execution error messages in the `error_message` variable when a DSL script fails.
* Error Handling in Python: Enhanced error handling in `src/api/python/daphne/script_building/script.py` to process and log execution error messages during DSL script execution.

### DIFF
--- a/src/api/daphnelib/DaphneLibResult.h
+++ b/src/api/daphnelib/DaphneLibResult.h
@@ -17,6 +17,8 @@
 #pragma once
 
 #include <cinttypes>
+#include <string>
+
 
 struct DaphneLibResult {
     // For matrices.
@@ -28,4 +30,6 @@ struct DaphneLibResult {
     int64_t* vtcs;
     char** labels;
     void** columns;
+    // To pass error message to Python code.
+    std::string error_message;
 };
--- a/src/api/daphnelib/DaphneLibResult.h
+++ b/src/api/daphnelib/DaphneLibResult.h
@@ -30,6 +30,6 @@ struct DaphneLibResult {
     int64_t* vtcs;
     char** labels;
     void** columns;
-    // To pass error message to Python code.
+    // To pass error messages to Python code.
     std::string error_message;
 };
--- a/src/api/internal/daphne_internal.cpp
+++ b/src/api/internal/daphne_internal.cpp
@@ -95,6 +95,13 @@ void handleSignals(int signal) {
     longjmp(return_from_handler, gSignalStatus);
 }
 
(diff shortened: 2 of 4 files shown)


pdamme (approved):
Thanks for this contribution, @saminbassiri. Converting DAPHNE status codes into Python exceptions in DaphneLib makes DaphneLib much more robust in presence of errors. Overall, your approach and code **looks good to me**.

So far, your code only addressed execution errors from DAPHNE. As there are more things that could go wrong (e.g., parser errors, compiler errors, etc.), I generalized the code a bit such that these kinds of errors are also handled (see my commit). Your idea of suppressing the information on the DaphneDSL code line responsible for the error is understandable, since the DaphneDSL script is generated by DaphneLib and not human-friendly. However, as it can be very hard to find the error without further information, I think it's better to keep this indication. Ideally, we should improve the error location information in the future to simplify identifying the responsible *DaphneLib* expression.

pdamme:
Here is an example for triggering a DAPHNE execution error from DaphneLib:

Contents of `py-example.py`:
```python
from daphne.context.daphne_context import DaphneContext
dc = DaphneContext()
X = dc.seq(1, 3) + dc.seq(1, 2) # mismatching shapes
X.print().compute()
```
Output of `./run-python py-example.py`:
```
Traceback (most recent call last):
  File "py-example.py", line 11, in <module>
    X.print().compute()
  File "/daphne/src/api/python/daphne/operator/operation_node.py", line 126, in compute
    self._script.execute()
  File "/daphne/src/api/python/daphne/script_building/script.py", line 106, in execute
    raise RuntimeError(f"Error in DaphneDSL script: {error_message}")
RuntimeError: Error in DaphneDSL script: Execution error: The kernel-function _ewAdd__DenseMatrix_int64_t__DenseMatrix_int64_t__DenseMatrix_int64_t failed during runtime with the following message [ EwBinaryMat(Dense) - lhs and rhs must either have the same dimensions, or one of them must be a row/column vector with the width/height of the other, but lhs has shape (3 x 1) and rhs has shape (2 x 1) ]
    | Source file -> "/tmp/DaphneLib/tmpdaphne.daphne":3:6
    | 
  3 | V2=V0 + V1;
    |       ^~~


```

PR #809, opened 2024-08-15
Codegen for AggRow/Col summation

AlexRTer:
This PR implements a lowering pass for `sumRow` and `sumCol` operations on a `DenseMatrix` of any value type (except `bool`) using MLIR. Both of these passes are part of the optional MLIR codegen pass (`--mlir-codegen`) and extend the amount of operations that can be compiled into MLIR instructions instead of calling precompiled C++ kernels. Currently, neither `sumRow` nor `sumCol` support `bool` types in normal execution (without the `--mlir-codegen` flag) which seems to be a separate issue.

**Changes:**
- Add codegen for `sumRow` and `sumCol`
- Script level test cases
- MLIR test cases (using `FileCheck`)
- Added instantiations of the `aggRow`, `aggCol` kernels with `DenseMatrix<uint64_t>`
- Restored alphabetic sorting of codegen passes in `ir/daphneir/Passes.h`

Both passes rely on nested `AffineForOp`s with a simple body that relies on the `Affine` dialect for loading and storing values, as well as the `Arith` dialect for the running sum. Therefore, loop fusion of multiple `AffineForOp`s from different operations should be possible (e.g. fusing two separate nested loops for `sumRow` and `sumCol` into a single nested loop) but hasn't been added in this PR. Furthermore, it relies on the `Memref` dialect to allocate memory for the result.

A short example:
```
// ./bin/daphne --mlir-codegen
print(sum(fill(1.0, 3, 3), 0));   // sumRow
```
Relevant slice of the IR after `sumRow` lowering:
```
...
%cst = arith.constant 0.000000e+00 : f64
...
%5 = "daphne.convertDenseMatrixToMemRef"(%4) : (!daphne.Matrix<3x3xf64>) -> memref<3x3xf64>
%alloc = memref.alloc() : memref<3x1xf64>
%intptr = memref.extract_aligned_pointer_as_index %alloc : memref<3x1xf64> -> index
%6 = "daphne.convertMemRefToDenseMatrix"(%intptr, %c0, %c3, %c1, %c1, %c1) : (index, index, index, index, index, index) -> !daphne.Matrix<3x1xf64>
affine.for %arg0 = 0 to 3 {
  %7 = affine.for %arg1 = 0 to 3 iter_args(%arg2 = %cst) -> (f64) {
    %8 = affine.load %5[%arg0, %arg1] : memref<3x3xf64>
    %9 = arith.addf %arg2, %8 : f64
    affine.yield %9 : f64
  }
  affine.store %7, %alloc[%arg0, 0] : memref<3x1xf64>
}
...
```

Testing a combination of `AffineVectorLoadOp` (with different vector sizes) and `vector::ReductionOp` in an attempt to vectorize the loops seemd to be slightly but consistently slower. Since it could possibly make loop fusion more difficult due to the added complexity and did not offer obvious benefits, elements are instead simply loaded individually.


AlexRTer:
As discussed, this PR will be replaced with #889 which uses Linalg GenericOps instead of lowering directly to Affine Loops.

Issue #772, opened 2024-07-02
Constant folding crashes for values of different types


pdamme:
[Constant folding](https://mlir.llvm.org/docs/Canonicalization/#canonicalizing-with-the-fold-method) enables the DAPHNE compiler to calculate many operations on constants already at compile-time. The respective code can be found in `src/ir/daphneir/DaphneDialect.cpp`.

For instance, in the following DaphneDSL script, constant folding evaluates 1 * 1 at compile-time, as can be seen when looking at the IR after parsing (no constant folding yet) and some initial simplifications (including constant folding) (use `--explain parsing,parsing_simplified`):

```R
print(1 * 1);
```

```
IR after parsing:
module {
  func.func @main() {
    %0 = "daphne.constant"() {value = 1 : si64} : () -> si64
    %1 = "daphne.constant"() {value = 1 : si64} : () -> si64
    %2 = "daphne.ewMul"(%0, %1) : (si64, si64) -> si64
    %3 = "daphne.constant"() {value = true} : () -> i1
    %4 = "daphne.constant"() {value = false} : () -> i1
    "daphne.print"(%2, %3, %4) : (si64, i1, i1) -> ()
    "daphne.return"() : () -> ()
  }
}
IR after parsing and some simplifications:
module {
  func.func @main() {
    %0 = "daphne.constant"() {value = 1 : si64} : () -> si64
    %1 = "daphne.constant"() {value = false} : () -> i1
    %2 = "daphne.constant"() {value = true} : () -> i1
    "daphne.print"(%0, %2, %1) : (si64, i1, i1) -> ()
    "daphne.return"() : () -> ()
  }
}
1
```

However, currently, constant folding requires the operands to have the same type. For instance, the following two DaphneDSL scripts would be perfectly amenable to constant folding, but make the DAPHNE compiler crash with the respective error messages:

- ```R
  print(1 * as.si32(1));
  ```
  ```
  daphne: /daphne/thirdparty/llvm-project/llvm/lib/Support/APInt.cpp:226: llvm::APInt llvm::APInt::operator*(const llvm::APInt &) const: Assertion `BitWidth == RHS.BitWidth && "Bit widths must be the same"' failed.
  ...
  Segmentation fault (core dumped)
  ```
- ```R
  print(1.0 * as.f32(1.0));
  ```
  ```
[...]

PR #813, opened 2024-08-22
[DAPHNE-#772] Fix Constant folding crashes for values of different types

ldirry:
Constant folding required the operands to have the same type. For instance, the following two DaphneDSL scripts would be perfectly amenable to constant folding, but make the DAPHNE compiler crash with the respective error messages:
`print(1 * as.si32(1));`

`daphne: /daphne/thirdparty/llvm-project/llvm/lib/Support/APInt.cpp:226: llvm::APInt llvm::APInt::operator*(const llvm::APInt &) const: Assertion BitWidth == RHS.BitWidth && "Bit widths must be the same"' failed.
...
Segmentation fault (core dumped)`

FIxed it by comparing the bit width in the `constFoldBinaryOp` method. If they are not the same, the smaller one is getting type promoted to the bigger one (reusing the `CastOp` method to perfom this type promotion).

Aded testcases for binary op of different types

### DIFF
--- a/src/ir/daphneir/DaphneDialect.cpp
+++ b/src/ir/daphneir/DaphneDialect.cpp
@@ -569,6 +569,8 @@ mlir::LogicalResult mlir::daphne::VectorizedPipelineOp::canonicalize(mlir::daphn
 // For families of operations.
 
 // Adapted from "mlir/Dialect/CommonFolders.h"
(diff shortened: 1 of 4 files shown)


corepointer:
LGTM
Tests run fine
Thx for your contribution @ldirry

PR #820, opened 2024-09-03
[DAPHNE-#77] Avoid stack overflow in long-running DaphneDSL loops.

### LINKED ISSUE #77: DaphneDSL loops crash after a certain number of iterations
In GitLab by @pdamme on Jun 12, 2021, 20:00

The following two simple programs both crash with a segfault after approximately 261800 iterations. The precise number of iterations is individual for each attempt:

```
for(i in 1:1000000)
    print(i);
```

```
i = 0;
while(i < 1000000) {
    print(i);
    i = i + 1;
}
```

The reason is unclear, but might be some kind of memory leak due to the way we call kernels.

Furthermore, endless loops also crash with a segfault. It is unclear if this has the same cause.

```
while(1==1)
    i = 1;
```

-----

This is not a very urgent problem, but might indicate that something is going wrong internally.

pdamme:
- As described in #77, DaphneDSL loops with many iterations (e.g., >100k) crash DAPHNE.
- The reason is that LowerToLLVMPass creates LLVM::AllocaOp in several places and for various purposes.
- These AllocaOps could also be inside a loop, which means that the memory they allocate piles up until the stack is exhausted, because it is only freed at the end of the "scope" (function).
- This commit solves the problem by ensuring that all AllocaOps created in LowerToLLVMPass are inserted at the beginning of the function surrounding the currently considered operation.
- With that, there are no AllocaOps in loops anymore.
- Note that the memory allocated by the AllocaOps can safely be reused by repeated kernel calls in different loop iterations.
- Added script-level test cases employing loops with many iterations.
- This bug has existed for a long time (partly because it wasn't a blocking issue until recently).
  - Thanks to @corepointer and @philipportner for investigating the causes (this fix is based upon their insights).
  - Thanks to @J-Brest and @borkob for providing a work-around for DAPHNE users.
- Closes #77.

-----

I think, in the future, we could even reduce the number of `AllocaOps` by making different kernel calls use the same memory slots for their result pointers. However, the most important goal for now is to get this bug fixed on main.

### DIFF
--- a/src/compiler/lowering/LowerToLLVMPass.cpp
+++ b/src/compiler/lowering/LowerToLLVMPass.cpp
@@ -45,6 +45,20 @@
 
 using namespace mlir;
 
+// Remark on the creation of mlir::LLVM::AllocaOp
+// ==============================================
+// This pass creates an mlir::LLVM::AllocaOp in several places and for various purposes,
+// e.g., to store the result pointer of a kernel call, for variadic operands/results, etc.
+// AllocaOp should not be inside a loop, as its repeated execution at run-time can lead
+// to a stack overflow (depending on the number of iterations, the number of AllocaOps
+// inside the loop, and the stack size). The reason is that the memory allocated by AllocaOp
+// is freed only at the end of the scope (i.e., function).
+// To avoid such problems, we don't create AllocaOps at the original insertion point of
+// the rewriter, but at the beginning of function surrounding the currently considered op.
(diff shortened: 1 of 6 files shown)


corepointer:
Awesome @pdamme  
Will test it right away :+1:

corepointer:
I've been playing around with various memory profilers to test the issue and the solution. Stack usage looks quite normal with your changes applied. Also the explain_llvm output shows the improvements nicely. Thx for fixing this long standing bug at last @pdamme

Issue #516, opened 2023-04-25
Split generated kernels.cpp for faster builds


pdamme:
tba

philipportner:
FYI: I'm working on this one.

philipportner:
I'll post this as an information dump for now:

`ninja` provides a log file which can be converted to `chrome` tracing format with the tool [`ninjatracing`](https://github.com/nico/ninjatracing). Using [ui.perfetto](https://ui.perfetto.dev) these can be simply viewed in the browser.
I looked at this trace, the biggest offender is the `kernels.cpp` file mentioned in this issue, the second largest contribution to compile time is compilation of `DaphneDialect.cpp`.

As the granularity from the `ninja` log is not that helpful, I tried to use `-ftime-trace` from `clang`. As compilation with `clang` does somewhat work even thought execution fails, the time trace can still be extracted.
With that, we have some profiling information about why individual compilation units take long.

`kernels.cpp` frontend takes roughly 25% of the time, with the backend taking the other 75%. Roughly half of the frontends time is spent doing template instantiations. Here the biggest offender is `Eigen::EigenSolver<Eigen::Matrix<T` with two types, `double` and `float`. The backend does not appear to have a single source of time spent, rather a lot of function-level optimization passes + codegen passes are performed.

Overall it looks like splitting `kernels.cpp` into multiple compilation units should do the trick for `kernels.cpp`.

Looking at the fine-grained `-ftime-trace` of `DaphneDialect.cpp` I believe that we should do the same there.

To reproduce the traces:
- For the `.ninja_log` trace clone `ninjatracing` and run `./ninjatracing ../daphne/build/.ninja_log > trace.json`
- For `-ftime-trace` one needs to compile `daphne` with `clang`, `export CXX=/path/to/clang++` and `export CC=/path/to/clang` and add `set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -ftime-trace")` to `CMakeLists.txt`. The traces can be found right alongside the compilation unit, for `kernels.cpp` that would be `daphne/build/src/runtime/local/kernels/`

 
[ninja_log_trace.json](https://github.com/daphne-eu/daphne/files/15185336/trace.json)
[DaphneDialect.cpp.json](https://github.com/daphne-eu/daphne/files/15185333/DaphneDialect.cpp.json)
[kernels.cpp.json](https://github.com/daphne-eu/daphne/files/15185335/kernels.cpp.json)

corepointer:
[...]

Issue #474, opened 2023-02-16
Suggestion to speed up LLVM/MLIR dependency check out


corepointer:
The initial checkout of the LLVM mono repo submodule takes a long time. In the course of making DAPHNE compilation from released source artifacts possible I found a way to make the compilation process work without the git submodule (by downloading a "snapshot-from-commit-hash"). See the source lines of build.sh around the following link for details:
https://github.com/daphne-eu/daphne/blob/f314fdd34d85df10444d2878d5e2b22cad075349/build.sh#L778

I timed both methods of fetching the LLVM mono repo and the time spent comes down from 5 minutes 7 seconds to about 36 seconds. With this major time saver I suggest to remove the submodule check out from build.sh. People who want to have the checked out submodule can still use pull.sh but most people will not bother anyway and get a quicker initial build process.

PR #829, opened 2024-09-23
Improve compile times

philipportner:
This PR aims to improve compile times of DAPHNE itself (`./build.sh`). Improving the compilation times of a fully fresh build, including dependencies is **_not_** the goal of this PR.

To establish some baselines, on my machine a fresh DAPHNE build normally takes about **2min 35s**.

### Before ~2min 35s
Here we can see a profiled run the current state (9a1b93e4fea27e91d14c18821be2c940adf63bf4)
image

I've made some changes, mainly splitting up the `kernels.cpp` file that is being generated by `genKernelInst.py` into compilation units per kernel. This improves parallelization as each translation unit can be built by itself. Yes, this does introduce some overhead in parsing the same headers in multiple translation units but for now the results are way better than before. I've also tried splitting into fewer translation units and splitting the kernels each into a translation unit provided the fastest compile times.

Modifying and recompiling a single/couple headers now also only takes a few seconds as they only need to recompile the changed translation units instead of all the kernels.

Additionally, I've made the `libAllKernels` an `OBJECT` library. This enables parallel compilation of the kernels library and the compiler parts.

### Current ~1min 30s
With that, compilation times come down to about **1min 30s**. The result can be seen here.
image

There still are some heavy hitters, the biggest offenders being the files that depend on `MLIRDaphne` and include `daphne.h` and the `EigenCal.h` kernel, which brings in the `Eigen` dependency.

Removing these 4 kernels from compilation would reduce the compilation time down to **1min**. IMO we should move these few kernels out of the `AllKernels` target to decouple the kernel runtime from the compiler infrastructure.
Including `daphne.h` introduces a significant amount of overhead in multiple translation units and is included all over the codebase. I think decoupling some of the MLIR related parts from `daphne.h` would improve overall compilation times and should be done in the future as this PR concerns itself only with the kernels library.

Open issues:

- [x] fix duplicate symbols when compiling tests
- [ ] split kernel also for CUDA compilation, although I have to look more


corepointer:
Nice progress with this issue :ok_hand: 

In general it is more beneficial for compile times to split declarations from definitions to create more parallelism. We're doing that here and there but it's rather up to the individual developer. As with the code formatting this is an effect of not having the discussion about coding style.

I've seen quite some time spent while linking. Using the LLVM linker didn't improve things too dramatically. It seems that this is also greatly improved with these changes from what I gathered from the images above (how did you generate those btw?).

corepointer:
The CUDA kernels don't have so much impact as there are fewer of them, but they can also benefit of this imho. In general I'm positive towards more granularity in compilation and reduce the weight of libAllKernels.

corepointer:
I don't want to hijack the discussion here but take the opportunity to raise attention to #474 where I never received feedback. Running ``build.sh`` from scratch does not only take a lot of time and resources for building but also the git checkout of the LLVM mono repo is a matter of a few minutes. Using a snapshot (which I'm mostly doing as the feature is implemented, but I didn't want to change defaults without discussion) speeds this up considerably. :speedboat:

philipportner:
> Nice progress with this issue 👌
> 
> In general it is more beneficial for compile times to split declarations from definitions to create more parallelism. We're doing that here and there but it's rather up to the individual developer. As with the code formatting this is an effect of not having the discussion about coding style.
> 
> I've seen quite some time spent while linking. Using the LLVM linker didn't improve things too dramatically. It seems that this is also greatly improved with these changes from what I gathered from the images above (how did you generate those btw?).

Mhm from what I've seen linking shouldn't take too long, are you sure `cmake` is using `lld`? How did you measure that?

The images are generated by parsing the `build/.ninja_log` file with `ninjatracing` and viewing [here](http://ui.perfetto.dev). See #516 for more.

philipportner:
`lld` is linking Clang 19 (1.56 GiB) and  Chromium 124 (1.35 GiB) in ~5 seconds, can't imagine it being a huge part in DAPHNE. Also not sure what other options we have besides [mold](https://github.com/rui314/mold) which is from the same person as `lld`.

corepointer:
> > Nice progress with this issue 👌
> > In general it is more beneficial for compile times to split declarations from definitions to create more parallelism. We're doing that here and there but it's rather up to the individual developer. As with the code formatting this is an effect of not having the discussion about coding style.
> > I've seen quite some time spent while linking. Using the LLVM linker didn't improve things too dramatically. It seems that this is also greatly improved with these changes from what I gathered from the images above (how did you generate those btw?).

I was relying on the [configuration in our main CMakeLists to use LLD](https://github.com/daphne-eu/daphne/blob/73bf668d85bfa572d77a7e86fd75d47b874b59fa/CMakeLists.txt#L148-L153). I didn't do a study on it. I just didn't see a lot of speedup after this was introduced and also, at the end of the compilation process there's a single core on 100% for minutes. I thought LLD can do some some magic there without further looking into it :see_no_evil: 

> 
> Mhm from what I've seen linking shouldn't take too long, are you sure `cmake` is using `lld`? How did you measure that?
> 
> The images are generated by parsing the `build/.ninja_log` file with `ninjatracing` and viewing [here](http://ui.perfetto.dev). See #516 for more.

:+1:

philipportner:
> > > Nice progress with this issue 👌
> > > In general it is more beneficial for compile times to split declarations from definitions to create more parallelism. We're doing that here and there but it's rather up to the individual developer. As with the code formatting this is an effect of not having the discussion about coding style.
> > > I've seen quite some time spent while linking. Using the LLVM linker didn't improve things too dramatically. It seems that this is also greatly improved with these changes from what I gathered from the images above (how did you generate those btw?).
> 
> I was relying on the [configuration in our main CMakeLists to use LLD](https://github.com/daphne-eu/daphne/blob/73bf668d85bfa572d77a7e86fd75d47b874b59fa/CMakeLists.txt#L148-L153). I didn't do a study on it. I just didn't see a lot of speedup after this was introduced and also, at the end of the compilation process there's a single core on 100% for minutes. I thought LLD can do some some magic there without further looking into it 🙈
> 
> > Mhm from what I've seen linking shouldn't take too long, are you sure `cmake` is using `lld`? How did you measure that?
> > The images are generated by parsing the `build/.ninja_log` file with `ninjatracing` and viewing [here](http://ui.perfetto.dev). See #516 for more.
> 
> 👍

There should be an early message output saying `Using lld`. Otherwise I'm not really sure, I don't seem to have that problem :')

corepointer:
> > I was relying on the [configuration in our main CMakeLists to use LLD](https://github.com/daphne-eu/daphne/blob/73bf668d85bfa572d77a7e86fd75d47b874b59fa/CMakeLists.txt#L148-L153). I didn't do a study on it. I just didn't see a lot of speedup after this was introduced and also, at the end of the compilation process there's a single core on 100% for minutes. I thought LLD can do some some magic there without further looking into it 🙈
> 
> There should be an early message output saying `Using lld`. Otherwise I'm not really sure, I don't seem to have that problem :')

I guess this does not work as it is saying both ``-- Using lld`` and before that ``-- Linker detection: GNU ld`` and link times stay slow.

corepointer:
I just tested these changes and can report an improvement from 2m47 -> 1m17 :1st_place_medal: :tada: :rocket:

philipportner:
> Results from running in the container: Build reports success but libAllKernels.so is not there. Therefore, running anything fails.

That should only have happened if you didn't pull updates in this PR as this has been fixed in 6532eaa9651c30e61a30e913761b16ac891b67a3 . Could that be the case? For GPU to work you'd need the up to date PR.

src/runtime/local/kernels/CMakeLists.txt
```diff
@@ -21,10 +21,17 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
 # The library of pre-compiled CUDA kernels
 if(USE_CUDA AND CMAKE_CUDA_COMPILER)
+
+    set(CUDA_CODEGEN_CPP_FILES)
+    foreach(i RANGE 0 20)
```
corepointer, line 26: This looks like a hard coded value that will cause trouble in the future. When does this need adjusting? Maybe document in a comment above this statement.

src/runtime/local/kernels/CMakeLists.txt
```diff
@@ -21,10 +21,17 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
...
+        list(APPEND CUDA_CODEGEN_CPP_FILES ${PROJECT_BINARY_DIR}/src/runtime/local/kernels/CUDAkernels_${i}.cpp)
```
corepointer, line 27: I don't know why it had to be CUDAkernels_${i}.cpp - the filename you had there before did not work but this did.

src/runtime/local/kernels/CMakeLists.txt
```diff
@@ -74,9 +81,15 @@ if(USE_CUDA AND CMAKE_CUDA_COMPILER)
     set_target_properties(CUDAKernels PROPERTIES LIBRARY_OUTPUT_DIRECTORY ${PROJECT_SOURCE_DIR}/lib)
 endif()
 
+set(CODEGEN_CPP_FILES)
+foreach(i RANGE 0 104)
```
corepointer, line 85: Hard coded value - see comment above in line 27

corepointer (changes requested):
Tests are running fine now with and without  CUDA. I'd merge it in once the mystery of 20 and 104 is resolved.

corepointer:
Oh the PR is still in draft state. Further changes planned for this?

philipportner:
> Oh the PR is still in draft state. Further changes planned for this?

Actually it's no longer in draft, maybe I shouldn't put that in the title :)

src/runtime/local/kernels/CMakeLists.txt
philipportner: That's simply the amount of files (kernels) that are generated and would need to be incremented for each new kernel. I didn't find any better way to do that as the amount of kernels is not known until the script is run. The script runs after the point where CMake needs to know about those files.

If we don't find any better way this has to be documented and added to the steps to add a new kernel. Maybe you have an idea? My CMake knowledge is admittedly somewhat limited :)

philipportner:
> Tests are running fine now with and without CUDA. I'd merge it in once the mystery of 20 and 104 is resolved.

Thanks a lot for testing this with CUDA @corepointer !

src/runtime/local/kernels/CMakeLists.txt
corepointer: We could try to write the number of output files to a text file in the buld directory and [read that in cmake](https://cmake.org/cmake/help/v3.14/command/file.html)?

corepointer:
> > Oh the PR is still in draft state. Further changes planned for this?
> 
> Actually it's no longer in draft, maybe I shouldn't put that in the title :)

my bad to jump to conclusions :see_no_evil:

src/runtime/local/kernels/CMakeLists.txt
philipportner: Wouldn't that just move the hard-coded number to somewhere else?

Maybe we should replace the `add_custom_command` call with an `execute_process`, [here](https://cmake.org/cmake/help/latest/command/execute_process.html).

Maybe that already solves the problem cause it runs at configuration time, so before build time.
If not, I think it should be possible to return the list of files created by the script and use `RESULT_VARIABLE` to set the `CODEGEN_CPP_FILES` with that output.

Either way, just speculation and I'll have to try that first.

corepointer: > Wouldn't that just move the hard-coded number to somewhere else?
I didn't think that through :blush:  We need the number for the add_custom_command. So your method with execute_process might be a better idea. Another (less elegant) solution would be to run that custom command once in a "dry run" mode that only determines the number of produced files.

philipportner: > > I didn't think that through 😊  We need the number for the add_custom_command. So your method with execute_process might be a better idea. Another (less elegant) solution would be to run that custom command once in a "dry run" mode that only determines the number of produced files.

Similar idea: 

What do you think about moving the python execution out of `cmake` and into to `build.sh`? That would massively simplify the `CMakeLists.txt`. The files would be generated in the `build/.../kernels/` path before `cmake` is run so we can simply do

```
file(GLOB sources CONFIGURE_DEPENDS "${path}/*.cpp")
```
Would have to double check if this works properly when recompiling single kernels, but that seems like a very simple solution. I'm not sure how many use the `cmake --build` manually, but proper documentation and a note in the `build.sh` should be enough to avoid any problem in such cases.

philipportner: Meaning you don't like that idea? I also run the `cmake` commands manually but wouldn't have a problem with running `python` manually when adding a kernel, but I'm also fine with not splitting that up. I'll try the `execute_process` tomorrow, maybe that solves that problem anyways.

corepointer: In the context of building DAPHNE I'd say the more automation the better. The cmake docs say that ``execut_process`` runs at config time and ``add_custom_command`` at build time. So setting up the file list with the former to consume in the latter should be doable. You could even bake in that cmake for loop in your favorite scripting language. The file list generation is already there in the main function of ``genKernelInst.py``

corepointer:
LGTM - ran the test suite once again locally and fine tuned the commit message a bit while merging ;-)

Thx for your efforts @philipportner

Issue #822, opened 2024-09-10
Saving sparse matrices to file (MM write support)


corepointer:
DAPHNE has support for reading Matrix Market [1] files. This is implemented in [readMM.h](https://github.com/daphne-eu/daphne/blob/main/src/runtime/local/io/ReadMM.h) and enables reading sparse data from MM/MTX files. 

Unfortunately, there is [no write support for sparse matrices yet](https://github.com/daphne-eu/daphne/blob/main/src/runtime/local/kernels/Write.h). Since there is read support for MM, it would make sense to also have the functionality to write sparse data to disk in the MM format. 

This feature is also part of a student project description in #517 but can be dealt with in this separate issue as half of the task (read **and** write support of more file formats) is already there. Furthermore, this feature has already been asked for, so speeding up development in this regard is desirable. 

[1] https://math.nist.gov/MatrixMarket/formats.html

ldirry:
Hi, please assign me

corepointer:
as you wish :bow:

PR #832, opened 2024-09-25
[DAPHNE-#822] Added MatrixMarket write support

ldirry:
- Write support for CSRMatrix in coordinate system
- Write support for DenseMatrix in array format
- Support for recognizing symmetric, skew-symmetric matrices, and storing them accordingly in .mtx
- Added testcases, calling readMM, then writeMM and then comparing if the files are still the same


src/runtime/local/io/WriteMM.h
```diff
@@ -0,0 +1,239 @@
...
+        if (!values) {
+            throw std::runtime_error("WriteMM::apply: Null pointer for 'values' in DenseMatrix");
+        }
+
+        if (strcmp(symmetry, MM_GENERAL_STR) == 0) {
+            for (size_t i = 0; i < cols; ++i) {
+                for (size_t j = 0; j < rows; ++j) {
+                    size_t idx = j * cols + i;
+                    if (strcmp(field, MM_REAL_STR) == 0)
+                        f << std::scientific << std::setprecision(13) << values[idx] << std::endl;
```
corepointer, line 68: Would this value be a candidate for an item in our user configuration?

src/runtime/local/io/WriteMM.h
```diff
@@ -0,0 +1,239 @@
...
+        for (size_t i = 0; i < rows; ++i) {
+            for (size_t idx = rowOffsets[i]; idx < rowOffsets[i + 1]; ++idx) {
+                size_t j = colIdxs[idx];
+                colEntries[j].emplace_back(i, values[idx]);
+            }
+        }
+
+        if (strcmp(symmetry, MM_GENERAL_STR) == 0) {
+            nnz = arg->getNumNonZeros();
+        } else {
```
corepointer, line 231: I think I don't fully understand why the computation above is needed if  you return getNumNonZeros or fail.

test/api/cli/io/WriteTest.cpp
```diff
@@ -21,9 +21,49 @@
...
 
 #include <filesystem>
+#include <fstream>
+#include <iostream>
 #include <string>
 
 const std::string dirPath = "test/api/cli/io/";
+const std::string dirPath2 = "test/runtime/local/io/";
+
+bool compareFiles(const std::string &filePath1, const std::string &filePath2) {
```
corepointer, line 31: Maybe this one could go into Utils.h? Ideally with a documenting comment (doxygen format) above the function signature ;-)

corepointer (commented):
Generally, this is looking good. Maybe in addition to the things I mentioned in the inline comments to the code, you could separate your code into header and cpp file?

Issue #833, opened 2024-09-26
run clang-format on push / PR


philipportner:
With the new `.clang-format` file we should make sure that new code also adheres to the coding style defined by `clang-format`. I don't think we should add something that automatically enforces formatting for every user, e.g., pre-commit hooks. Rather, I would suggest to add a CI action that simply fails when the code in the commit/PR does not follow our code style.

This can easily be done with https://github.com/marketplace/actions/clang-format-check

philipportner:
closed with #835

PR #835, opened 2024-09-28
[ci] adds clang-format action (#833)

philipportner:
With 12839d6d, cb999f20, and 9a1b93e4 a `.clang-format` file was introduced and
`src/` and `test/` were formatted according to the `.clang-format` file using
`clang-format` version `18.1.8`.

This patch adds an action that runs clang-format on pushes to main and on events
in pull requests. The action runs `clang-format` with version `18.1.8` on
the `src/` and `test/` folders. If the pushed code contains ill-formatted code,
the actions displays the diff.

With that, I added `clang-format-18` to the dependencies of the Dockerfile to make 
it readily available for contributors.


containers/daphne-dev.Dockerfile
```diff
@@ -34,7 +34,16 @@ RUN apt-get -qq -y update && apt-get -y upgrade && apt-get -y --no-install-recom
...
+    lsb-release gnupg \
     && apt-get clean && rm -rf /var/lib/apt/lists/*
+
+RUN set -xe && wget https://apt.llvm.org/llvm.sh && \
+    chmod +x llvm.sh && \
+    ./llvm.sh 18
+
+RUN set -xe &&  \
+    apt-get -y --no-install-recommends install clang-format-18
+
```
corepointer, line 46: Maybe we should add this method of installation of llvm/clang to the documentation section ?

src/runtime/local/kernels/Append.h
```diff
@@ -17,10 +17,10 @@
 #pragma once
 
 #include <runtime/local/context/DaphneContext.h>
+#include <runtime/local/datastructures/CSRMatrix.h>
```
corepointer, line 20: Are these header changes necessary for the clang-format update? If not I'd wrap them into a separate commit while merging (if I knew the cause for the change).

.github/workflows/clang-format-check.yml
```diff
@@ -0,0 +1,20 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: DoozyX/clang-format-lint-action@v0.18.1
+      with:
+        source: 'src test'
+        exclude: './third_party ./external'
+        extensions: 'h,cpp'
+        clangFormatVersion: 18.1.8
```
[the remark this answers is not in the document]
philipportner: The thing is I wasn't aware that something changed between `18.1.3` and newer versions when I formatted the code in cb999f20b6f2934ad7c94b10d2b02f6acf74aab4 and 9a1b93e4fea27e91d14c18821be2c940adf63bf4 . But thinking about it again, you're right. I'll make it easier for us and just reformat again with `18.1.3`, already tried that locally and very it's not that bad (20 files or so).

Than we can simply `apt install clang-format` and not have this `llvm.sh` script at all.

containers/daphne-dev.Dockerfile
philipportner: See above, I'll remove this.

src/runtime/local/kernels/Append.h
philipportner: Yes, `clang-format` sorts the includes in a way and these few files have had a few includes added after I formatted the codebase and I apparently did not `clang-format` it myself :)

philipportner:
> LGTM generally. Maybe resolve the inline questions :)

Thanks a lot @corepointer for the instant review :)

.github/workflows/clang-format-check.yml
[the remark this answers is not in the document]
corepointer: There's even a clang-format-18 package. But that's the default anyway. While we're at it: I submitted my request to change the line length in a separate PR #839 as this might cause a lot of formatting change :see_no_evil:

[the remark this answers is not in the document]
philipportner: > There's even a clang-format-18 package. But that's the default anyway. While we're at it: I submitted my request to change the line length in a separate PR #839 as this might cause a lot of formatting change 🙈

Just saw that, it's fine ;) I'll handle the line length and `18.1.3` changes.

auge:
@philipportner
sorry for my late comments … I highly value the move towards more maintain- and readable code.
I have a small suggestion, could you pls. update the checkout step to a more modern version? We've recentely updated these in the other actions too - see https://github.com/daphne-eu/daphne/blob/main/.github/workflows/main.yml

`uses: actions/checkout@v4`

philipportner:
> @philipportner sorry for my late comments … I highly value the move towards more maintain- and readable code. I have a small suggestion, could you pls. update the checkout step to a more modern version? We've recently updated these in the other actions too - see https://github.com/daphne-eu/daphne/blob/main/.github/workflows/main.yml
> 
> `uses: actions/checkout@v4`

Hi @auge , I totally missed that. Thanks for the suggestion! #841 and I'll push an update for that tomorrow :)

Issue #696, opened 2024-04-19
Cost-aware Function Specialization


pdamme:
**Motivation:** DaphneDSL supports user-defined functions with typed and untyped parameters and results. DAPHNE's MLIR-based optimizing compiler automatically infers the types and properties (e.g., shape, sparsity) of intermediate results in a DaphneDSL script and uses this information for various optimizations. Function calls pose a special challenge as a function must be generic, but should also benefit from information on the inputs to further optimize the code. To fully exploit this information inside functions, DAPHNE currently creates a separate copy of each function per call-site and specializes it with the information available on the parameters, which can be quite costly.

**Task:** Implement (C++) a pass in DAPHNE's MLIR-based optimizer, which automatically decides in which cases it makes
sense to specialize a function, such that a balance between code size and improved runtime performance is achieved.

**Hints:** *coming soon*

PR #836, opened 2024-09-30
[DAPHNE–#696] Cost-aware function specialization

XejgaToast:
Batur (Exion007) + Ahmad (XejgaToast)

## Naming convention
- Added a new naming convention of specialized Functions (more readable IR)
- Added a duplicate specialization prevention (no duplicate specializations anymore) depending on the naming convention

## Recursion Detection/Prevention in specialization
- Added a call graph to keep track of recursion
- Added a recursion detection and prevention with an hard coded upper limit of 5 recursive calls per recursive cycle to prevent too many specializations

## Removal of non-beneficial specializations
- Added a hash map to keep track of function bodies and compare them to find potentially equal functions with different names
- Delete specialized functions which already have an equal function and replace callee attributes to point to existing function

## Documentation
- Added documentation for the Specialization Pass itself
- Adjusted documentation for a couple of functions


pdamme:
Thanks @XejgaToast and @Exion007 for this contribution. Solving the task described in #696 is very welcome. I will make an initial pass over this PR soon.

Exion007:
## Added new test script

This test script tests 3 aspects simultaneously by comparing the IR with the expected IR output (specializeIRTest.txt), which are the following: 

- Similar Function detection and prevention (function c is expected to be replaced with function  b)
- Recursion detection and prevention (function a is expected to only have 5 recursive calls, meaning 6 specializations)
- Naming Convention (All function names in the IR should follow the naming convention)

PR #840, opened 2024-10-01
[DAPHNE-#840] Switch to C++20 in CMake config

corepointer:
To accommodate various new language features, we switch the used C++ version to C++20.

Moving this change to a separate PR as requested.


philipportner (approved):
Thanks, LGTM :)

PR #842, opened 2024-10-03
Split fmt and spdlog dependencies

corepointer:
In order for nvcc (CUDA compiler) to not fail the DAPHNE build after switching to C++20, we introduce fmt as a separate dependency and upgrade the spdlog to the tip of the spdlog default branch as of 2024-10-03 for the time being (we'll switch back to a release once that compiles without troubles) because the released spdlog 1.14.1 did not compile with the newer fmt 11.0.2.


corepointer:
The CI build failure seems to be unrelated and the tests run fine on my laptop and the scale-up box. I'll proceed with the merge no not cause any further troubles (see #843).

corepointer:
Updating the docker images right away to accommodate this change. At least for x86-64. ARM folks please get in touch if you run into troubles that you can not resolve on your own :see_no_evil:

philipportner:
> Updating the docker images right away to accommodate this change. At least for x86-64. ARM folks please get in touch if you run into troubles that you can not resolve on your own 🙈

Hi @corepointer , we have a user that runs into this problem with `latest_ARMV8_BASE`. Any way you could update the container for ARM?

corepointer:
I don't have a working ARM development environment at hand to do this quickly (my ARM VM from the OCI always-free tier that I used for this task is occupied with something else atm). Maybe that user of yours can try running the container build scripts themselves (put an "exit" into the container build script after the BASE dev container and before the CUDA stuff).

PR #864, opened 2024-10-17
[kernels] fix usage of C stdlib abs (#863)

### LINKED ISSUE #863: abs() casts to integer
abs() casts to integer although it still returns double matrix

CODE:

m = as.matrix([1.11,-2.22,-3.312,4.123123]);
print(m);
n = abs(m);
print(n);

STDOUT:

DenseMatrix(4x1, double)
1.11
-2.22
-3.312
4.12312
DenseMatrix(4x1, double)
1
2
3
4

EXPECTED:

DenseMatrix(4x1, double)
1.11
2.22
3.312
4.12312

philipportner:
The EwUnary kernel used the C stdlib `abs` function which instead of the appropriate C++ `std::abs` function. This caused a bug where calling `abs` with floating point values were cast to integer values during conversion.

Closes #863

### DIFF
--- a/src/runtime/local/kernels/EwUnarySca.h
+++ b/src/runtime/local/kernels/EwUnarySca.h
@@ -154,7 +154,7 @@ template <typename TRes, typename TArg> TRes ewUnarySca(UnaryOpCode opCode, TArg
 // One such line for each unary function to support.
 // Arithmetic/general math.
 MAKE_EW_UNARY_SCA(UnaryOpCode::MINUS, -arg);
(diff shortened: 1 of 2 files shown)


auge:
would it make sense to also add a test case when we spot a bug like this?

philipportner:
> would it make sense to also add a test case when we spot a bug like this?

yes ofc! I already have them locally but forgot to push them with the PR and just want to let the CI run finish before I push again :)

m-birke:
I tested it, works fine

PR #867, opened 2024-10-18
[DAPHNE-#867] Change ConfigParser::fileExists() behavior to not throw

corepointer:
I consider this a bugfix but I'll run it through a PR to raise attention to the change. Tests are running fine (after fixing the one that tests the previous behavior). If there were no concerns raised by the time CI ran through I'll merge it right in.

Changing the behavior of fileExists() to a boolean operation as suggested by the method's name. Throwing an exception us up to the caller of this method.

Closes #867

### DIFF
--- a/src/parser/config/ConfigParser.cpp
+++ b/src/parser/config/ConfigParser.cpp
@@ -31,8 +31,10 @@ int readLogLevel(const std::string &level) {
 bool ConfigParser::fileExists(const std::string &filename) {
     // Open the given config file.
     std::ifstream ifs(filename, std::ios::in);
-    if (!ifs.good())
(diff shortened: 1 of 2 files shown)


corepointer:
The same semantic problem exists with keyExists() imho. Bit I'll leave that one for future work as my main concern to change this in the first place was a failing distributed test that did not find its config file.

philipportner (approved):
Not sure if you wanted to close a related issue but you put the PR# itself in the description :)

corepointer:
> Not sure if you wanted to close a related issue but you put the PR# itself in the description :)

Since issues and pull requests share the same sequence number, I frequently abuse this to have a ticket number and a place for discussion right here in the PR.

Issue #881, opened 2024-10-25
VectorizeComputation pass fails when iterating over uses of old results


philipportner:
```
bin/daphne --vec test/api/cli/vectorized/pipeline_2.daphne
==================
WARNING: ThreadSanitizer: heap-use-after-free (pid=3006835)
  Read of size 8 at 0x7b20000035e0 by main thread:
    #0 mlir::ValueUseIterator<mlir::OpOperand>::operator++() /home/philipportner/daphne/thirdparty/installed/include/mlir/IR/UseDefLists.h:269 (daphne+0x1c8a803)
    #1 runOnOperation /home/philipportner/daphne/src/compiler/lowering/VectorizeComputationsPass.cpp:378 (daphne+0x1c8a803)
    #2 mlir::detail::OpToOpPassAdaptor::run(mlir::Pass*, mlir::Operation*, mlir::AnalysisManager, bool, unsigned int) <null> (daphne+0x23b460e)
    #3 startDAPHNE(int, char const**, DaphneLibResult*, int*, DaphneUserConfig&) /home/philipportner/daphne/src/api/internal/daphne_internal.cpp:590 (daphne+0x17ce256)
    #4 mainInternal(int, char const**, DaphneLibResult*) /home/philipportner/daphne/src/api/internal/daphne_internal.cpp:668 (daphne+0x17d5bee)
    #5 main /home/philipportner/daphne/src/api/cli/daphne.cpp:19 (daphne+0x17882a7)

  Previous write of size 8 at 0x7b20000035e0 by main thread:
    #0 free ../../../../src/libsanitizer/tsan/tsan_interceptors_posix.cpp:711 (libtsan.so.0+0x37ab8)
    #1 llvm::iplist_impl<llvm::simple_ilist<mlir::Operation>, llvm::ilist_traits<mlir::Operation> >::erase(llvm::ilist_iterator<llvm::ilist_detail::node_options<mlir::Operation, true, false, void>, false, false>) <null> (daphne+0x25a5736)
    #2 mlir::detail::OpToOpPassAdaptor::run(mlir::Pass*, mlir::Operation*, mlir::AnalysisManager, bool, unsigned int) <null> (daphne+0x23b460e)
    #3 startDAPHNE(int, char const**, DaphneLibResult*, int*, DaphneUserConfig&) /home/philipportner/daphne/src/api/internal/daphne_internal.cpp:590 (daphne+0x17ce256)
[...]

PR #882, opened 2024-10-25
[daphneir] fix null value dereference (#881)

philipportner:
VectorizeComputation pass fails when iterating over uses of old results (#881).


philipportner:
closes #881
closes #884

PR #920, opened 2024-11-19
Broadcasting for ewBinary codegen

AlexRTer:
This PR changes the existing ewBinaryOp codegen to broadcast `rhs` if possible.
So far, the ewBinary codegen only accepted scalars, matrices of equal shape or combinations of `lhs` being a matrix and `rhs` being either a scalar or singleton.
Now `rhs` can also be given as a matching (equal to `lhs` in one dimension and size `1` in the other) row or column vector.

An example script (--explain mlir_codegen is optional to show the generated IR):
```
// RUN: ./bin/daphne --select-matrix-repr --mlir-codegen --explain mlir_codegen ./fileName.daphne

lhsMat = [1, 2, 3, 4, 5, 6](2,);

rhsMat = [10, 20, 30, 40, 50, 60](2,);
rhsRowMat = [10, 20, 30](1,);
rhsColMat = [10, 20](,1);
rhsScalar = 10;
rhsSingleton = [10];

print(lhsMat + rhsMat);
print(lhsMat + rhsRowMat);
print(lhsMat + rhsColMat);
print(lhsMat + rhsScalar);
print(lhsMat + rhsSingleton); // not supported outside of codegen yet
```

A more thorough description will be given once some tests have been added.

### DIFF
--- a/src/compiler/lowering/EwOpsLowering.cpp
+++ b/src/compiler/lowering/EwOpsLowering.cpp
@@ -31,6 +31,7 @@
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/Math/IR/Math.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/IR/AffineMap.h"
 #include "mlir/IR/Builders.h"
 #include "mlir/IR/BuiltinDialect.h"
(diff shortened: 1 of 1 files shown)


philipportner (approved):
Thanks for the PR, I'll merge this in as we want to have it in main. Please create an additional PR with any outstanding changes.

Issue #903, opened 2024-11-12
`group()` built-in function in DaphneDSL


pdamme:
DaphneIR has a `GroupOp` for relational-style grouping and aggregation, plus a corresponding `group`-kernel. However, at the moment, this operation can only be created through DAPHNE's SQL parser, but not through DaphneDSL.

This task is to add a `group()` built-in function to DaphneDSL that creates a `GroupOp` in DaphneIR. The interface should be (following the notation used in the [docs](https://daphne-eu.github.io/daphne/DaphneDSL/Builtins/): `group(arg:frame, groupCols:str, ..., sumCol:str)`. That is, the built-in function gets a frame, an arbitrary number of columns to group on, and a single column to calculate the sum on. While this interface does not allow to use all features of the `GroupOp`/`group`-kernel (e.g., multiple aggregates, other aggregate functions than sum), it would be a good first step and sufficient for implementing the Star Schema Benchmark in DaphneDSL. We can reflect the full functionality of the `GroupOp` is DaphneDSL later.

**Hints:**
- Add the new built-in function in the DaphneDSL parser.
- Add a few script-level test cases for the `group` built-in function.

saminbassiri:
Hi, I will work on this issue.

pdamme:
Great, please go ahead!

PR #921, opened 2024-11-20
Add 'group()' built-in function to DaphneDSL.

saminbassiri:
#### Description
This pull request introduces a new `group()` built-in function to DaphneDSL, enabling the creation of a `GroupOp` in DaphneIR and closes issue #903 .

---

#### Changes Implemented
1. **`group()` Built-in Function in DaphneDSL**:
   - Interface: `group(arg:frame, groupCols:str, ..., sumCol:str)`
   - Accepts:
     - A frame as input.
     - An arbitrary number of columns to group on.
     - A single column to compute the sum.
   - **Aggregation Support**:
     - Only supports `SUM` as the aggregation function.

2. **Kernel Function Updates**:
   - Updated `order` and `extractCol` kernel functions to process string values correctly.
   - Extended `group` kernel function to handle string values.

3. **Test Cases**:
   - Added script-level tests to validate the functionality of the `group()` function in DaphneDSL.


pdamme (approved):
Thanks for this contribution, @saminbassiri. Overall, **your code looks good to me**. I've only made some smaller changes, e.g., I updated the DaphneDSL docs and changed the parameter order of the new `groupSum()` built-in to match #903. To focus on #903, I also simplified the new test case not to use string columns. Consistently, I also reverted the changes to the kernels, because they are not relevant for #903. We will add these changes in a separate commit later.

pdamme:
I committed the changes not directly related to the topic of this PR in commit a3748141adc98d0544416492b04c48feb3b31c24.

PR #926, opened 2024-11-29
Add hash join and more support for strings value types in kernel functions

saminbassiri:
### Description:
This pull request introduces enhancements to the system by adding a new `hashJoin` operation and extending support for string value types in kernel functions.

### Key Changes:
1. **Hash Join Implementation**:
   - Added `hashJoin` as a kernel function.
   - Integrated `hashJoin` as a built-in function in DaphneDSL.
   - Created `HashJoinOp` in DaphneIR.

2. **String Value Support**:
   - Enhanced kernel functions to handle string values.
   - Updated `FilterRow` kernel function for compatibility with string values.
   - Introduced `EwBinary` operations to compare `char*` strings with `std::string` .
   - Updated `kernel.json` to reflect the new string value support.

### Next Steps:
Add tests to validate correct behavior across various scenarios.


pdamme:
I committed the changes not directly related to the topic of this PR in commit a3748141adc98d0544416492b04c48feb3b31c24.

pdamme (approved):
Thanks for this contribution @saminbassiri. A hash-based join implementation was long overdue in DAPHNE. You added the hash-join throughout the DAPHNE stack (DaphneDSL built-in function, DaphneIR op plus inference, runtime kernel).

While this integration looks good from a technical point of view, I think we should not expose the physical join variant to users. Instead, it's much more useful to replace the existing innerJoin-kernel (which was very inefficient and had memory leaks) by your hash-based implementation. Consequently, I removed the new `hashJoin()` built-in and `HashJoinOp`.

I fixed a few little things in the hash-join code: memory leaks related to `Frame::getColumn()` and made the kernel support n:m-joins (by using a `std::unordered_map<VT, std::vector<size_t>>` for the hash table). See my commits for details.

With that, **this PR is ready to be merged**.

Issue #693, opened 2024-04-18
Codegen-only compilation pipeline for LA operations


philipportner:
**Motivation**: By default, DAPHNE lowers domain-specific operations from linear and relational algebra to calls to pre-
compiled C++ kernels (physical operators). However, being based on MLIR/LLVM, on-the-fly code generation for these operations is an alternative option.

**Task**: Implement a proof-of-concept for a compilation pipeline in DAPHNE, which only involves code-generated
operations. For simplicity, you can focus on linear algebra operations, whereby dense and sparse matrices should be
supported. The resulting code may target x86 CPUs or, optionally CUDA-enabled GPUs.

To start, take DaphneDSL of an example (e.g., from `scripts/`) and save the generated DaphneIR, or write an example DaphneIR file by hand. Use the `daphne-opt` tool to quickly iterate on your lowering passes and the whole lowering pipeline. Once this works for a set of scripts, bring the alternative compilation pipeline to the DAPHNE compiler through an optional command-line argument.

PR #935, opened 2024-12-21
[LDE: Project] 'Codegen-only compilation pipeline for LA operations' 

WangYuyao:
This is an implementation of [Codegen-only compilation pipeline for LA operations #693](https://github.com/daphne-eu/daphne/issues/693) as the coursework for 'LDE: Project'

### DIFF
--- a/containers/entrypoint-interactive.sh
+++ b/containers/entrypoint-interactive.sh
@@ -13,10 +13,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+USER=ubuntu
 /usr/sbin/sshd -f /etc/ssh/sshd_config
-/usr/sbin/groupadd -g "$GID" dockerusers
-/usr/sbin/useradd -c 'Docker Container User' -u $UID -g "$GID" -G sudo -m -s /bin/bash -d /home/"$USER" "$USER"
+# /usr/sbin/groupadd -g "$GID" dockerusers
+# /usr/sbin/useradd -c 'Docker Container User' -u $UID -g "$GID" -G sudo -m -s /bin/bash -d /home/"$USER" "$USER"
 printf "${USER} ALL=(ALL:ALL) NOPASSWD:ALL" | sudo EDITOR="tee -a" visudo #>> /dev/null
 mkdir -p /home/"$USER"/.ssh
 chmod 700 /home/"$USER"/.ssh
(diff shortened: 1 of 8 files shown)


containers/entrypoint-interactive.sh
philipportner, line 1: Remove these changes.

containers/run-docker-example.sh
philipportner, line 1: Remove these changes.

scripts/examples/slice.daph
philipportner, line 1: Remove this file and add proper testcases. Look at the testcases in `test/codegen/` and `test/api/cli/codegen`, there you will find different kind of tests for our codegen. `FileCheck` based tests that verify the IR after a certain pass(es), and end-to-end tests that execute `daphne` and verify the output. You should add a single file, e.g., `test/codegen/sliceop.mlir` for the IR tests and a single `.cpp` file for the end-to-end tests that compare the non codegen with the codegen'd execution of `daphne`.

src/compiler/execution/DaphneIrExecutor.cpp
philipportner, line 1: Combine these into a single pass which has two rewrite patterns.

src/compiler/lowering/SliceColOpLowering.cpp
```diff
@@ -0,0 +1,178 @@
...
+                                            AffineMap::getMultiDimIdentityMap(2, rewriter.getContext())};
+
+        SmallVector<utils::IteratorType, 2> iterTypes{utils::IteratorType::parallel,
+                                                      utils::IteratorType::parallel};
+
+        rewriter.create<linalg::GenericOp>(loc, TypeRange{}, ValueRange{selMemref}, ValueRange{resMemref},
+                                           indexMaps, iterTypes,
+                                           [&](OpBuilder &OpBuilderNested, Location locNested, ValueRange arg) {
+                                               OpBuilderNested.create<linalg::YieldOp>(locNested, arg[0]);
+                                           });
```
philipportner, line 115: This shouldn't be needed, as we want to create a view of a matrix we don't want to copy the data. The `memref::SubViewOp` should be enough. You'll need to properly lower it afterwards by adding the following passes to the pipeline:
`mlir::memref::createExpandStridedMetadataPass;` and `mlir::createFinalizeMemRefToLLVMConversionPass`.

philipportner (changes requested):
Thanks for the PR.
This already looks promising! A few changes are needed and we should add test cases before merging this in.

Issue #203, opened 2022-03-01
Elementwise binary operations with matrix and scalar


daphne-eu:
In GitLab by @pdamme on Mar 1, 2022, 19:29

In linear algebra programs, elementwise binary operations where one input is a matrix and the other input is a scalar are commonplace. When mixing matrices and scalars, currently the DAPHNE system only supports a matrix on the left-hand side and a scalar on the right-hand side, whereby the user must take care to adhere to this order when writing algorithms in DaphneDSL, DAPHNE's domain-specific language. This requirement shall be relaxed.

The task is twofold: On the one hand, we need canonicalization rewrites for DAPHNE's EwBinaryOp which swap the left-hand side and right-hand side input to ensure matrix-scalar operations, for commutative operations. On the other hand, non-commutative operations shall be handled by a dedicated kernel for scalar-matrix operations (which shall be an adaptation of the existing matrix-scalar kernel). Optionally, non-commutative operations may also be handled by a compiler pass adapting the operands and the operation, e.g., `1 – mat` to `mat + -1`, but note that this is not possible for all operations.

RetoKrummenacher:
Given the fact that left scalar multiplication is not supported yet, the example in the [DaphneDSL documentation](https://github.com/daphne-eu/daphne/blob/118-doc-daphnedsl/doc/DaphneDSL.md) should be changed from ```1.5 * X @ y + 0.001``` to ```X @ y * 1.5 + 0.001``` as the first is currently not working with ```X``` being a matrix and ```y``` a column vector.

pdamme:
Commit 1033729696a362739f56cdba8d774ca70573318c just introduced scalar-matrix ops for `+`, `-`, `*`, `/` via rewrites to matrix-scalar ops that use the existing kernels. The commit message may be worth reading for someone working on this issue. However, this is just a quick fix, the issue needs to be addressed more systematically.

PR #940, opened 2025-01-14
SQL Feature extention for SSB

DerSchmidt:
**This PR adds String Support to the SQL Parser and makes it possible to use BETWEEN. Both features are commonly needed for benchmarks like SSB**

- It is now possible to use Strings in the SQL Parser, for this the LEXER and Parser needed to be adjusted to reflect Strings. Furthermore, the Visitor got extended by a String builder. All the other support gets provided by the DAPHNE infrastructure. Strings as per SQL standard are using single quotes, given that most SQL database servers also support double-quoted string literals, we opted to support them too. Because the sqlOp has a String as an input, we have to escape these double-quotes. 
-- `result = sql("SELECT ... FROM ... WHERE ... AND p.P_CATEGORY = 'MFGR#12' ... ;" );`
-- `result = sql("SELECT ... FROM ... WHERE ... AND p.P_CATEGORY = \"MFGR#12\" ... ;" );`

- Furthermore, we now support BETWEEN. This is a simple implementation to simply enable BETWEEN and not be performant. The Between operation generates four comparisons operations, two and operations and one or operation. It checks both `a <= x <= b` and `b <= x <= a`.

During the implementation of between, it became clear that, the SQL parser is vulnerable to the Issue #203. This PR doesn't resolve this issue but tries to provide some renewed attention to this issue.


pdamme (changes requested):
Thanks for this contribution, @DerSchmidt! **Overall, the code looks good to me.** However, some points need to be addressed:

**Necessary changes:** (need to be addressed before the merge)
1. I don't fully understand why you created a separate `generalExpr2` in the SQL grammar. At a quick glance, it seems to me like the only effect is to avoid nested use of `BETWEEN`. Though nested `BETWEEN` doesn't seem to be very useful, I don't see why we should prohibit expressions like `a BETWEEN 0 AND (3 * CAST((2 BETWEEN 1 AND 3) AS INT))`. At the same time, the parser code seems to get more complicated through the additional `generalExpr2` (e.g., duplication of parantheses expression for `generalExpr` and `generalExpr2`). Can't we treat `BETWEEN` like any other boolean operation? I would just like to clarify that point, if there is a good reason, we can also leave it as it is.
2. The SQL grammer now supports string literals and they can contain escaped special characters. However, the SQL parser (`visitiLiteral()`) does not replace these escaped characters. For instance, `"ab\"cd"` would remain `ab\"cd`, but should become `ab"cd`. Please either add the unescaping to the SQL parser (have a look at how the the DaphneDSL parser does it) or remove the escapes from the SQL grammar.
4. Test cases, `*.daphne` files: Please update the comment line at the top of each file to describe what the test case does.

**Optional changes:** (could be addressed now or after the merge)
1. All your test cases for strings in SQL contain only strings with exactly one character. To be on the safe side, it would be a good idea to also involve some other cases, such as multiple characters, empty strings, strings containing escaped special characters.

-----

Besides that, thanks for pointing out #203 in this context. I saw in your code that you circumvented this problem by creating the elementwise comparison operations with a suitable operand order. #203 is referenced by a commit that fixes the problem for some binary operators. I think similar fixes could be implemented for `>=` and `<=`, e.g., `X >= c` to `not (c < X)`. Feel free to give it a try if you like (but not required for this PR).

DerSchmidt:
Thank you @pdamme,

Concerning 1: 
I needed to add generalExpr2 because otherwise the AND in BETWEEN wouldn't bind strongly enough. 
So, for instance:` ... f.a BETWEEN 1 AND 3 AND f.c < 5` where the second AND is just another filter. ANTLR parsed this, no matter the order in of the BETWEEN and AND rules, as: ` f.a BETWEEN (1 AND 3) AND f.c < 5` which results obviously in a wrong result. The introduction of generalExpr2 fixes this issue, since we don't have any Boolean Operators inside generalExpr2.
Fixing nested BETWEENS is simply a happy accident.
There might be a better solution for this problem, but this was the least invasive step I could think of. 

Concerning 2: 
I applied the code from the DSL Parser onto the SQL Parser. See [f86b9e2](https://github.com/daphne-eu/daphne/pull/940/commits/f86b9e27b3ef6b79c60536f86a26fedfda4c7ea4)

Concerning 3 and the Optional Changes:
I updated the Comments on top of the tests to correctly reflect what is tested. 
For completeness’s sake, two test cases were added for multi character strings. See [2635ae1](https://github.com/daphne-eu/daphne/pull/940/commits/2635ae1df8826037d80e4bb84bee01fc18a54773)

I am now aware of #203, and as soon as I am free I might address it, but as you said, it is outside the scope of this PR.

DerSchmidt:
@pdamme can you maybe have a quick look again?
Thank you very much!

pdamme (approved):
Thanks for addressing all requested required and optional points, @DerSchmidt. **This PR is ready to be merged now.**

Issue #688, opened 2024-04-16
Enable reading files without a meta data file


pdamme:
**Motivation:** When reading a data set from a file, DAPHNE requires a so-called meta data file to be present. This meta data file contains basic information on the data set, such as the number of rows and columns, the value type of a matrix or the column types of a frame, the number of non-zeros, and potentially additional interesting properties. See the [documentation](https://daphne-eu.github.io/daphne/FileMetaDataFormat/) for details. This information is used by the DAPHNE compiler for various purposes (including optimization decisions). Even though the meta data file format is simple, creating such a file manually can be an additional burden to users. Besides that, some file formats already contain some of the information anyway, such that a separate meta data file is redundant.

**Task:** Make sure that DAPHNE can read a file even if the meta data file is missing. There can be different strategies to handle the missing information, e.g., (1) determine it on the fly, or (2) leaving it *unknown*. Crucial information like the value type or the number of rows/columns should determined, for other properties, it could be optional. Furthermore, after an on-the-fly analysis, the obtained meta data could optionally be stored as a meta data file, such that it does not need to be determined in the next invocation of the program.

**Hints:**
- *Will be provided on request.*
- Related to #63.

Issue #63, opened 2021-06-11
Detection of CSV schema based on data


daphne-eu:
In GitLab by @pdamme on Jun 11, 2021, 19:14

The DAPHNE system is able to read CSV files into frames, whereby each column may have its individual value type. However, at the moment we require the user to provide a metadata file, which specifies the value type of each column. Unfortunately, this complicates the ad-hoc use of CSV files. Thus, akin to libraries like pandas, we want to be able to automatically detect the value type of each column based on the data.

The task is to implement (C++) an I/O utility for that purpose. The value types to consider are signed/unsigned integers of different widths (8, 32, 64 bit), floating-point types of different widths (single, double), boolean, and string.

Thenz:
Assigned to @Thenz

pdamme:
Thanks for deciding to work on this issue, @Thenz! Don't hesitate to get in touch if you have any questions or experience any difficulties.

pdamme:
Thanks for giving it a try last summer semester @Thenz. As this issue was not completed yet, we will offer it to AMLS students again this semester.

EricBenschneider:
Please assign this issue to @EricBenschneider as I am currently working on it!

pdamme:
Sure! Please go ahead.

PR #942, opened 2025-02-04
Generation of meta data files based on CSV data

EricBenschneider:
This PR adds CSV schema detection based on data as described in issue #63.
This PR also enables reading a CSV file without manually adding a meta data file as discussed in issue #688.
Once a CSV file is read and the MetaDataParser cannot find an appropriate meta data file, it will now be automatically generated and stored in the location it was expected.
The changes are tested by the added `generateMetaDataFileTest.cpp`
A new test case in `MetaDataParserTest.cpp` is added to test saving of the generated meta data file.


auge:
Thanks a lot for this PR, I'm looking into the usage part of this (no real code review from my side).

Currently, I see a lot of debug statements generated, I guess they should rather be silenced? (I have 1e9 rows in my csv file): `inferred valueType: ...` (aborting execution as it will take long time …)

when trying on a smaller file (10 rows), I get a segfault on my system.

As I have not looked into the code, I would like to inquiry if the failure to write the .meta file will be silent? I expect daphne to continue normally even when the meta file cannot be written - like when the csv file lies at a location where the user should not be allowed to write.

pdamme (changes requested):
Thanks for this contribution, @EricBenschneider! Being able to read a CSV file without a meta data file present is great for users (even though it will typically come with a certain performance penalty).

Thanks @auge for having a first look at this new feature. I can confirm that it crashes when used from a DaphneDSL script. The case of writing the meta data file to a non-writable location was most likely solved by @EricBenschneider's recent commit e4810392d8aebf3eec3eb03cad3a402ec0c21fd1, but would need additional testing.

So overall, this is a good first step towards solving #63, but a few points need to be addressed:

**Required changes:** (must be addressed before the merge)
1. Make the CI checks pass. Several test cases failed; however, on my system all tests pass... Let's have a look again after you've addressed the other points. The clang-format check failed; please format all changed files with the clang-format specification in `.clang-format` (maybe you can use something like "format document" in your IDE).
1. Extend the test cases. Please consider the following:
    1. Employ a quoted/multi-line string somewhere in the test CSV files (DAPHNE's CSV reader supports such strings).
    1. Add at least one [script-level test case](https://daphne-eu.github.io/daphne/development/Testing/#writing-script-level-test-cases) to check if your feature works from DaphneDSL. Currently, it doesn't seem to work, as @auge found out. For instance, the following DaphneDSL script crashes with the following error message when reading the following CSV file (and no meta data file).
        ```
        print(readFrame("my-frame.csv"));
        ```
        ```
        1,1.1,one
        ```
        ```
        inferred valueType: 0, 1.
        inferred valueType: 6, 1.1.
        inferred valueType: 8, one.
        bin/daphne(+0x14a4422)[0x565112219422]
        /lib/x86_64-linux-gnu/libc.so.6(+0x45320)[0x7b512b129320]
        bin/daphne(+0x1e3abed)[0x565112bafbed]
        ...
        bin/daphne(+0x1557c67)[0x5651122ccc67]
        Segmentation fault (core dumped)
        ```
    1. Test if CSV values like "12+34" are correctly identified as strings, not as integers.
    1. Test if CSV values like "1.23abc" are correctly identified as strings.
3. Make sure your changes to the read kernel are reflected in DaphneDSL and DaphneIR. You added an additional parameter specifying if a header/labels row is present. First, that must not come after the `DaphneContext` (`DCTX`) in the kernel's parameter list, otherwise there will be problems when lowering DaphneIR's `ReadOp` to the kernel call. The additonal parameter must be reflected in the registration of kernels (`src/runtime/local/kernels/kernels.json`), DaphneIR's `ReadOp` (`src/ir/daphneir/DaphneOps.td`), and the `readMatrix()`/`readFrame()` DaphneDSL built-in functions (`src/parser/daphnedsl/DaphneDSLBuiltins.cpp`). This point will be necessary for writing a script-level test case.
3. Please remove all custom prints to stdout/stderr from your code. Consider using DAPHNE's logger (with an appropriate log level like info or debug) for important messages (but I think it's okay to just omit the ones you have). We generally want to avoid unnecessary output.
4. Some of the newly added test cases assume that a meta data file doesn't exist in the beginning, but don't check if this is really the case. They do (try to) remove the (supposedly) created meta data file afterwards, but in case something goes wrong there, the next test run could still see the meta data file created by the previous test run.

**Optional changes:** (can be addressed before the merge or later)
1. Please make sure that the corner case of reading a CSV file with zero rows (and no meta data file) is handled gracefully. To this end, please add a respective test case. Most importantly, DAPHNE should not simply crash in such a case. There should either be an error (because types cannot be inferred when there is no data), or the most general value type (i.e., string) should be assumed for all columns. It would be worthwhile to find out how other systems/libraries (e.g., pandas) handle this case.
1. Make sure that malformed CSV files are handled gracefully. Currently, you say `numCols = std::max(numCols, colIndex);` in `generateFileMetaData()`. To my mind, a different number of columns per rows should raise an error.
2. Currently, the entire CSV file is read once for detecting the schema and once again for the actual read. That will likely result in a 2x slow-down for reading CSV files without a meta data file, which is quite expensive. It would be great to think about more efficient ways, such as scanning the input file just once, interleaving read and schema detection or detecting the schema based on a sample (with the option of correcting it in case not all data fits).
3. Furthermore, the code for detecting the schema implements a little CSV reader on its own (but doesn't seem to support all features of DAPHNE's existing CSV reader, e.g., multi-line strings). It would be great to avoid redundant code.

**Let me know if you have any comments/questions or need further advice to address these points.**

EricBenschneider:
Thanks you guys for the feedback. Most of it could be incorporated. 

Please note that I decided to implement the meta data generation in the scope of MetaDataParser::readMetaData().
Having it directly in the read kernel would be more fitting for on the fly type inferring, which would severely reduce the current overhead introduced by parsing the file twice. On the other hand, we currently need meta data for shape, frame label and type inference which are all called before the read kernel. Therefore an implementation only in the read kernel would not be sufficient.

More details about the emerging overhead can be extracted from the evaluation results:
create_read_breakdown

Issue #858, opened 2024-10-14
Compare different string representations


pdamme:
*Description coming soon!*

PR #946, opened 2025-02-17
#858 Compare different string representations

ts289:
Extend FixedStr16 to 32,64,128 and 256
Implement Simple dictionary-encoded strings (non-order-preserving and order-preserving and 2 implementations of Umbra Strings

Experiments are provided to test the different String Types in their runtime on different operations (comparisons, concat, etc...)


pdamme:
Thanks for this contribution. I'm closing this PR now, because it contains the same code as PR #948.

PR #954, opened 2025-04-02
fix use after free by removing instrumentation from `destroyDaphneContext`

cmrschwarz:
Kernel instrumentation currently causes a use after free bug in  `destroyDaphneContext`.

After instrumentation, `_destroyDaphneContext` looks like this:

```c++
preKernelInstrumentation(kId, ctx);
destroyDaphneContext(ctx);
postKernelInstrumentation(kId, ctx); // <- use after free in here
```

This PR fixes this by disabling instrumentation for destroyDaphneContext.

As [`RewriteToCallKernelOpPass`](https://github.com/daphne-eu/daphne/blob/aa1fb14badc0083c2d927210b782745544260f38/src/compiler/lowering/RewriteToCallKernelOpPass.cpp#L489) assumes that the `kId` and `ctx` parameters are present for all kernels besides `createDaphneContext`, this parameter is retained in `destroyDaphneContext` to keep it simple.

### DIFF
--- a/src/compiler/lowering/RewriteToCallKernelOpPass.cpp
+++ b/src/compiler/lowering/RewriteToCallKernelOpPass.cpp
@@ -484,16 +484,18 @@ class KernelReplacement : public RewritePattern {
         auto kId = rewriter.create<mlir::arith::ConstantOp>(
             loc, rewriter.getI32IntegerAttr(KernelDispatchMapping::instance().registerKernel(kernelFuncName, op)));
 
-        // Inject the current DaphneContext as the last input parameter to
-        // all kernel calls, unless it's a CreateDaphneContextOp.
-        if (!llvm::isa<daphne::CreateDaphneContextOp>(op)) {
+        // Inject the KernelDispatchMapping id as the second to last input parameter to
+        // all kernel calls, unless it's a non-instrumented operation (CreateDaphneContextOp, DestroyDaphneContextOp).
+        if (!llvm::isa<daphne::CreateDaphneContextOp, daphne::DestroyDaphneContextOp>(op))
             // NOTE: kId has to be added before CreateDaphneContextOp because
             // there is an assumption that the CTX is the last argument
             // (LowerToLLVMPass.cpp::623,702). This means the kId is expected to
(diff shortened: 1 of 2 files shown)


pdamme (approved):
**Thanks @cmrschwarz for this contribution.** Yes, there is a bug. I’m not aware that it ever caused a problem (most likely because `destroyDaphneContext` is one of the last things that happen at DaphneDSL run-time) and I can hardly think of a test case, but it should definitely be fixed. **The fix you propose solves the problem.** The `kId` parameter is still passed to the `destroyDaphneContext`-kernel, but the pre/post kernel instrumentation are not added. To this end, you introduced a special case to `genKernelInst.py`. From a functional point of view, that’s alright.

However, for slightly better clarity of this central piece of code, I decided to apply a little change: Let’s say there are two kinds of kernels: non-instrumentable ones (`createDaphneContext`, `destroyDaphneContext`, and `map`) and instrumentable ones (all others). A kernel should get the `kId` parameter and do the pre/post kernel instrumentation if and only if it is instrumentable. With this simple rule, we can avoid a special case for `destroyDaphneContext` in `genKernelInst.py` (which is good because this central script is already quite complex). It suffices to take `destroyDaphneContext` into account in `isInstrumentedOp()`, as you do. As a consequence, the generated `_destroyDaphneContext`-kernel does not have the `kId` parameter anymore. To make it work end-to-end, we need to omit this parameter when creating the `CallKernelOp` in `RewriteToCallKernelOpPass.cpp` for all non-instrumentable operations (`MapOp` is not handled by `RewriteToCallKernelOpPass`, so we can omit it here). Thus, I slightly changed this file, too. Finally, the `DaphneContext` is still added as the last parameter for all kernels but `CreateDaphneContextOp`.

*Edit: I've just noticed that you mention `RewriteToCallKernelOpPass` in your PR description. It's a valid argument, but still, I think it's conceptually a little bit simpler now, even though the PR changes a few more lines. In any case, your contribution is valuable!*

-----

**Thanks again and congrats on your first contribution to DAPHNE!** It would be great to see more contributions from you in the future :) .

Issue #668, opened 2024-02-27
Push-down of arithmetics before source operations


pdamme:
**Motivation:** DaphneDSL offers a couple of *source* operations, which take only scalars as arguments, but produce a matrix as their result. Examples include: `fill()`, `rand()`, and `seq()`. Subsequent arithmetic operations on this result could be pushed before the source operation, in many cases. That way, the program execution could be improved by avoiding unnecessary large intermediate results, thereby reducing the number processor instructions and the memory footprint.

For instance, `fill(a, 10^6, 1) + b` creates a *(1Mx1)* column matrix where all values are the scalar `a` and performs an addition with the scalar `b`, thereby creating another matrix of the same size. The memory footprint is 2 x 1M x 8 bytes (=16 MB)(assuming 64-bit floating-point numbers, and no update-in-place optimizations are done). Alternatively, one could calculate `fill(a + b, 10^6, 1)`, where the addition happens on scalars and only one matrix is created, i.e., the footprint is halved (8 MB).

*Some further examples:*
(Lower case letters indicate scalars, upper case letters indicate matrices.)
- `fill(a, h, w) + b` -> `fill(a + b, h, w)`
- `fill(a, h, w) * b` -> `fill(a * b, h, w)`
- `rand(h, w, a, b, 1, -1) + c` -> `rand(h, w, a + c, b + c, 1, -1)`
- `seq(s, e, i) + x` -> `seq(s + x, e + x, i)`
- `seq(s, e, i) * x` -> `seq(s * x, e * x, i * x)`
- ...

**Task:**
- Get familiar with DaphneDSL operations (operators and built-in functions) in the user documentation.
- The above cases are really just examples. Think of which arithmetic operations (elementwise unary/binary, some are operator symbols in DaphneDSL, some are built-in functions) can be pushed before which source operations with which adaptations to the source operation's arguments). This can be done with pen and paper first.
- Implement these simplification rewrites as [canonicalizations](https://mlir.llvm.org/docs/Canonicalization/) on DaphneIR (implementation in C++ using the MLIR framework). At the time of this writing, all existing canonicalizations reside in `src/ir/daphneir/DaphneDialect.cpp`.
- Add script-level test cases and potentially test cases checking the IR (up to discussion).

**Possible task extensions for teams:**
- Look broadly beyond just arithmetic operations, what other operations could be pushed down? Remember, the goal is to avoid matrix intermediates, and do operations on individual scalars instead. A few ideas:
[...]

PR #966, opened 2025-05-21
# 668: Push-down of arithmetics before source operations

b-m-f:


b-m-f:
WIP on #668

src/ir/daphneir/Canonicalize.cpp
```diff
@@ -249,6 +249,25 @@ mlir::LogicalResult mlir::daphne::SparsityOp::canonicalize(mlir::daphne::Sparsit
 mlir::LogicalResult mlir::daphne::EwAddOp::canonicalize(mlir::daphne::EwAddOp op, PatternRewriter &rewriter) {
     mlir::Value lhs = op.getLhs();
     mlir::Value rhs = op.getRhs();
+    // This will check for the fill operation on the left hand side to push down the arithmetic inside
+    // of it
+    auto lhsFill = lhs.getDefiningOp<mlir::daphne::FillOp>();
+    if (lhsFill) {
+        auto lhsOp = lhs.getDefiningOp();
+        auto fillValue = lhsOp->getOperand(0);
```
pdamme, line 257: Accessing the arguments of the `FillOp` by their name can make the code more readable, e.g., `auto fillValue = lhsFill.getArg();`. You can find the names of the arguments in `src/ir/daphneir/DaphneOps.td`.

src/ir/daphneir/Canonicalize.cpp
```diff
@@ -249,6 +249,25 @@ mlir::LogicalResult mlir::daphne::SparsityOp::canonicalize(mlir::daphne::Sparsit
...
+        auto width = lhsOp->getOperand(1);
+        auto height = lhsOp->getOperand(2);
+        const bool rhsIsSca = !llvm::isa<mlir::daphne::MatrixType, mlir::daphne::FrameType>(rhs.getType());
```
pdamme, line 260: I recommend relying on `CompilerUtils::isScaType()` to check if an `mlir::Value` is a scalar. Beyond matrices and frames, there are more types that are not scalars (e.g., lists, columns (since recently), and a few more technical ones).

src/ir/daphneir/Canonicalize.cpp
```diff
@@ -249,6 +249,25 @@ mlir::LogicalResult mlir::daphne::SparsityOp::canonicalize(mlir::daphne::Sparsit
...
+        if (rhsIsSca) {
+            mlir::daphne::EwAddOp newAdd = rewriter.create<mlir::daphne::EwAddOp>(op.getLoc(), fillValue, rhs);
+            mlir::daphne::FillOp newFill =
+                rewriter.create<mlir::daphne::FillOp>(op.getLoc(), op.getResult().getType(), newAdd, width, height);
+            rewriter.replaceOp(op, {newFill});
+            // releasing the old fill operation
+            rewriter.eraseOp(lhsOp);
```
pdamme, line 267: Note that `lhsOp` could have other uses in the IR. Instead of erasing it, you could just leave it as it is and, if unused, it will be removed by subsequent CSE passes (common sub-expression elimination, which also includes DCE (dead code elimination)). For that to work, the operation must have suitable MLIR traits (e.g., `Pure`).

pdamme:
Just some initial comments that might be helpful for the start, but no pressure :) .

src/ir/daphneir/Canonicalize.cpp
b-m-f: I am not able to find `getArg` in https://mlir.llvm.org/doxygen/classmlir_1_1Operation.html .

```
/daphne/src/ir/daphneir/Canonicalize.cpp:259:29: error: 'class mlir::Operation' has no member named 'getArg'
  259 |         auto width = lhsOp->getArg("numCols");
```

src/ir/daphneir/Canonicalize.cpp
b-m-f: Thanks.

Further down in the existing code there are 2 more checks

```
  292 ▎   ▎   const bool lhsIsSca = !llvm::isa<mlir::daphne::MatrixType, mlir::daphne::FrameType>(lhs.getType());
  293 ▎   ▎   const bool rhsIsSca = !llvm::isa<mlir::daphne::MatrixType, mlir::daphne::FrameType>(rhs.getType());

```

Should I rewrite those as well while I am in this function?

b-m-f:
I need to add `codegen/CodegenTest.cpp` to `test/CMakeLists.txt` in order to activate the `runLIT` test cases. 
Should I keep this activated only for testing and not include this change into my commits?

philipportner:
> @philipportner I need to add `codegen/CodegenTest.cpp` to `test/CMakeLists.txt` in order to activate the `runLIT` test cases. Should I keep this activated only for testing and not include this change into my commits?

Please add them to `test/ir/daphneir/` and run them independently to the `lit` tests in `test/codegen/`.

src/ir/daphneir/Canonicalize.cpp
philipportner: > Thanks, let me double check my LSP setup.
> 
> Unfortunately I still get the same error during compilation though:
> 

It's not a pointer.

I only made this change in your branch and this works on my side.

```diff
-        auto fillValue = lhsFill->getOperand(0);
+        auto fillValue = lhsFill.getArg();
         auto height = lhsFill->getOperand(1);

```
I'd also declare it like so, `auto lhsFill = ..` as `clang-tidy` complains (rule `modernize-use-auto`).

src/ir/daphneir/Canonicalize.cpp
```diff
@@ -249,6 +252,22 @@ mlir::LogicalResult mlir::daphne::SparsityOp::canonicalize(mlir::daphne::Sparsit
...
+    auto lhsOp = lhs.getDefiningOp();
+    if (lhsOp) {
+        bool lhsOpHasScalarFirstArgument = lhsOp->hasTrait<OpTrait::DynamicFirstArgument>();
+        if (lhsOpHasScalarFirstArgument) {
+            auto value = lhsOp->getOperand(0);
+            const bool rhsIsScalar = CompilerUtils::isScaType(rhs.getType());
+            if (rhsIsScalar) {
+                auto newAdd = rewriter.create<mlir::daphne::EwAddOp>(lhsOp->getLoc(), value, rhs);
+                lhsOp->setOperand(0, newAdd);
+                rewriter.replaceOp(op, lhsOp);
```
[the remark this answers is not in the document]
b-m-f: Got a response on the LLVM forums: https://discourse.llvm.org/t/why-does-replaceop-need-to-match-the-operation-variant/86458/4

Since my LSP server can not leverage an LSP container namespace (clangd seems to be missing?) and I want to try the approach with a newer LLVM version I will try to get the project build on my Nix machine before continuing :)

But if I am already on the wrong path pls let me know 🙇‍♂️

PR #967, opened 2025-06-05
668: Push-down of arithmetics before source operations

b-m-f:
Changed the branch to integrate upstream changes into my fork easier.

Still working on https://github.com/daphne-eu/daphne/issues/668

Previous discussions: https://github.com/daphne-eu/daphne/pull/966

### DIFF
--- a/src/ir/daphneir/Canonicalize.cpp
+++ b/src/ir/daphneir/Canonicalize.cpp
@@ -363,6 +363,7 @@ template <class Operation> bool tryPushDownUnary(Operation op, mlir::PatternRewr
  * and skips the intermediate.
  */
 template <class Operation> bool tryPushDownBinary(Operation op, mlir::PatternRewriter &rewriter) {
+    mlir::Type u = mlir::daphne::UnknownType::get(rewriter.getContext());
     mlir::Value lhs = op.getLhs();
     mlir::Value rhs = op.getRhs();
     mlir::daphne::FillOp lhsFill = lhs.getDefiningOp<mlir::daphne::FillOp>();
@@ -379,8 +380,7 @@ template <class Operation> bool tryPushDownBinary(Operation op, mlir::PatternRew
         auto fillValue = lhsFill.getArg();
         auto height = lhsFill.getNumRows();
         auto width = lhsFill.getNumCols();
-        auto newOp = rewriter.create<Operation>(op.getLoc(), CompilerUtils::getValueType(op.getResult().getType()),
-                                                fillValue, rhs);
+        auto newOp = CompilerUtils::retValWithInferredType(rewriter.create<Operation>(op.getLoc(), u, fillValue, rhs));
         rewriter.replaceOpWithNewOp<mlir::daphne::FillOp>(op, op.getResult().getType(), newOp, height, width);
         return true;
     }
@@ -401,8 +401,8 @@ template <class Operation> bool tryPushDownBinary(Operation op, mlir::PatternRew
             // ranges not available in between the original values
             return false;
         }
-        auto newMax = rewriter.create<Operation>(op.getLoc(), max, rhs);
-        auto newMin = rewriter.create<Operation>(op.getLoc(), min, rhs);
+        auto newMax = CompilerUtils::retValWithInferredType(rewriter.create<Operation>(op.getLoc(), u, max, rhs));
+        auto newMin = CompilerUtils::retValWithInferredType(rewriter.create<Operation>(op.getLoc(), u, min, rhs));
         rewriter.replaceOpWithNewOp<mlir::daphne::RandMatrixOp>(op, op.getResult().getType(), height, width, newMin,
                                                                 newMax, sparsity, seed);
         return true;
@@ -415,8 +415,8 @@ template <class Operation> bool tryPushDownBinary(Operation op, mlir::PatternRew
         auto to = lhsSeq.getTo();
         auto inc = lhsSeq.getInc();
 
-        auto newFrom = rewriter.create<Operation>(op.getLoc(), from, rhs);
-        auto newTo = rewriter.create<Operation>(op.getLoc(), to, rhs);
+        auto newFrom = CompilerUtils::retValWithInferredType(rewriter.create<Operation>(op.getLoc(), u, from, rhs));
+        auto newTo = CompilerUtils::retValWithInferredType(rewriter.create<Operation>(op.getLoc(), u, to, rhs));
 
         if (supportsPushDownIncrementUpdate) {
             auto incInt = CompilerUtils::isConstant<int>(inc);
@@ -430,7 +430,7 @@ template <class Operation> bool tryPushDownBinary(Operation op, mlir::PatternRew
                 // change
                 return false;
             }
-            auto newInc = rewriter.create<Operation>(op.getLoc(), rhs, inc);
+            auto newInc = CompilerUtils::retValWithInferredType(rewriter.create<Operation>(op.getLoc(), u, rhs, inc));
 
             rewriter.replaceOpWithNewOp<mlir::daphne::SeqOp>(op, op.getResult().getType(), newFrom, newTo, newInc);
             return true;


b-m-f:
I also encountered errors in the test suite when activating Pushdown for Fill and EwSub:

https://github.com/daphne-eu/daphne/blob/517877294b5c77930b9a245bc8accbaf504ecd01/scripts/algorithms/decisionTree_.daph#L224C1-L226C76
```
  if( aggMax(R[1,]) > 0 ) {
    # TODO .0 in fill() shouldn't be necessary, omiting it requires special ctable instantiation.
    p = ctable(seq(beg+1, beg+v, 1) - 1, fill(1.0, v, 1) - 1, ncol(X2), 1);
```

I also have a hard time understanding what this code is supposed to do. Why use `fill(1.0, v, 1) - 1` over `fill(0, v, 1)`?

b-m-f:
Also https://github.com/daphne-eu/daphne/blob/main/test/api/cli/distributed/DistributedTest.cpp is failing if I mark `RandMatrixOp` as Pure.

Is there another safe way to remove the OP without setting `Pure`? Otherwise it seems that I am currently ending up with 2 calls to `randMatrix`.

pdamme:
Yes, the result of `EwLogOp` *should* always be of a floating-point value type. More precisely, the fact that `EwLogOp` has the `ValueTypeFromArgsFP` trait means that type inference will always assign a floating-point value type to the result of this op. However, it is possible (but actually not intended) to create an `EwLogOp` with an integer value type, because the definition of `EwLogOp` in `src/ir/daphneir/DaphneOps.td` only requires a numerical value type (could also be an integer). Type inference only runs if the result type is unknown or has an unknown value type. Thus, when an `EwLogOp` is created with an integer result type, type inference will *not* change the result type anymore.

Your `EwLogOp` `newLog` has the result value type `si64`, because you don't specify a result type when you create the op (the result type would follow after the location). All elementwise binary ops have a custom builder (defined in `EwBinaryOp` in `DaphneOps.td`) that infers the result type from the argument types, if no result type is specified. However, this inference is wrong (as I realize now, and there is also a todo attached to it): it always uses the type of the lhs argument. I will fix that soon by removing this custom builder.

To solve the problem, you should either (a) specify an unknown type (`mlir::daphne::UnknownType::get(...)`) for the result of `newLog` (then the result type will be inferred later), or (b) specify the value type of the result type of `op` (`CompilerUtils::getValueType(op.getResult().getType())`) for the result type of `newLog` (then the result type is correct from the start).

pdamme:
**Regarding the failing `DistributedTest.cpp` when `RandMatrixOp` has the `Pure` trait:** Attaching `Pure` to `RandMatrixOp` is the right way to go. The reason why the distributed test fails is most likely because `test/api/cli/distributed/distributed_4.daphne` has two identical calls to `rand()`, which I assume get simplified to a single one by common-subexpression elimitation (CSE) when `RandMatrixOp` is `Pure`. However, that should not crash the distributed runtime, so it's most likely a bug in the distributed runtime. For the time being, please simply change this test file, e.g., to the following:

```
m1 = rand(10, 10, 0.0, 20.0, 1.0, 1);
m2 = rand(10, 10, 0.0, 30.0, 1.0, 1);
print(m1 @ m2);
```

If we haven't solved the problem in the distributed runtime by the time we merge your PR, we will change this test file and create an issue for the problem.

philipportner:
Just took a brief look at the code, and I'm wondering if implementing this push down optimization as canonicalization on every operator that we want to push down, for every operator that can be pushed into, is the right place to do that. It seems quite hard to maintain.

@b-m-f  @pdamme  What do you think about an interface based approach?

```
def PushDownOpInterface : OpInterface<"PushDownOpInterface"> {
      let description = [{
          Interface for operations that can be pushed into source/other operations.
      }];

      let methods = [
          InterfaceMethod<
              "Check if operands are compatible for pushdown",
              "bool", "isScalarCompatible",
              (ins "Value":$operand),
              /*methodBody=*/[{}],
              /*defaultImplementation=*/[{
                  // Check if operand is a scalar, same as is currently done in each canonicalize method
              }]
          >,

          InterfaceMethod<
              "Create scalar value",
              "Value", "createScalarValue",
              (ins "OpBuilder &":$builder,
                   "Location":$loc,
                   "Value":$lhs,
                   "Value":$rhs),
              /*methodBody=*/[{}],
              /*defaultImplementation=*/[{
                  // Must be overridden by each op
                  llvm_unreachable("must implement createScalarValue");
              }]
          >
      ];
  }

```

With that interface, every op that wants to can implement it:
```
  def AddOp ...
      [PushDownOpInterface]> {
      let extraClassDeclaration = [{
          Value createScalarValue(OpBuilder &builder, Location loc,
                                    Value lhs, Value rhs) {
              return builder.create<daphne::EwAddOp>(loc, lhs, rhs);
          }

      }];
  }

  def MulOp : ... [PushDownOpInterface]> {
      let extraClassDeclaration = [{
          Value createScalarValue(OpBuilder &builder, Location loc,
                                    Value lhs, Value rhs) {
              return builder.create<daphne::EwMulOp>(loc, lhs, rhs);
          }

      }];
  }
```
That can either be done in the TableGen definition directly or similar to our current interfaces in DAPHNE.

And with that we have a rewrite pass on, for example, the `FillOp`:

```c++
struct FillOpPushdownPattern : public OpRewritePattern<FillOp> {
    LogicalResult matchAndRewrite(FillOp fillOp,
            PatternRewriter &rewriter) const override {
        auto pushdownOp = dyn_cast< PushDownOpInterface >(fillOpArgument);
        if (!pushdownOp)
            return failure();

        // Call the interface functions to:
        // 1. Check if it's actually a scalar like is currently happening in the canonicalize method
        // 2. Call the `createScalarValue` method that creates the correct, pushed down, computation

        auto newFill = rewriter.create<FillOp>(result of pushed down computation...);
        rewriter.replaceOp(fillOpArgument, newFill);
    }
}
```

Additionally, instead of having a `FillOpPushdownPattern` for each op that can be pushed into, another interface could be used to generalize across all those operations. At the end, we would have interface methods for source and push-down ops and a single rewrite pattern doing all the "work" in applying those.

New operators that can be pushed down now only need to implement a single interface and source operators that want to enable this do it with the other interface.

I feel like this approach could eliminate all this duplicated code that's currently in the draft, and makes it quite easy to extend push down optimizations to other operations by implementing the corresponding interface. What do you think?

b-m-f:
@philipportner ,

I wanted to implement all pushdown operations mentioned in the issue and then refactor from there - to have tests in place etc. ( duplication was easier for me l, as I am still getting used to the framework ).
I played around with Traits so far, but thanks for the interface based solution.
Next week I should be able to tackle this 👍

b-m-f:
@philipportner I was not able to get it running using interfaces.

Instead I discussed with @pdamme to use C++ template functions as well as Traits, which is now completed and pushed here for review.
The individual Operations still have to enable Canonicalization with `hasCanonicalizeMethod = 1`.
This could be avoided with a dedicated compiler pass -> out of scope for this ticket.

pdamme:
> All elementwise binary ops have a custom builder (defined in `EwBinaryOp` in `DaphneOps.td`) that infers the result type from the argument types, if no result type is specified. However, this inference is wrong (as I realize now, and there is also a todo attached to it): it always uses the type of the lhs argument. I will fix that soon by removing this custom builder.

**Just as a quick follow-up:** I fixed the problem with the incorrect type inference in the custom builder of the elementwise binary ops today in b55d1a88358196f8070622cf6a59a26e515c0295.

b-m-f:
Hi @pdamme,

I have added all suggestions into the MR.

There is one inconsistency with `rand + EwMinus` for the `PushDownLinear + Unary` branch though.

```
Filters: [operations]

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
run_tests is a Catch v2.13.8 host application.
Run with -? for options

-------------------------------------------------------------------------------
randEwMinus
  randEwMinus_1.daphne
-------------------------------------------------------------------------------
/home/max/daphne/test/api/cli/operations/CanonicalizationPushDownTest.cpp:57
...............................................................................

/home/max/daphne/test/api/cli/Utils.h:350: FAILED:
  CHECK( out.str() == exp )
with expansion:
  "DenseMatrix(1x3, int64_t)
  -6 -6 -3
  "
  ==
  "DenseMatrix(1x3, int64_t)
  -3 -3 -6
  "

-------------------------------------------------------------------------------
randEwMinus
  randEwMinus_2.daphne
-------------------------------------------------------------------------------
/home/max/daphne/test/api/cli/operations/CanonicalizationPushDownTest.cpp:57
...............................................................................

/home/max/daphne/test/api/cli/Utils.h:350: FAILED:
  CHECK( out.str() == exp )
with expansion:
  "DenseMatrix(1x3, int64_t)
  3 3 6
  "
  ==
  "DenseMatrix(1x3, int64_t)
  6 6 3
  "

-------------------------------------------------------------------------------
randEwMinus
  randEwMinus_3.daphne
-------------------------------------------------------------------------------
/home/max/daphne/test/api/cli/operations/CanonicalizationPushDownTest.cpp:57
...............................................................................

/home/max/daphne/test/api/cli/Utils.h:350: FAILED:
  CHECK( out.str() == exp )
with expansion:
  "DenseMatrix(1x3, int64_t)
  -6 -4 1
  "
  ==
  "DenseMatrix(1x3, int64_t)
  3 1 -4
  "
```

I have left the failing tests included for now instead of accepting the new results to discuss whether they should be accepted or if I should just remove the `PushDownLinear` branch from the `tryPushDownUnary` function.

While the results are valid even though they differ from **main**. This is because the `max` and `min` are swapped during **push-down**, which in turn gives new results - even with the same seed.

For me this still seems fine, as the results are in the correct range, but for absolute consistency with the **non pushed down** version this might not be desired.

Just let me know and I will either:
- check in the new test results to get tests passing 
Or
- remove the `PushDownLinear` + `Unary` branch from the code

b-m-f:
My assumption from the previous comment seems to have been correct. I updated the calls to `create` according to what is written in that commit.

The [questions](https://github.com/daphne-eu/daphne/pull/967#issuecomment-3121608370) for the tests still remain before merge.

Please let me know whether this "change" in behavior is acceptable.

Issue #235, opened 2022-03-22
Right indexing with negative numbers


daphne-eu:
In GitLab by @pdamme on Mar 22, 2022, 18:26

By right indexing, we can access parts of data objects (matrices/frames) by position. For instance, `y = X[, 5]` extracts the 5th column from `X` and stores it in `y`. At the moment, we silently assume that positions are *non-negative*. However, for convenience, it would be great to support right indexing with *negative* numbers, too. The semantics should be the same as in many other languages: start counting from the end. For instance, `y = X[, -1]` should extract the *last* column in `X`, which is equivalent to `y = X[, ncol(X) - 1]`.

PR #970, opened 2025-06-29
[DAPHNE-#235] Support negative indexing in left and right indexing operations on matrices and frames

BlaiPuchol:
This PR adds **support for negative indexing in both left and right indexing operations** on matrices and frames #235. The changes include:

Right indexing enhancements:
- Support for negative row/column positions in Frame, Matrix and DenseMatrix right indexing with scalars and ranges.
- New test cases (right_indexing_success_7/8) covering various combinations of negative indices.
- Removal of outdated error handling for out-of-bounds negative indices.

Left indexing enhancements:
- Support for negative indices in Matrix and DenseMatrix left indexing with scalars and ranges.
- New test cases (left_indexing_success_4/5/6) validating negative indexing for various input types and configurations.
- Removal of outdated error handling for out-of-bounds negative indices.


BlaiPuchol:
Thanks for your comments @pdamme . Here I'm going to either explain why the code is like that or state if the point has been solved:

- [x] **Ensure the CI checks pass.** I think now it will pass. 

- [x] **Script-level test cases checking out-of-bounds situations for left and right indexing.** I have added 4 new failure tests: `right_indexing_failure_{11,12}.daphne` and `right_indexing_failure_{8,9}.daphne`.
- [x] **Conversion of negative indices.**. The suggestion was to simplify the negative index conversion. However, this change breaks a fundamental feature: single-element indexing with negative numbers, such as `X[-1, -1]`.

  If I understand it correctly, in DAPHNE, an indexing operation like `X[-1, -1]` is lowered by the compiler into `Slice` operations. Accessing a single index `i` is represented as a slice from `i` up to (but not including) `i+1`.

  *   The row index `[-1]` becomes a `SliceRowOp` with `lowerIncl = -1` and `upperExcl = -1 + 1 = 0`.
  *   The column index `[-1]` becomes a `SliceColOp` with `lowerIncl = -1` and `upperExcl = -1 + 1 = 0`.

  The core of the problem lies in correctly converting these `[-1:0]` slices into positive indices.

  **Why the simplified code fails**

  Let's trace the row slice `lowerIncl=-1`, `upperExcl=0` on a matrix with 10 rows (`srcNumRows = 10`) using the proposed simplified logic:

  ```cpp
  if (loInPos < 0)
      loInPos += srcNumRows;
  if (upExPos < 0)
      upExPos += srcNumRows;
  ```

  1. `loInPos` starts as `-1`. The first `if` converts it to `loInPos = -1 + 10 = 9`.
  2. `upExPos` starts as `0`. The second `if` condition (`upExPos < 0`) is false, so `upExPos` remains `0`.
  3. The resulting slice is `[9:0]`.
  4. The subsequent validation check `if (loInPos > upExPos)` correctly identifies this as an invalid range (`9 > 0`) and throws an error. This is why the tests fail with this logic.

  **Why the original logic is correct**

  The original logic was specifically designed to handle this `[-1:0]` pattern.
  ```cpp
  if (loInPos < 0) {
      loInPos += srcNumRows;
      if (upExPos <= 0)
          upExPos += srcNumRows;
  } else if (upExPos < 0)
      upExPos += srcNumRows;
  ```
  1.  `loInPos` starts as `-1`. The outer `if` is true, and `loInPos` is converted to `9`.
  2.  The crucial nested `if (upExPos <= 0)` is also true, because `upExPos` is `0`.
  3.  `upExPos` is then converted to `upExPos = 0 + 10 = 10`.
  4.  The resulting slice is `[9:10]`, which correctly selects the last row (index 9).

  This demonstrates that the more complex logic is essential for supporting Python-style negative indexing for single elements.

- [x] **Please clarify:** `InsertRow.h`, line 112 `if(rowLowerIncl < 0 && rowUpperExcl == 0):` **What is this check for?**

  This code is another implementation of the exact same principle discussed above, but it's specific to the `InsertRow` and `InsertCol` kernels. It explicitly handles the `[-i:0]` slicing pattern used for single-element assignment.

  Let's trace an assignment like `X[-1, ...] = ...`, which becomes an `InsertRow` call with `rowLowerIncl = -1` and `rowUpperExcl = 0` on a 10-row matrix:
  1.  The code first converts `rowLowerIncl` to its positive equivalent: `rowLowerIncl_Pos` becomes `10 + (-1) = 9`.
  2.  The check `if (rowLowerIncl < 0 && rowUpperExcl == 0)` is **true**.
  3.  It then explicitly sets the upper bound: `rowUpperExcl_Pos = rowLowerIncl_Pos + 1`, which is `9 + 1 = 10`.
  4.  This yields the correct insertion range `[9, 10)`.

  This confirms that this pattern is intentional and required for correct functionality. While the implementation in `InsertRow.h` is slightly different from `SliceRow.h`, they both solve the same problem of interpreting single-element negative indexing. The logic in `SliceRow.h` is simply more general.

- [x] **Please clarify:** `SliceCol.h`/`SliceRow.h`: **Why did you change the condition in** `validateArgsSliceRow()`/`validateArgsSliceCol()`?

  The primary benefit of changing the logic is for **clarity**. The proposed new logic is a much more direct and understandable implementation of the boundary rules.

  **New logic:**
  ```cpp
  if (lowerIncl < 0 || static_cast<size_t>(lowerIncl) >= numRowsArg || upperExcl < lowerIncl ||
      numRowsArg < static_cast<size_t>(upperExcl))
  ```

  This directly translates to:
  *   `lowerIncl` cannot be negative.
  *   `lowerIncl` must be less than `numRowsArg`.
  *   `upperExcl` cannot be less than `lowerIncl`.
  *   `upperExcl` cannot be greater than `numRowsArg`.

  While the old logic may have been functionally correct in most cases, it was hard to reason about. The new logic is a direct and clean implementation of the required validation rules. 

- [x] **Unit test cases in** `test/runtime/local/kernels/`: I have updated the "check throws" tests to use out-of-bounds negative indices.
- [x] **Test cases:** `right_indexing_success_{7,8}.{daphne,txt}`: Now we extract elements from the same positions as in the corresponding test cases with positive indices from which I adapted these test files. The same .txt files are used for the reference output (except for the print-outs of the indexing expressions themselves).
- [x] **Test cases:** I have removed superfluous spaces at the line endings in the `.daphne` files.

pdamme (approved):
**Thanks for incorporating the suggested changes and clarifying the open questions, @BlaiPuchol.** With your updates and a few little changes that I've just applied, **this PR is ready to be merged now**.

*In response to your answers:*
1. Very good.
2. Very good (I've just fixed a little bug here, see my commit, but that's fine).
3. Okay, I see, thanks for the clarification; I totally overlooked the `<=` instead of `<`. This does make sense. Unfortunately, it also means users can write `X[-1:0, ]` in DaphneDSL now, which is not intended, i.e., that is currently but should not be equivalent to `X[-1, ]`. However, fixing this problem is not trivial: at DaphneDSL parse-time, we cannot reliably know the values used in indexing; in DaphneIR we cannot distinguish anymore, if the user specified only a single row/column (as in `X[-1, ]`, which is valid) or a lower/upper bound (as in `X[-1:0, ]`, which is invalid). Possible solutions include (1) adding variants of the slice and insert ops that take only a single position (not lower/upper) as input, or (2) changing the semantics of upper exclusive to upper inclusive. In any case, I think that's beyond the scope of this PR, so we will accept this little issue for now.
4. Same as 3.
5. Okay, fair enough and fine by me.
6. Very good.
7. Very good.
8. Very good.

By the way, good catch regarding the bug in `ExtractColOp::inferShape()`!

-----

**Congrats on your first contribution to DAPHNE!**

PR #977, opened 2025-07-15
Optimistic spliting on sum in col agg grp

saminbassiri:
# Optimistic Splitting in Daphne

This PR implements **Optimistic Splitting** in Daphne’s `ColAggGrp` kernel function. The goal is to improve the performance of columnar operations by applying optimistic splitting to aggregation functions.

### Optimistic Splitting Approach

Each aggregation function is divided into three separate steps:

1. Compute the smaller half of the result.

2. Check for overflow.

3. If overflow occurs, update the main result.

This method improves cache efficiency by reducing the size of intermediate results and limiting access to the main result only when necessary.

* * *

### Key Changes

* Implement optimistic splitting in grouped aggregation via `ColAggGrp`.

* Map `AggOpCode` to the corresponding `BinaryOpCode` for optimistic splitting.
  
  * Use `isSupportOptimistic` to check if the data type is supported.
  
  * Extend `AggOpCodeUtils` to correctly retrieve the matching `BinaryOpCode`.

* Add support for optimistic splitting in aggregation using `BinaryOpCode`:
  
  * Introduce `BinaryOpCode` logic for optimistic split.
  
  * Currently supports the `SUM` aggregation.
  
  * Add tests to verify correctness.

* Add a template in `ValueTypeUtils` to retrieve the half-size type.

* * *

Performance Test
================

To evaluate performance improvements, we compared the execution time of the `groupSum` built-in function in Daphne. The size of intermediate results depends on the number of distinct group values, which were chosen based on cache size. Overall, the test showed modest improvements in system performance with optimistic splitting enabled.

## System cache sizes:

```
Name:                      13th Gen Intel(R) Core(TM) i5-13420H
L2CacheSize:               7168
L3CacheSize:               12288
NumberOfCores:             8
NumberOfLogicalProcessors: 12
```

## Test script in Daphne DSL:

```
fr = createFrame(groups, values, "a", "b");

res = groupSum(fr, "a", "b");
```

The table shows the Wall-clock time computed by the `time` command in Linux for calling the test script 3 times for columnar and non columnar aggrigation.

| Test Case   | Target Size | Distinct Groups | No columnar                             | Optimistic split and columnar           |
| ----------- | ----------- | --------------- | --------------------------------------- | --------------------------------------- |
| L2 Small    | 1 MiB       | 262144          | 0m2.587s<br/>0m2.688s<br/>0m3.563s<br/> | 0m2.590s<br/>0m3.545s<br/>0m2.762s<br/> |
| L2 Fit      | 6.5 MiB     | 1703936         | 0m2.577s<br/>0m3.411s<br/>0m3.128s      | 0m2.675s<br/>0m3.282s<br/>0m2.713s<br/> |
| L2 Overflow | 8 MiB       | 2097152         | 0m3.504s<br/>0m3.359s<br/>0m3.193s      | 0m2.617s<br/>0m2.731s<br/>0m2.806s<br/> |
| L3 Fit      | 11 MiB      | 2883584         | 0m3.340s<br/>0m3.138s<br/>0m3.463s      | 0m3.149s<br/>0m2.733s<br/>0m2.977s<br/> |
| L3 Overflow | 13 MiB      | 3407872         | 0m3.982s<br/>0m3.648s<br/>0m3.538s      | 0m3.606s<br/>0m3.454s<br/>0m3.209s<br/> |


pdamme (approved):
**Thanks for this PR, @saminbassiri.** This variant of optimistic splitting could improve the performance of grouped aggregation and is a welcome addition to DAPHNE. Thanks also for the detailed PR message describing your contribution. **Overall, the core of the algorithm in `ColAggGrp.h` looks good to me.** *However, regarding the integration into the code base, a few points need to be addressed before we can merge this PR.*

**Required changes:** (necessary before we can merge this PR)
1. **Keep the changes more local to the `colAggGrp`-kernel.** You tried to factor out some parts of it to `ValueTypeUtils.h/cpp` (deriving the `HalfType`s for the splitting), `AggOpCode.h` (deriving binary op codes for specific parts of optimistic splitting), and `BinaryOpCode.h`/`EwBinarySca.h` (specific primitives for optimistic splitting). I appreciate your attempt to integrate with some of the code base’s central files. However, given that optimistic splitting is an algorithmic trick which is quite specific to grouped aggregation, I think the changes should mostly be kept local to that kernel. With the current approach, I see the following downsides:
   1. The central files mentioned above get unnecessary cluttered and harder to understand, while the reuse potential of the newly added things is unclear.
   2. The code of the optimistic splitting-based grouped aggregation is harder to read as it calls functions like `funcOpt`, `funcOverflow`, and `funcExp` instead of performing the necessary calculations directly.
   3. The performance of the grouped aggregation with optimistic splitting may suffer, e.g., due to calling 2-3 function pointers per value to aggregate. This overhead contradicts the goal of improving performance. More detailed experiments would be necessary to see if this is really a problem, though.

   Overall, I recommend reducing the level of abstraction in the `colAggGrp`-kernel by tailoring the algorithm to the few cases it supports (`SUM` on 64/32-bit integers).
1. **Apply optimistic splitting only when beneficial or make it conditional.** At the moment, optimistic splitting is unconditionally applied for `SUM` aggregation and 64/32-bit integer value types. While optimistic splitting could improve the performance in certain cases (especially when the randomly accessed array fits into the cache with optimistic splitting, but doesn’t fit without), there are cases when optimistic splitting will not be beneficial (e.g., when the data already fits into the cache without splitting or doesn’t fit into the cache even with splitting). It would be great to have a clear idea of the overhead of optimistic splitting when it is not beneficial (see comment on experiments below). If the overhead is non-negligible, either (a) optimistic splitting should be off by default and only be applied by the kernel if an optional command-line argument (e.g., `--optimistic-splitting`) is provided, or (b) the kernel should automatically figure out if optimistic splitting is beneficial (e.g., depending on the number of groups, value type, and cache sizes). (b) would be more useful but also more involved. (a) would be totally fine for this PR and the AMLS project.
1. **Add tests for grouped aggregation with optimistic splitting.** You added a unit test case in `EwBinaryScaTest.cpp`, which focuses on the new binary ops you introduced (but see also point 1), but does not invoke the `colAggGrp`-kernel. While there are existing test case for the correctness of `colAggGrp` in `test/runtime/local/kernels/ColAggGrpTest.cpp`, these use only tiny input data and don’t cover interesting cases w.r.t. optimistic splitting, so they are not sufficient. Before we merge this PR, additional test cases for grouped aggregation focusing on optimistic splitting are required. These could be unit test cases (extending the existing ones in `ColAggGrpTest.cpp` by value types and data that is interesting for optimistic splitting) or script-level test cases (also with value types and data for optimistic splitting; these tests could leverage a flag like `--optimistic-splitting`, see point 2, to check if the result with and without optimistic splitting is the same).

**Optional changes:** (recommended, but not necessary for merging)
1. **Make the experiment more insightful.** It’s great that you conducted a little experiment to check if optimistic splitting actually improves the performance of grouped aggregation. However, at the moment, it’s hard to draw reliable conclusions from the experiment, because:
   1. It is unclear how the data (variables `groups` and `values`) is generated. Could you provide the full DaphneDSL script you used?
   2. While measuring time using the `time` command yields the end-to-end runtime of DAPHNE, a runtime break-down telling how long the grouped aggregation took would be more insightful. I assume your time measurement also includes the data generation and that may dominate the script runtime. You could use the DaphneDSL built-in function `now()` to measure a specific part of the script (e.g., `t0 = now(); res = groupSum(...); t1 = now(); d = t1 - t0;`) or invoke `daphne` with the the `--statistics` flag (which reports the runtimes of the most time consuming kernels).
   3. The comparison to the non-columnar grouped aggregation as the only baseline does not reveal if the performance differences are due to columnar processing or due to optimistic splitting. It would be best to include columnar processing without optimistic splitting as an additional baseline (which would be easy with a flag like `--optimistic-splitting` mentioned above).

**I hope these comments are helpful for improving the PR. Feel free to share your thoughts.**

pdamme (changes requested):
Sorry, with my comment above, I meant to "request changes" (not "approve" this PR).

Issue #513, opened 2023-04-25
Support for more aggregation functions in group-join operation/kernel


pdamme:
tba

PR #1005, opened 2026-06-02
513: support for more aggregation functions in group-join operation/kernel

khangkontum:
Issue: #513 
Work in progress


khangkontum:
Hi @pdamme, before I wrap up the implementation and go to the extending task, do you think using std::variant inside `GroupJoin.h` is a good approach here? since I can't really come up with a way using `template` to handle different type of aggregate state inside a shared join hashmap. Pre-initialized the res frame also not viable since we don't know the size of joined state at that point. 

Another approach I could think of is temporary create the whole frame with potentially null right hand side group rows, so that we can handle each aggregate to each column individually. Then finally trim those rows before returning.

pdamme:
Hi @khangkontum, **thanks for the good work so far**. At a quick glance, **your approach to re-implementing the `groupJoin`-kernel is comprehensible and seems to be functionally correct** (but I didn't test it; the CI checks show that it doesn't compile at the moment).

However, I think its **efficiency can be improved**. For hash-based algorithms, like this hash join plus hash group-by, it is crucial to make the hash table as small as possible such that it ideally fits into the processor cache. In that regard, your current implemention has opportunities for improvement:

- Using `std::variant` to represent an accumulator value of any possible value type is not very storage-efficient. The `sizeof` your `AggValue` will (most likely) be that of the largest supported type. That means, even if you just use the `AggValue` for a `int64_t` (or a narrower integer type), you will end up needing `sizeof(std::string)` bytes (could be sth like 32 bytes), a significant waste of space.
- `JoinState` consumes more space in the hash table than is strictly necessary. For instance, `std::vector<AggState>` in `JoinState` will consume some space just for the vector object (but you don't need `std::vector`'s ability to grow, since all these vectors contain the same number of elements). Furthermore, the actual data in the vector (the `AggState`s) may reside in a different memory location and incur additional random memory accesses that could slow down the kernel.
- `AggState` redundantly stores the meta information on the columns to aggregate. All its member variables except for `value`, `count`, and `initialized` are global information that applies to *all* entries in the hash table, but currently, this information is replicated for *each* entry.

Nevertheless, it is indeed **non-trivial to come up with an implementation that is both efficient and general enough** to handle arbitrary combinations of columns to aggregate. Staying closer to the previous implementation of the `groupJoin`-kernel, you would need to template-specialize it to different combinations of value types of the aggregation columns, which would lead to an exponential explosion of kernel variants to instantiate at C++ compile-time. **I see two approaches** how to circumvent these problems:

1. Replace `JoinState::aggStates` by a byte buffer that you dynamically allocate to be exactly as many bytes as you need to store the accumulators for each aggregation column. Then, interpret and process these bytes depending on the respective value types. Nevertheless, due to the dynamic allocation, the buffers of all the entries in the hash table may not be in adjecent memory locations, thereby incurring additional random memory accesses. A solution could be to implement your own hash table that uses a single large array that you allocate for all entries.
2. Switch to a columnar processing style, i.e., process the aggregation columns one-by-one with dedicated routines template-specialized to the respective value types. Maybe that is what you mean by your idea above. Essentially, you could first do the join and grouping and output a column of group ids for each rhs input tuple. Then, you could do one columnar aggregation step for each aggregation column. You may want to take a look at the lowering from frame operations to columnar operations (`src/compiler/lowering/RewriteToColumnarOpsPass.cpp`, `daphne --columnar`, `ColGroupFirstOp`/`ColGroupNextOp`).

In any case, it's a good idea to keep different variants of the `groupJoin`-kernel and to compare their performance experimentally.

pdamme:
PS: Note that the lowering from join and group-by to group-join (the extended task) can be addressed independently of a more efficient implementation of the `groupJoin`-kernel.

khangkontum:
Hi @pdamme, what I meant in the previous comment was handling each aggregate column individually (in a columnar manner). You can check what I meant in the latest commit, which should complete the task of expanding to other aggregate types over variadic input.

pdamme:
**Thanks for continuing the implementation, @khangkontum!** Here are some comments:

**Regarding the re-write of the `groupJoin`-kernel:** Yes, this is roughly what I meant by a columnar approach. Maybe your implementation could be further optimized, here are some ideas:

- I think in `aggTypedColumn()`, the intermediate vector `acc` could be ommited to save space and to avoid the second loop (the one over `numGroups`). Instead of `acc[rhsGroup[r]] += rhsAggColumn[r];`, you could write `out[groupToOut[rhsGroup[r]]] += rhsAggColumn[r];`.
- You use `std::vector` most of the time even though in most cases you already know the size of the vector upfront (and pass it to the `std::vector` constructor). Using `std::vector` instead of a dynamically allocated raw array *might* add some overhead (but this overhead might as well get optimized away by the C++ compiler...).

It would also be valuable to find out if your re-written `groupJoin`-kernel achieves a performance improvement compared to (a) join followed group as separate operations, and (b) the original `groupJoin`-kernel before your changes (just for the case of `SUM` on a single rhs column). For (a) I recommend considering two alternatives, `daphne` with and without `--columnar`. This flag effects a lowering from `Frame`-based operations to `Column`-based operations (you can check that in the IR by using `daphne` with `--explain columnar`). I would encourage you to perform a little experiment with sufficiently large data; you could report the outcome here, but wouldn't need to add it to the codebase.

**Regarding the tests:**

- Thanks for extending the unit test cases!
- It would be great if you could also add some [script-level tests](https://daphne-project.github.io/daphne/development/Testing/#writing-script-level-test-cases) (e.g., in `test/api/cli/operations/`) that check if the group-join works correctly end-to-end in a DaphneDSL script (e.g., if multiple aggregate columns are parsed correctly etc.). So far, your changes to `DaphneDSLBuiltins.cpp` are not covered by test cases yet.
- You might want to simplify your result checking by using DAPHNE’s `order`-kernel to sort the result before comparing it to a hard-coded sorted reference output. Just an idea.
- Currently one test case fails (`select_asterisk`). Maybe that is due to your changes to type inference.

khangkontum:
Hi @pdamme, thanks for your feedback. I have addressed your two points in the above commits.
I have also run the benchmark. These are the operators that I run through:
  1. `gj_new` - `groupJoin(f1, f2, "id", "rid", "SUM", "v0")` on the PR binary
  2. `gj_orig` - `groupJoin(f1, f2, "id", "rid", "v0")` (old 5-arg form) on the main binary
  3. `join_group` - `innerJoin(f1, f2, "id", "rid", M)` + `groupSum(j, "id", "v0")` on the PR binary
  4. `join_group_col` - same script as 3, run with `--columnar` (that the rewrite to Column ops took effect is evident from the results: identical group counts at ~240× the speed of the plain Frame variant)

Benchmark machine: AMD Ryzen 9 3900X, 125 GiB RAM, Linux x86_64, inside the `daphneeu/daphne-dev` image container.

This is the collected result (median of 5 runs) per cell

| N (groups) : M (rhs rows) | variant | median | ns / rhs row |
|---|---|---:|---:|
| 100k : 10M | **gj_new** | **168 ms** | 16.8 |
| | gj_orig | 810 ms | 81.0 |
| | join_group | 15.36 s | 1536 |
| | join_group_col | 64 ms | 6.4 |
| 1M : 10M | **gj_new** | **750 ms** | 75.0 |
| | gj_orig | 2.13 s | 213 |
| | join_group | 17.95 s | 1795 |
| | join_group_col | 370 ms | 37.0 |
| 1M : 50M | **gj_new** | **3.41 s** | 68.2 |
| | gj_orig | 9.80 s | 196 |
| | join_group | 89.03 s | 1781 |
| | join_group_col | 1.22 s | 24.4 |

The advantages of the `--columnar` flag are clear. That said, with the semester ending, I won't be able to implement the extended task in time, though I'd be glad to pick it back up once I'm through most of my finals. Should I continue that work in this PR, or would it be better to open a new one for the extended task?

Issue #985, opened 2025-10-13
Efficient File I/O Plug-ins for Widely-used File Formats


pdamme:
**Motivation:** The input to integrated data analysis pipelines, that combine query processing, machine learning, and high-performance computing, could be provided in various general-purpose and domain-specific file formats. To embrace a large variety of such formats, DAPHNE is extensible w.r.t. to file readers/writers (pull request coming soon), i.e., expert users can add their own file I/O plug-ins without touching the source code of DAPHNE. However, so far there is only a limited number of plug-ins available.

**Task:** (in C++) Implement additional file I/O plug-ins for a range of widely used file formats for different data modalities, such as tabular data, (sparse) matrices/graphs, audio, images, and time series. These plug-ins may be based on existing open-source libraries (with compatible license). Apply format-specific and format-agnostic tricks to make the file I/O efficient, e.g., by exploiting parallelism or pushing down certain operations into the readers.

**Hints:**
*coming soon*

pdamme:
This issue is related to #510, #517, and #822, but goes beyond them due to the focus on the implementation as plug-ins/extensions. They might be solved in the context of this issue.

PR #1006, opened 2026-06-21
feat(io): ORC reader for DenseMatrix and Frame (initial prototype, refs #985)

ShreyasGS:
## Summary

Initial-prototype reader for Apache ORC files. Part of #985 ("Efficient
File I/O Plug-ins for Widely-used File Formats") on the ORC side.

The reader uses Apache Arrow's ORC adapter
(`arrow::adapters::orc::ORCFileReader`), which keeps the implementation
the same shape as the existing Parquet reader. The public signature
matches the reader-signature contract from PR #993:

    void(void* res, const FileMetaData& fmd, const char* filename,
         const std::map<std::string, std::string>& options, DaphneContext* ctx)

Until PR #993 lands, the reader is dispatched from `Read.h`'s existing
if/else on the `.orc` extension.

## What's in this PR

- `build.sh:899` — adds `-DARROW_ORC=ON` to the Arrow cmake invocation.
  Enables Arrow's ORC adapter (the `arrow/adapters/orc/adapter.h` header
  and the corresponding symbols linked into `libarrow`).
- `src/runtime/local/io/ReadOrc.h` — new header-only reader for
  `DenseMatrix<double>`, `DenseMatrix<int64_t>`, and `Frame`. Opens
  the file via Arrow, reads into an `arrow::Table`, and copies into
  the destination buffer.
- `src/runtime/local/kernels/Read.h` — `.orc` dispatch in both
  `Read<DenseMatrix<VT>>::apply` and `Read<Frame>::apply`.
- `test/runtime/local/io/ReadOrcTest.cpp` + 5 fixture pairs — 12
  Catch2 cases: 3 reading valid ORC files, 2 end-to-end through
  `Read.h`, 7 error paths (missing file, row/col mismatch, string
  columns, null values, type mismatch, unsupported value types).
- `test/runtime/local/io/gen_orc_test_data.py` — fixture generator
  using `pyarrow.orc`. Run once to regenerate fixtures deterministically.

## Test plan

- [x] Build cleanly against Arrow with `-DARROW_ORC=ON`, verified in
  a locally derived image.
- [x] End-to-end verified via DaphneDSL: `readMatrix()` and `readFrame()`
  on each fixture produce the expected values; null-rejection and
  string-rejection paths throw the expected runtime errors.

Refs #985.


pdamme (commented):
**Thanks for this PR, @ShreyasGS! Support for reading ORC files in DAPHNE is highly welcome!**

**Overall, your code looks good to me.** Unfortunately, **I cannot test it at the moment** because the build fails. Even if I locally modify `build.sh` to download and build Arrow with `-DARROW_ORC=ON` inside the container, I get linker errors due to undefined symbols (e.g., `ld.lld: error: undefined symbol: arrow::adapters::orc::ORCFileReader::Open(...)`). It's currently not obvious to me how to fix those. Anyway, here are some comments on your code:

**Major comments:** *These are most relevant and should ideally be addressed in this project.*
1. The ORC reader **copies** the data read by Arrow/ORC into the DAPHNE data objects (`DenseMatrix` and `Frame`). For efficiency, it would be great to avoid this copying if possible. Maybe you could (a) wrap a DAPHNE data object around an existing buffer from Arrow/ORC or (b) make Arrow/ORC store the data it reads from the file directly into a buffer that is provided by the DAPHNE data object. I understand that the chunked representation of Arrow/ORC is a particular challenge in that context.
1. At the moment, **string columns** in ORC files are not supported. As strings are commonplace in real-world data sets, and DAPHNE supports strings as the value type of `DenseMatrix` and `Frame` columns, it would be great to support string columns in the ORC reader as well.
1. Ideally, the ORC reader should be integrated as a file IO **extension**, once #993 has been merged (*I'm working on that…*).
1. Support for **pushing down operations** (such as projections on a subset of columns, filters, aggregations, ...) into the ORC reader would be awesome. Push-downs will also touch the DAPHNE compiler and could be addressed in this PR or a follow-up PR once the basic support for reading ORC files is finished.

**Minor comments:** *These would be nice to have, but they are not that important for this project and could become follow-up issues to be addressed by someone else.*
1. Currently, ORC files that contain `NULL` values are not supported, which makes sense as DAPHNE does not support `NULL` yet. For the time being, it might be handy to allow replacing `NULL` values by some user-defined constant (e.g., `0`, `-1`, `""`). The ORC reader could have an option for that purpose.
1. Currently, a meta data file is required for reading an ORC file. This is consistent with the other file formats supported by DAPHNE. ORC also stores the same information anyway. It's great that you check for consistency between the meta data file and the meta data inside the ORC file; this avoids unexpected outcomes. Nevertheless, users might perceive the need for a meta data file as an unnecessary burden given that the information is present in the ORC file anyway. Maybe the meta data could be extracted from the ORC file itself instead of relying on a meta data file.

Issue #1002, opened 2026-04-26
Automatic fusion of result data analysis into linear/relational algebra operators


pdamme:
*Description coming soon...*

PR #1007, opened 2026-06-21
Result Analysis Fusion

migraine-user:
# Prototype
Prototype implementation of #1002 by compile time constructs in C++. 
Implementation of `EwUnaryMat` with fused computation of:
- mean
- min
- symmetricity
- sparsity
- number of distinct elements

Benchmarks available to run in comparison to:
- original `EwUnaryMat`
- original `EwUnaryMat` and other invocations of kernels corresponding to each result characteristic


migraine-user:
Now, the test cases for the newly implemented, refactored fused kernels should be correct and ready for benchmarking.

Issue #986, opened 2025-10-13
DAPHNE and the Python Data Science Ecosystem: Efficient String Data Transfer


pdamme:
**Motivation:** As Python is the language of choice for most data scientists nowadays, DAPHNE supports efficient bi-directional data transfer with popular Python libraries like numpy, pandas, SciPy, TensorFlow, and PyTorch. So far, only the transfer of numeric data is supported. However, real-world data sets often contain string-valued attributes that need to be transformed to numbers before applying ML algorithms and this transformation could happen either in Python or in DAPHNE.

**Task:** (in C++ and Python) Extend DAPHNE’s data transfer to/from Python (e.g., numpy, pandas) by efficient support for string data. Investigate if performing typical feature transformations like recoding and one-hot-encoding is more efficiently done in Python or in DAPHNE and optimize the transformations in DAPHNE.

**Hints:**
*coming soon*

PR #1008, opened 2026-06-21
Support FIXEDSTR16 in NumPy imports

Mango-Clark:
Refs #986 

Only implement part of #986

Support FIXEDSTR16 in NumPy imports


pdamme (changes requested):
**Thanks for this PR, @Mango-Clark**. *Enabling the transfer of string data from Python to DAPHNE would be a great step forward for connecting DAPHNE to the Python data science ecosystem.*

So far, DaphneLib is not able to transfer string data from numpy via shared memory at all. In that sense, this PR (while not fully addressing #986) already adds a valuable new feature. Before we can merge this feature, a few points need to be addressed:

**Mandatory changes:**
1. The new script-level test cases should check if the correct data was received by DAPHNE. Currently, they only check the data/value type of the received data object. Instead of using `typeOf`, you could simply `print()` the transferred data object; if necessary, add an instantiation of DAPHNE’s `printObj`-kernel for `DenseMatrix<FixedStr16>` in `src/runtime/local/kernels/kernels.json`.
1. The new script-level test cases should check if strings longer than 15 characters are treated gracefully (e.g., raising an error, but neither leading to a crash nor to truncating values).

**Optional changes:**
1. Please don’t create instances of `OperationNode` directly in your DaphneLib scripts (you do that in your test cases, most likely because we do not support DaphneDSL’s `typeOf`-function in DaphneLib yet). Instead, please use the functions of the `DaphneContext` or the individual node types (e.g., `Matrix.print()`). Feel free to add any unsupported functions you need (e.g., `typeOf`).

-----

**It’s fine to merge this PR, once these points have been addressed.** *Nevertheless, a more comprehensive support for efficiently exchanging string data with Python would be highly desirable.* The following points would be great to address, too (either in this PR or in separate PRs):
1. Data transfer from DAPHNE to Python (opposite direction)
1. Support for arbitrary-length strings.
1. Support for `pandas.Series` with string data and `pandas.DataFrame` with (among others) string columns.

*Furthermore, the efficiency of these data transfers should be thoroughly investigated through experiments.*

src/api/python/daphne/context/daphne_context.py
```diff
@@ -119,7 +118,7 @@ def from_python(self, mat: [], shared_memory=True, verbose=False, return_shape=F
         # Return the matrix, and the original shape if return_shape is set to True.
         return (matrix, (original_mat_length, original_mat_dim2_length, original_mat_dim3_length)) if return_shape else matrix
     
-    def from_numpy(self, mat: np.array, shared_memory=True, verbose=False, return_shape=False):
+    def from_numpy(self, mat: np.ndarray, shared_memory=True, verbose=False, return_shape=False):
```
pdamme, line 121: Is there a reason why you changed the type from `np.array` to `np.ndarray`?

src/api/python/daphne/context/daphne_context.py
```diff
@@ -171,12 +233,23 @@ def from_numpy(self, mat: np.array, shared_memory=True, verbose=False, return_sh
...
+                    " to DAPHNE via shared memory is not supported yet"
+                )
+            elif d_type.kind == 'O':
+                raise RuntimeError("transfering a numpy array of objects to DAPHNE via shared memory is not supported yet")
             else:
                 # TODO Raise an error here?
                 print("unsupported numpy dtype")
 
+            # Data transfer via shared memory.
+            address = mat.ctypes.data
```
pdamme, line 254: Besides moving the extraction of the memory address down here (which is fine), you also switched from `mat.ctypes.data_as(...).value` (with some parameters) to `mat.ctypes.data`. Can you explain the motivation?

src/compiler/utils/CompilerUtils.h
```diff
@@ -226,6 +230,8 @@ struct CompilerUtils {
             // strings) when inserted into the typical "const DT *" template of
             // kernel input parameters.
             return "char";
+        else if (llvm::isa<mlir::daphne::FixedStr16Type>(t))
```
pdamme, line 229: It should be sufficient to add the mapping from the MLIR type to the C++ type name only here. I think the other two cases above are unnecessary for `FixedStr16Type`; we only have them for `StringType` because `StringType` is mapped to `"char"` when used as a scalar, but to `"std::string"` when used as the value type of data object.

src/api/python/daphne/context/daphne_context.py
Mango-Clark: `np.array` is a function, whereas `np.ndarray` is the actual class.
Since Python type annotations expect a class, `np.ndarray` is the correct type hint.

Mango-Clark:
Implemented **Mandatory changes** but working on further.

- 5e1b571165e831f54243bea8065e21de0d0a6d53
      - Before: `std::ostream::write` always wrote all 16 bytes, including trailing NUL padding (`\0`).
      - After: `FixedStr16::operator<<` writes only the actual payload length, so `DenseMatrix<FixedStr16>::print()` outputs
        strings without trailing NUL bytes.
      - Reason: `DaphneLibTest` compares captured stdout exactly. So, trailing NUL bytes caused output mismatches.
        Fixed 16 byte storage and serialization remain unchanged.
        
- 824f79a70b220d73caf5c30e9c3184dc1aea76f8
  - Before: `FixedStr16` shared-memory tests printed only `typeOf`, validating matrix type but not received values.
  - After: Tests print matrices and compare actual bytes, Unicode, object values, shape, and ordering.
    Added `printObj<DenseMatrix<FixedStr16>>` kernel instantiation.

- e877f6cb62e1795fbb3cb6838bfae6546c8cc9e0
  - Before: Boundary tests checked only exception types for strings longer than 15 bytes and unsupported object values.
  - After: Tests also compare exact error messages for 16-byte bytes, UTF-8 Unicode, object bytes, and unsupported objects.

src/api/python/daphne/context/daphne_context.py
pdamme: Okay, good catch!

pdamme (approved):
Thanks @Mango-Clark for addressing both the mandatory and optional changes mentioned in my previous review. It would be great if you could also respond to my two remaining comments on concrete code lines above. Apart from that, **we could already merge this PR**.

------

For a more comprehensive support of string data transfer, please see the suggestions in my previous review:
1. Data transfer from DAPHNE to Python (opposite direction)
1. Support for arbitrary-length strings.
1. Support for `pandas.Series` with string data and `pandas.DataFrame` with (among others) string columns.

Plus experiments on the efficiency of the data transfer.

src/compiler/utils/CompilerUtils.h
Mango-Clark: Good point.
I’ll remove the two context-specific cases and keep only the generic mapping below.

src/api/python/daphne/context/daphne_context.py
Mango-Clark: `data_as(ndpointer(...))` does not validate the specified restrictions in this usage.
This is also evident from `ndim=1`: there was `mat = mat.reshape(-1, 1)`, but no `TypeError` is raised.
The restrictions are only validated when `ndpointer` is used as a ctypes function argument type.

- 8a169216457eaef791cc03dd3134eb266fd1c0ad
  - Added `np.ascontiguousarray(mat)` before extracting the address to ensure C-contiguous storage.

PR #1010, opened 2026-07-07
fix(build): prevent kernel-compile OOM on memory-constrained hosts

### LINKED ISSUE #1009: build: KernelObjLib compile OOMs on Docker Desktop / WSL 2 default sizings (≤ 8 GiB)
### Summary

`./build.sh` on a memory-constrained host (Docker Desktop's default
7.66 GiB VM, WSL 2 default, or any container with `--memory ≤ 8g`) is
killed by the OOM-killer during the DAPHNE build phase. Root cause is
the combination of Ninja's default `nproc + 2` parallelism and one
kernel translation unit (`kernels_94.cpp`, wrapping
`DistributedPipeline`) whose isolated peak compile RSS is **3.63 GiB**
— roughly 4× the median kernel TU — because its header pulls in gRPC,
protobuf, MPI, and the full MLIR dialect closure. New contributors
following `GettingStarted.md`'s recommended dev-container flow hit
this on their first build.

### What happens

Building DAPHNE from source with `./build.sh` on a memory-constrained host
is killed by the OOM-killer during the DAPHNE build phase. The failure
surfaces as:

```
[N/161] Building CXX object src/runtime/local/kernels/CMakeFiles/KernelObjLib.dir/kernels_NN.cpp.o
FAILED: src/runtime/local/kernels/CMakeFiles/KernelObjLib.dir/kernels_NN.cpp.o
g++-13: fatal error: Killed signal terminated program cc1plus
compilation terminated.
```

The victim TU number varies run-to-run because the OOM-killer picks
whichever `cc1plus` happens to be largest at the moment memory runs out.
The failure is unhelpful in the sense that "Killed signal terminated
program cc1plus" gives no hint that the cause is memory pressure or how to
work around it — a new contributor's first build fails silently.

### Where this reproduces

- Docker Desktop with the default Linux VM (7.66 GiB on macOS as of the
  current release).
- WSL 2 with the default memory allocation (typically ~50% of host RAM).
- Any container started with `docker run --memory=<N>g` where `N ≤ 8`.
- Cloud VMs / small CI runners with ≤ 8 GiB RAM.

Verified reproduction (Docker Desktop for macOS, 7.66 GiB Linux VM,
12 vCPUs):

```bash
docker run --rm --entrypoint /bin/bash --memory=8g --memory-swap=8g \
    --mount type=bind,source=$PWD,target=/daphne \
    --mount type=tmpfs,destination=/daphne/thirdparty/installed \
    --workdir /daphne \
    daphneeu/daphne-dev:latest_X86-64_BASE \
    -c "./build.sh --no-deps --installPrefix /usr/local/"
```

Also reproduces at 4 GiB with `--target KernelObjLib`.

### Why it happens

Three independent causes stack. All three have to be present in the
default configuration for the failure to occur reliably.

**Cause 1: default Ninja parallelism is `nproc + 2`.**

`build.sh` calls `cmake --build` with no `--parallel` flag. On a 12-vCPU
laptop that is 14 concurrent `cc1plus` processes. Each generated kernel
TU (`kernels_*.cpp`) peaks around **~1 GiB of RAM** at compile time
because every one of them `#include`s DAPHNE's runtime, which drags in a
substantial MLIR/LLVM header closure. 14 × ~1 GiB comfortably exceeds any
container smaller than ~16 GiB.

This is the standard "one C++ compile job per 2–4 GB of RAM" problem,
worse than usual because DAPHNE's runtime headers are heavy.

**Cause 2: one kernel is dramatically heavier than the rest.**

Isolated per-`cc1plus` peak RSS (measured via `/proc/<pid>/status`
`VmHWM`, single-process, no concurrency, exact command from
`compile_commands.json`; container `daphneeu/daphne-dev:latest_AArch64` on
`g++-13 13.3.0`):

| Translation unit | Kernel-specific include | Peak `VmHWM` | Preprocessed size | Headers |
|---|---|---:|---:|---:|
| `kernels_11.cpp` | `CreateDaphneContext.h` | 0.97 GiB | 8.5 MiB | 656 |
| `kernels_0.cpp`  | `AggAll.h`              | 1.07 GiB | 8.7 MiB | 669 |
| `kernels_1.cpp`  | `AggCol.h` (112 template instantiations) | 1.06 GiB | 8.7 MiB | 669 |
| `kernels_108.cpp`| `BatchNorm2DTestForward.h` | 0.99 GiB | 8.8 MiB | 669 |
| **`kernels_94.cpp`** | **`DistributedPipeline.h`** | **3.63 GiB** | **34.1 MiB** | **1349** |

`kernels_94.cpp` is a 14-line auto-generated file containing a single
function call. Its lone `#include <runtime/local/kernels/DistributedPipeline.h>`
transitively pulls in gRPC, protobuf, MPI, and `<mlir/InitAllDialects.h>`
— roughly 4× the preprocessed size and 2× the header count of any other
kernel TU.

The reason lies inside `DistributedPipeline.h`: the header defines the
body of `distributedPipeline<>()` inline, and the body references
`DistributedWrapper<DTRes>`. The header therefore has to
`#include <runtime/distributed/coordinator/kernels/DistributedWrapper.h>`,
which drags in the entire distributed-runtime closure into every TU that
includes the header — even though only one caller in the whole build ever
instantiates the template.

**Cause 3: the peak RSS of that one file alone can OOM a 4 GiB container.**

Even with `-j 1`, `kernels_94.cpp` alone peaks at 3.63 GiB. On a 4 GiB
container that leaves no headroom for the linker, the shell, or the OS
page cache — the compile succeeds only intermittently. At `-j 2` on 8 GiB,
if the scheduler places `kernels_94.cpp` alongside a second normal kernel
TU, worst-case concurrent RSS is `3.63 + 1.0 ≈ 4.6 GiB` — fine on 8 GiB,
but any third concurrent kernel takes it over.

### Concrete worst-case arithmetic

Assuming Ninja's default `-j` and no cap:

| Container RAM | Default `-j` | Worst-case concurrent peak | Outcome |
|---:|---:|---:|:---|
| 4 GiB  | `nproc + 2` = 6+   | 3.63 (kernels_94) + N×1  | OOM immediately |
| 8 GiB  | `nproc + 2` = 14   | 3.63 + 13×1 ≈ 16.6 GiB   | OOM |
| 16 GiB | `nproc + 2` = 14   | 3.63 + 13×1 ≈ 16.6 GiB   | Marginal — OOMs on some hosts |
| 32 GiB | `nproc + 2` = 14   | 3.63 + 13×1 ≈ 16.6 GiB   | Fine |

The 16 GiB row is why the issue slipped past CI: `ubuntu-latest` runners
have ~16 GiB and complete the build, but sit close to the OOM boundary.

### Environment

- Host: macOS 15.5 (Apple Silicon), Docker Desktop 4.32, 7.66 GiB Linux VM,
  12 vCPUs. Also reproduces on Ubuntu 22.04 in WSL 2 with default memory.
- Container: `daphneeu/daphne-dev:latest_X86-64_BASE` and
  `daphneeu/daphne-dev:latest_AArch64`.
- Compiler: `g++-13` (13.3.0) as shipped in the daphne-dev images.
- Build config: `./build.sh --no-deps --installPrefix /usr/local/`,
  default CMake generator (Ninja), no env vars set.
- Commit: reproduces on current `main` (verified against `c080fe91` on
  fork/main equivalent).

### Impact

New contributors on standard developer hardware (macOS laptops with
Docker Desktop, Windows laptops with WSL 2) cannot build DAPHNE without
knowing to change container memory limits or manually cap `-j`. The
project's own `GettingStarted.md` recommends the dev container, so the
"just run `./build.sh`" onboarding path fails silently.

grapentt:
Closes #1009.

## The issue

Building DAPHNE inside Docker Desktop / WSL 2 with the default VM sizes OOMs. Three things stack:

1. Ninja defaults to `nproc + 2` jobs. On a 12-core laptop that's 14 concurrent `cc1plus`. Each kernel TU peaks around ~1 GiB so 8 GiB VMs die quickly.
2. One kernel is far worse than the rest: `kernels_94.cpp` (the one that wraps `distributedPipeline`) peaks at **3.63 GiB** by itself. `DistributedPipeline.h` had the function body inline, and that body pulls in gRPC, protobuf, MPI, and `<mlir/InitAllDialects.h>`. So every TU including it re-parses ~34 MiB of headers, and one of them ends up carrying the whole cost.
3. First-time contributors have no way to know any of this. The failure is just `Killed signal terminated program cc1plus`.

More detail and measurements in #1009.

## How this PR addresses the issue

**1. Split `DistributedPipeline.h` into declaration + definition.** Standard decl/def split — same pattern already used by `MatMul.cpp`, `Pooling.cpp`, `CreateDaphneContext.cpp` in the same directory. The template body and its heavy includes move into a new `DistributedPipeline.cpp`; the header gets an `extern template` for `DenseMatrix<double>` (the only instantiation per `kernels.json`).

Effect: `kernels_94.cpp` peak drops from **3.63 GiB → 0.97 GiB**. The 3.63 GiB cost is relocated to one hand-written `.cpp` that CMake can target explicitly.

There's also a small drift-check script that runs at configure time (`checkExplicitInstantiations.py`) and fails fast if the explicit-instantiation list in the `.cpp` gets out of sync with `kernels.json` — otherwise a future maintainer adding a new `DTRes` would hit an ugly late linker error.

**2. Two opt-in env-var caps: `DAPHNE_COMPILE_JOBS` and `DAPHNE_KERNEL_COMPILE_JOBS`.** The first is forwarded to `cmake --build --parallel N` — global Ninja cap. The second is a new CMake cache variable that creates a target-scoped `JOB_POOL_COMPILE` on `KernelObjLib` (and `CUDAKernels` when `--cuda`) so kernel TUs specifically get throttled while everything else keeps scaling. Ninja-only; other generators get a `WARNING` and no-op.

Two knobs rather than one because kernel TUs are the memory outlier — you can keep the rest of the build parallelizing at full `nproc`.

**3. `build.sh` auto-detects available RAM and picks safe defaults when the env vars aren't set.** cgroup v2 `memory.max` → cgroup v1 → `/proc/meminfo` → macOS `sysctl` → 4 GB fallback. The formula is `floor((avail-1)/2)` clamped to `[1, nproc]` — roughly one slot per 2 GiB with headroom. If the derived value equals `nproc` (well-resourced host) nothing prints and the build receives `--parallel <nproc>`, which matches Ninja's default concurrency — so the observable behaviour is unchanged in practice.

Sample output on an 8 GiB container:
```
Memory-constrained host detected (8 GB available)
Auto-setting COMPILE_JOBS=3, KERNEL_COMPILE_JOBS=3 to prevent OOM
Override by setting DAPHNE_COMPILE_JOBS / DAPHNE_KERNEL_COMPILE_JOBS env vars
```

Explicit env-var values always win over auto-detection.

## Verified

- **Docker Desktop for macOS, `daphneeu/daphne-dev:latest_AArch64`, `--memory=8g`, no env vars** — auto-detects 8 GB, picks 3/3, clean build in ~6:30, peak container RSS ~5 GiB against the 7.65 GiB cap.
- **WSL 2 + Docker Desktop, `latest_X86-64_BASE`, `--memory=8g`, no env vars** — auto-detects 8 GB, picks 3/3, clean build, `bin/daphne` works.
- **Fork CI (`ubuntu-latest`)** — `_auto_cj = 4 = nproc` so the auto-detect message is correctly suppressed and no `--parallel` cap is passed. Build graph matches `main`. All tests pass (2462 test cases, 425216 assertions), `Hello World!` script prints.
- **Local test run** (AArch64 dev image + `--memory=8g`): 2349/2462 test cases pass. The 113 failures are all in `test/api/python/DaphneLibTest.cpp` due to `numpy` / `pandas` missing from the AArch64 dev image (unrelated to this PR); zero failures elsewhere, in particular no `[distributed]` failures — the header split linked cleanly.
- **Explicit override**: `DAPHNE_COMPILE_JOBS=4 DAPHNE_KERNEL_COMPILE_JOBS=2` takes precedence, auto-detect banner suppressed as designed.
- **`get_memory_gb()` / `calc_daphne_jobs()`** have shell-level unit tests covering cgroup v2 (single limit / all `max` / min-of-chain), cgroup v1 sentinel rejection, `/proc/meminfo`, malformed inputs, and the well-resourced-host suppression at `2·nproc+1 GB`.

## Changes

| File | What |
|---|---|
| `build.sh` | New `get_memory_gb()` + `calc_daphne_jobs()`. Auto-detect block picks safe defaults when env vars are unset. Forwards them to `cmake --build --parallel` and `-DDAPHNE_KERNEL_COMPILE_JOBS`. |
| `CMakeLists.txt` | New cache variable `DAPHNE_KERNEL_COMPILE_JOBS` with positive-integer validation and Ninja-only warning. Registers `JOB_POOLS`. |
| `src/runtime/local/kernels/CMakeLists.txt` | Attaches `JOB_POOL_COMPILE daphne_kernel_compile` to `KernelObjLib` (and `CUDAKernels` under `--cuda`). Adds `DistributedPipeline.cpp` to sources. Runs `checkExplicitInstantiations.py` at configure. |
| `src/runtime/local/kernels/DistributedPipeline.h` | Declaration-only. Drops `#include <DistributedWrapper.h>`, `extern template` for `DenseMatrix<double>`, forward-declares `class Structure`. |
| `src/runtime/local/kernels/DistributedPipeline.cpp` *(new)* | Template body + explicit instantiation. |
| `src/runtime/local/kernels/checkExplicitInstantiations.py` *(new)* | Configure-time drift check between `kernels.json` and `DistributedPipeline.cpp`. |
| `doc/development/BuildingDaphne.md` | New "Memory Requirements" section documenting auto-detect + overrides. |
| `doc/GettingStarted.md` | Hardware bullet pointing at the memory section. |

## Out of scope

- Below ~4 GB the build still OOMs, because `DistributedPipeline.cpp` itself peaks at 3.63 GiB and there's no `JOBS` below 1. Reducing that would need the same decl/def split applied one level deeper inside `DistributedWrapper.h` — larger refactor, follow-up.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 build/
 build-*/
 build_*/
+!test/build/
 /bin
 /lib
 /tmp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,6 +119,32 @@ find_package(fmt)
 add_definitions(-DSPDLOG_FMT_EXTERNAL)
 
 option(USE_CUDA "Whether to activate compilation of CUDA features" OFF)
+
+# Optional cap on parallel jobs for the memory-heavy kernel targets, via a Ninja job pool.
+# The generated kernels_*.cpp files instantiate a lot of templates, and each cc1 can
+# hold several GB resident, so building them all at once can hit the OOM killer on hosts
+# with little RAM per core (WSL default, small VMs, containers with a --memory limit).
+# Set to a positive integer to cap KernelObjLib (and CUDAKernels when USE_CUDA=ON);
+# leave empty for uncapped (the default). Ninja only, ignored with a warning otherwise.
+set(DAPHNE_KERNEL_COMPILE_JOBS "" CACHE STRING
+    "Max concurrent compile jobs for the pre-compiled kernel targets (Ninja only; empty = uncapped).")
+
+if(NOT "${DAPHNE_KERNEL_COMPILE_JOBS}" STREQUAL "")
+    if(NOT DAPHNE_KERNEL_COMPILE_JOBS MATCHES "^[1-9][0-9]*$")
+        message(FATAL_ERROR "DAPHNE_KERNEL_COMPILE_JOBS must be a positive integer, got '${DAPHNE_KERNEL_COMPILE_JOBS}'.")
+    endif()
+    if(NOT CMAKE_GENERATOR MATCHES "Ninja")
+        message(WARNING "DAPHNE_KERNEL_COMPILE_JOBS is set but generator is '${CMAKE_GENERATOR}'. JOB_POOLS is Ninja-only; ignoring.")
+    else()
+        set_property(GLOBAL APPEND PROPERTY JOB_POOLS daphne_kernel_compile=${DAPHNE_KERNEL_COMPILE_JOBS})
+        # Single flag used by the kernel target definitions to decide whether
+        # to attach JOB_POOL_COMPILE. Referencing the pool without registering
+        # it makes Ninja fail with "unknown pool name" at build-graph load,
+        # so both sides of the flag must stay in sync.
+        set(DAPHNE_KERNEL_POOL_ACTIVE TRUE)
+    endif()
+endif()
+
 include(CheckLanguage)
 check_language(CUDA)
 if(USE_CUDA AND CMAKE_CUDA_COMPILER)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,7 +136,7 @@ if(NOT "${DAPHNE_KERNEL_COMPILE_JOBS}" STREQUAL "")
     if(NOT CMAKE_GENERATOR MATCHES "Ninja")
         message(WARNING "DAPHNE_KERNEL_COMPILE_JOBS is set but generator is '${CMAKE_GENERATOR}'. JOB_POOLS is Ninja-only; ignoring.")
     else()
-        set_property(GLOBAL APPEND PROPERTY JOB_POOLS daphne_kernel_compile=${DAPHNE_KERNEL_COMPILE_JOBS})
+        set_property(GLOBAL PROPERTY JOB_POOLS daphne_kernel_compile=${DAPHNE_KERNEL_COMPILE_JOBS})
         # Single flag used by the kernel target definitions to decide whether
         # to attach JOB_POOL_COMPILE. Referencing the pool without registering
         # it makes Ninja fail with "unknown pool name" at build-graph load,
--- a/build.sh
+++ b/build.sh
@@ -138,8 +138,9 @@ calc_daphne_jobs() {
     local cj=$(( usable / 2 ))
     [ "$cj" -lt 1 ] && cj=1
     [ "$cj" -gt "$ncpu" ] && cj=$ncpu
-    local kj=$usable
-    [ "$kj" -gt "$cj" ] && kj=$cj
+    # Kernel pool never exceeds the global cap (kernel slots consume global
+    # slots too); today both knobs use the same value.
+    local kj=$cj
     echo "$cj $kj"
 }
 
--- a/build.sh
+++ b/build.sh
@@ -52,6 +52,97 @@ sed_inplace() {
     fi
 }
 
+# How much RAM this process can actually use, in whole GB (rounded down).
+# Tries the cgroup limits first (so it respects docker --memory), then
+# /proc/meminfo, then macOS sysctl, and falls back to 4 if none of those work.
+# /proc/meminfo reports host RAM, so it's only accurate when there's no container limit.
+get_memory_gb() {
+    local mem_bytes=0
+
+    if [ "$IS_DARWIN" = "1" ]; then
+        local raw
+        raw=$(sysctl -n hw.memsize 2>/dev/null) || raw=0
+        if [[ "$raw" =~ ^[0-9]+$ ]] && [ "$raw" -gt 0 ]; then
+            echo $(( raw / 1024 / 1024 / 1024 ))
+        else
+            echo 4
+        fi
+        return
+    fi
+
+    # cgroup v2: walk the cgroup hierarchy, take the minimum non-"max" memory.max
+    local cg_path
+    cg_path=$(awk -F: '/^0:/{print $3; exit}' /proc/self/cgroup 2>/dev/null)
+    if [ -n "$cg_path" ]; then
+        local path="$cg_path" val min_limit=0 mem_max_file
+        while true; do
+            mem_max_file="/sys/fs/cgroup${path}/memory.max"
+            if [ -r "$mem_max_file" ]; then
+                val=$(cat "$mem_max_file" 2>/dev/null)
+                if [ "$val" != "max" ] && [[ "$val" =~ ^[0-9]+$ ]]; then
+                    if [ "$min_limit" -eq 0 ] || [ "$val" -lt "$min_limit" ]; then
+                        min_limit="$val"
+                    fi
+                fi
+            fi
+            [ "$path" = "/" ] || [ -z "$path" ] && break
+            path=$(dirname "$path")
+            [ "$path" = "." ] && break
+        done
+        [ "$min_limit" -gt 0 ] && mem_bytes="$min_limit"
+    fi
+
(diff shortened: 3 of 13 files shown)


Cojo01:
I tested it locally on Windows. The OOM error is no longer occurring and the build finishes successfully. Looks good to me!

